PR #7, opened 2026-04-12
sdlc: file out-of-scope ideas as issues

dburkart:
## Summary
- Adds a "Capturing out-of-scope ideas" section to the `sdlc` skill directing any interesting idea, bug, or follow-up task that surfaces mid-work — from either the user or Claude — to become a GitHub issue immediately rather than living in chat history.
- Calls out the MCP tool to use (`mcp__github__create_issue`), what goes in the body (what + why with enough cold-read context), labels, and linking deferred items back from the originating PR.

## Test plan
- [x] N/A — docs-only change to a skill file; no build/test impact.
- [ ] Skim the new section on GitHub and confirm it reads the way you want.


dburkart (commented):
Looks good!

PR #9, opened 2026-04-12
M4: guard page under the #DF IST stack

### LINKED ISSUE #6: M4: guard page under #DF IST stack
Part 2 of milestone 4. Depends on #5 (paging foundation) being merged first — we need `mem::paging::unmap` to exist before we can install a guard page.

## Context
The double-fault ISR currently runs on a 20 KiB static IST stack defined in `kernel/src/arch/x86_64/gdt.rs`. If anything underflows that stack — a runaway recursion during a #DF, say — we silently scribble over whatever `.bss` sits below it and hope for the best. That's a debugging trap waiting to happen.

With the paging API from #5 in hand, we can put one unmapped page immediately below the IST stack. A stack overflow then hits a #PF with a fault address inside the guard range, which is loud and locatable instead of silent corruption.

## Goal
Unmap one 4 KiB page directly below `DOUBLE_FAULT_STACK` so IST overflow produces a #PF with a fault address inside the guard, not silent memory corruption.

### Approach: align + unmap-below (preferred over reallocating)
`DOUBLE_FAULT_STACK` stays a static, but we:
1. Add `#[repr(align(4096))]` so the stack's low end lands on a page boundary.
2. After `paging::init`, call a new `ist_guard::install()` that unmaps the 4 KiB page whose start address equals `&DOUBLE_FAULT_STACK[0]`.
3. The TSS stack-top pointer is unchanged — still points at the high end of the array. Overflow walks downward past address 0 of the array, into the now-unmapped guard page → #PF.

The alternative (allocate the stack from frames and HHDM-map 6 pages, leave the lowest unmapped) is more flexible but more moving parts for no current payoff. Going with the static + align approach.

### Init order
```
mem::init
  → frame::init
  → heap::init
  → paging::init
  → ist_guard::install   // new — requires paging to be up
```
Must happen before the first time we could take a #DF, which in practice means before `sti`. `mem::init` already runs before `x86_64::instructions::interrupts::enable()` in `main.rs`, so we're fine.

### Fault-test exercise
Under the existing `fault-test` feature, after `vibix::init()` call a `recurse()` fn that blows the IST stack. Expected: #PF serial print with a fault address inside the guard page. This is manual verification — hard to distinguish from other crashes in a CI gate.

### Smoke marker
Add `"paging: IST guard installed"` to `SMOKE_MARKERS`.

## Files to modify / add
- `kernel/src/arch/x86_64/gdt.rs` — align `DOUBLE_FAULT_STACK` to 4 KiB; expose `df_stack_guard_addr()` returning the VA of the page to unmap.
- `kernel/src/arch/x86_64/ist_guard.rs` — **new**, `install()` unmaps the page via `mem::paging::unmap`.
- `kernel/src/arch/x86_64/mod.rs` — export `ist_guard`.
- `kernel/src/mem/mod.rs` — call `arch::x86_64::ist_guard::install()` at the end of `mem::init`.
- `kernel/src/main.rs` — under `#[cfg(feature = "fault-test")]`, a `recurse()` helper invoked after `vibix::init()` but kept behind the fault-test gate along with the existing `ud2` trigger.
- `xtask/src/main.rs` — add `"paging: IST guard installed"` to `SMOKE_MARKERS`.

## Reuse
- `mem::paging::unmap` / `translate` (landed in #5).
- Existing `fault-test` feature gating pattern in `main.rs`.

## Verification
1. `cargo xtask run` — serial shows `paging: IST guard installed`.
2. `cargo xtask run --fault-test` — deep recursion produces a #PF print with fault address inside the guard page (one page below `DOUBLE_FAULT_STACK`).
3. `cargo xtask test` — existing integration tests (including `paging` from #5) continue to pass.
4. `cargo xtask smoke` — new marker present.

## Out of scope
- Per-task stack guard pages (this issue only does the IST; general kernel stacks come when we have tasks).
- Moving the IST stack off the static and onto a heap allocation.
- Catching #DF recursively (double-double-fault story is untouched).

Branch: `m4-ist-guard`.

Blocked by #5.

dburkart:
## Summary
- **gdt:** wrap `DOUBLE_FAULT_STACK` in a `#[repr(C, align(4096))]` shim so the low end lands on a page boundary; expose `df_stack_guard_addr()`.
- **arch::ist_guard:** new module, `install()` unmaps the 4 KiB page at the low end of the IST stack via `paging::unmap`. Called from the tail of `mem::init` once paging is up.
- **usable stack:** shrinks from 20 KiB to 16 KiB — the lowest page becomes the guard. TSS still points at the high end; overflow walks *down* into the unmapped page and surfaces as a #PF instead of silently corrupting `.bss`.
- **xtask:** new smoke marker `"paging: IST guard installed"`.
- **fault gating:** new `ist-overflow-test` feature flips on a `recurse()` in `_start` that deliberately blows the IST stack. Deliberately a separate feature from `fault-test` — manual verification only, since a successful overflow is indistinguishable from any other crash to CI.

Out of scope (see issue body): per-task stack guard pages, moving the IST stack off a static, catching #DF recursively.

Closes #6.

## Test plan
- [x] `cargo xtask test` — all 12 host + 5 QEMU integration tests pass (serial shows `paging: IST guard installed @ 0xffffffff8000e000`).
- [x] `cargo xtask smoke` — all 11 markers present.
- [ ] Manual: `cargo build --features ist-overflow-test ...` + boot under QEMU, confirm serial shows a #PF print with the fault address inside the guard page.
- [ ] Eyeball a `cargo xtask run` serial log for no regressions.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -27,6 +27,11 @@ pc-keyboard = "0.8"
 default = []
 # Enable to trigger a UD2 after init; verifies the invalid-opcode handler path.
 fault-test = []
+# Enable to deliberately overflow the #DF IST stack after init; verifies
+# that the guard page below the stack catches overflow as a #PF instead
+# of letting it silently corrupt .bss. Manual verification only —
+# successful overflow is indistinguishable from other crashes in CI.
+ist-overflow-test = []
 
 [[test]]
 name = "basic_boot"
--- a/kernel/src/arch/x86_64/gdt.rs
+++ b/kernel/src/arch/x86_64/gdt.rs
@@ -1,5 +1,10 @@
 //! GDT + TSS. A dedicated IST stack for #DF keeps double-faults from
 //! triple-faulting when the current kernel stack is the thing on fire.
+//!
+//! The IST stack is page-aligned so that `ist_guard` can unmap the
+//! lowest page as a guard: an overflow walks down past that boundary
+//! and takes a #PF whose fault address pinpoints the overflow, rather
+//! than silently scribbling through whatever `.bss` sits below.
 
 use spin::Lazy;
 use x86_64::instructions::segmentation::{Segment, CS, DS, ES, FS, GS, SS};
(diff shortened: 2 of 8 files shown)


dburkart:
@greptileai review

Issue #19, opened 2026-04-12
M7: ACPI + APIC/IOAPIC


dburkart:
The 8259 PIC was fine for M3 and fine for a single core. Two future things it can't do: SMP (APs need Local APIC IPIs to boot), and high-resolution / well-behaved interrupts beyond the legacy ISA lines. It's time to move off of it.

## Goal
Parse the ACPI tables enough to find every Local APIC and IOAPIC, disable the 8259, and route our existing IRQs (PIT, PS/2 keyboard) through the IOAPIC instead.

## Rough shape
- ACPI: Limine hands us an RSDP pointer on request. Walk RSDP → XSDT → MADT. A small parser is fine — we don't need a general ACPI/AML interpreter (no DSDT execution yet). Existing crates (`acpi`, `rsdp`) do this well if we'd rather not write it; the licensing is compatible.
- From the MADT: LAPIC base address, list of IOAPICs, interrupt source overrides (the legacy ISA-to-GSI remapping is almost always needed).
- Mask all lines on the 8259 and set up LAPIC: enable, set spurious vector, map the PIT and keyboard IRQs to IOAPIC redirection entries that point at the BSP's LAPIC.
- Rewrite `pic::notify_eoi` to write to the LAPIC EOI register instead of the 8259 command port.
- Keep the current vector numbers (0x20 for PIT, 0x21 for keyboard) so the IDT and ISRs don't need to move.

## Verification
- Smoke markers: `"acpi: MADT parsed"`, `"apic: BSP online"`, the existing PIT/keyboard markers continue to appear.
- Integration test: existing `timer_tick` still passes — it's a good oracle that the timer IRQ is still landing.
- Manual: type in QEMU, confirm keystrokes still echo.

## Out of scope
- AML / DSDT execution (needed for power management, not for IRQ routing).
- HPET (nice future timer; PIT stays for now).
- x2APIC mode (stick to xAPIC — simpler MMIO interface).
- SMP and IPIs. This issue just brings the BSP up on the APIC; AP startup is its own milestone.

dburkart:
@claude file an issue for HPET

PR #34, opened 2026-04-12
M7: ACPI + APIC/IOAPIC for the BSP

dburkart:
## Summary
- Parse ACPI (RSDP → XSDT/RSDT → MADT) to discover LAPIC base, IOAPIC(s), and ISA interrupt source overrides; hand-rolled — no new crate deps.
- Bring the BSP's xAPIC online (spurious vector 0xFF + software enable, TPR cleared) and mask all IOAPIC redirection entries.
- Route IRQ0 (PIT → 0x20) and IRQ1 (keyboard → 0x21) through the IOAPIC, honoring MADT source overrides for GSI, polarity, and trigger mode.
- Remap + fully mask the 8259 so any legacy spurious IRQ doesn't hit a CPU exception vector; `pic::notify_eoi` now writes to the LAPIC EOI register.
- Add `mem::paging::map_phys_into_hhdm` for on-demand r/w (and UC for MMIO) HHDM mappings of physical ranges Limine doesn't cover — ACPI tables in reserved/BIOS ROM memory and LAPIC/IOAPIC MMIO.
- Split `arch::init` into `init()` (GDT/IDT/PIC-mask) + `init_apic(rsdp, hhdm)` so ACPI/APIC bring-up runs after the mapper is alive.
- New integration test `apic_online`; three new smoke markers.

Out of scope: AML/DSDT, HPET (#33), x2APIC, AP startup and IPIs.

Refs #19.

## Test plan
- [x] `cargo xtask smoke` — all 15 markers present (`acpi: MADT parsed`, `apic: BSP online`, `ioapic: initialized` added).
- [x] `cargo xtask test` — all integration tests pass under QEMU (`timer_tick` + `preempt` continue to pass, proving timer IRQs still land; new `apic_online` verifies MADT/IOAPIC topology).
- [x] Manual: `cargo xtask run`, type in QEMU, confirm keystrokes echo.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -60,3 +60,7 @@ harness = false
 [[test]]
 name = "preempt"
 harness = false
+
+[[test]]
+name = "apic_online"
+harness = false
--- a/kernel/src/acpi.rs
+++ b/kernel/src/acpi.rs
@@ -1,0 +1,293 @@
+//! Minimal ACPI table parser — just enough to discover LAPIC / IOAPIC
+//! topology from the MADT. We do NOT walk the DSDT or execute AML; that
+//! is a future milestone. We read:
+//!
+//!   RSDP → XSDT (or RSDT, on rev 0 firmware) → MADT → entries
+//!
+//! Limine hands us a pre-mapped virtual pointer to the RSDP in its HHDM
+//! window (base revision ≥ 2), so we treat all physical addresses we
+//! encounter while walking tables as HHDM offsets too.
+//!
+//! Reference: ACPI Specification, section 5 (RSDP/XSDT) and section
+//! 5.2.12 (MADT).
+use core::mem;
+use core::slice;
+
+use spin::Once;
+use x86_64::structures::paging::PageTableFlags;
+
+use crate::mem::paging;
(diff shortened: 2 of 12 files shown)


dburkart:
Thanks CodeRabbit. Addressed in 2a39e59:

- **SDT length + MADT body guard (critical, critical):** both now assert length bounds before pointer arithmetic / slicing.
- **RSDP + SDT checksums (major):** validated (v1 sum over 20 bytes, v2 over `length` bytes; SDT sum over `header.length`).
- **`map_phys_into_hhdm` overflow (major):** rounded end now uses `checked_add` with fail-fast on overflow.
- **`ioapic_for_gsi` 240 window (minor):** removed the helper — `route_legacy_irq` already reads each IOAPIC's real max-redirection-entry count inline, so the speculative public API wasn't earning its keep.
- **`pic.rs` fmt (minor):** fixed in dfaef23.

Deferring the `main.rs` / `lib.rs::init()` boot-sequence dedup nit — `_start` interleaves framebuffer init and extra logging between the two calls, so extracting a shared helper would either leak those concerns into the helper or fracture the ordering further. Not worth it on this PR.

PR #71, opened 2026-04-12
task: add per-task stack guard pages

### LINKED ISSUE #17: Per-task stack guard pages
Generalization of #6. The IST gets a guard page; *kernel task* stacks should too, once they exist. Without it, a runaway recursion in a task silently corrupts whatever task's stack (or heap chunk) lives below it in memory — much harder to debug than a clean #PF.

## Goal
Every kernel task created via `task::spawn` gets a 4 KiB unmapped guard page at the low end of its stack. Stack overflow → #PF with a fault address inside the guard → we log which task overflowed and halt.

## Shape
- When `task::spawn` allocates a task stack, instead of `Box<[u8]>` off the heap, reserve a page-aligned VA range and `paging::map_range` all but the lowest page. The lowest page stays unmapped.
- `task::Task::stack_top()` points at the high end as before; nothing else in the scheduler changes.
- On #PF, a helper looks up the faulting address against every task's guard range and, if it hits, logs `task N stack overflow` before halting. Otherwise the handler behaves as it does today.
- Reuse the IST guard's discipline: `Page::from_start_address` to catch misalignment loudly, matching #9.

## Out of scope
- Growing task stacks (this issue is about catching overflow, not fixing it).
- User-mode task stacks (comes with userspace, likely wants a similar-but-separate mechanism).

**Blocked by:** #16 (no tasks → no task stacks to guard).

dburkart:
Each kernel task spawned via `task::spawn` now gets a dedicated VA slot from the `0xFFFF_D000_0000_0000` window: a 4 KiB unmapped guard page at the low end followed by 16 KiB of RW/NX-mapped stack pages. Stack overflow → #PF with fault address inside the guard → `task N stack overflow` diagnostic + halt.

Closes #17.

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -71,6 +71,19 @@ extern "x86-interrupt" fn page_fault(frame: InterruptStackFrame, code: PageFault
         }
     }
 
+    // Check whether the fault address lands inside a kernel task's guard
+    // page — if so, this is a stack overflow, not a generic fault.
+    if let Some(task_id) = crate::task::find_stack_overflow(addr_u64 as usize) {
+        serial_println!(
+            "EXCEPTION: #PF task {} stack overflow addr={:#x} code={:?}\n{:#?}",
+            task_id,
+            addr_u64,
+            code,
+            frame
+        );
+        hang();
+    }
+
     serial_println!(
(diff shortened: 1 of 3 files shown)


dburkart:
@claude address review feedback

PR #72, opened 2026-04-12
mem: derive kernel PML4 flags from ELF PT_LOAD headers

### LINKED ISSUE #48: Derive kernel section permissions from ELF PT_LOAD headers
Today `build_kernel_pml4` (issue #12) uses linker symbols (`__text_start`/`__text_end`, etc.) to decide the flags for each kernel-image region when building the fresh tree. This works but couples the paging code to the linker script — adding a new kernel section means touching both files, and the flags we apply are hard-coded per-section rather than derived from the section's actual permissions.

## What
Switch to reading the kernel ELF's program headers at boot via a Limine `ExecutableFileRequest`, and derive `{R, W, X}` flags from `p_flags` (`PF_X` / `PF_W`) directly. Each `PT_LOAD` segment becomes a range to map with the corresponding page-table flags.

## Why
- Section boundaries and flags live in one place (the ELF) instead of two (linker script + paging.rs).
- New kernel sections pick up correct permissions without code changes.
- Closer to what a future user-mode ELF loader will do anyway — shared plumbing.

## Related
Follow-up from #12. Not blocking anything, purely a cleanup / robustness win.

dburkart:
## Summary

- `populate_kernel_image` now iterates the kernel ELF's `PT_LOAD` program headers (via a new `limine::ExecutableFileRequest`) and derives `{R, W, X}` directly from each segment's `p_flags`. The linker-symbol array (`__text_start`, `__rodata_start`, ...) and the hard-coded per-section flag table are gone — adding a new kernel section no longer requires touching `paging.rs`.
- New `mem::elf` module: ~80 lines, hand-rolled ELF64 Ehdr/Phdr walker. Deliberately no `xmas-elf` / `goblin` — a future userspace ELF loader can grow this module instead of adding a second parser.
- New `paging::flags(va) -> Option<PageTableFlags>` helper lets tests inspect leaf PTE flags.

Closes #48.

## Test plan

- [x] `cargo xtask build`
- [x] `cargo xtask test` — 19 lib tests pass; `pml4_switch` now has 5 tests (added `elf_derived_segment_flags`) and all pass. Observed flags in-kernel: `text=PRESENT`, `rodata=PRESENT|NO_EXECUTE`, `data=PRESENT|WRITABLE|NO_EXECUTE`.
- [x] `cargo xtask smoke` — all 16 serial markers present.
- [x] `cargo fmt --all --check` clean.

### DIFF
--- a/kernel/src/boot.rs
+++ b/kernel/src/boot.rs
@@ -3,8 +3,8 @@
 //! take control.
 
 use limine::request::{
-    ExecutableAddressRequest, FramebufferRequest, HhdmRequest, MemoryMapRequest, RequestsEndMarker,
-    RequestsStartMarker, RsdpRequest, StackSizeRequest,
+    ExecutableAddressRequest, ExecutableFileRequest, FramebufferRequest, HhdmRequest,
+    MemoryMapRequest, RequestsEndMarker, RequestsStartMarker, RsdpRequest, StackSizeRequest,
 };
 use limine::BaseRevision;
 
@@ -34,6 +34,10 @@ pub static STACK_REQUEST: StackSizeRequest = StackSizeRequest::new().with_size(S
 #[used]
 #[link_section = ".limine_requests"]
 pub static KERNEL_ADDRESS_REQUEST: ExecutableAddressRequest = ExecutableAddressRequest::new();
+
+#[used]
+#[link_section = ".limine_requests"]
+pub static KERNEL_FILE_REQUEST: ExecutableFileRequest = ExecutableFileRequest::new();
 
 #[used]
 #[link_section = ".limine_requests"]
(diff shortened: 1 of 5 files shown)


dburkart:
Pushed `7a2e091` addressing the two actionable findings:

- **CodeRabbit P1 (overflow-safe phdr bounds)** — fixed. `phoff + phnum * phsz` is now `checked_mul` + `checked_add` with an explicit `expect`, and the subsequent `assert` uses the checked sum.
- **CodeRabbit P1 (per-page flag union)** — fixed. `populate_kernel_image` now walks segments into a `BTreeMap<u64, PageTableFlags>` first (WRITABLE if any covering segment is writable, NO_EXECUTE only if every covering segment is NX), then maps each page exactly once. The commit message spells out why this matters even though the linker's page-aligned segment boundaries mean no page is currently covered twice.
- **Greptile P2 (phdr alignment assertion)** — fixed. Added `align_offset(align_of::<Elf64Phdr>()) == 0` assertion before materializing the slice reference.

Deferred: Greptile P2 endianness/`e_machine` check and the KASLR `p_vaddr` comment. Both are defensive hardening for scenarios that don't exist today (non-x86_64 ELF, PIE/KASLR kernel) and Greptile itself flagged them as non-blocking. Happy to file a follow-up issue if preferred.

PR #145, opened 2026-04-13
task: exit path with stack, VMA, and PML4 reclaim

### LINKED ISSUE #132: task: exit path and resource reclaim (stack, PML4, VMA frames)
## Motivation
Tasks are declared \`fn() -> !\` and have no exit path. \`NEXT_STACK_VA\` is a monotonic bump allocator that never reclaims VA or physical frames; every spawned task's stack pages, PML4, and demand-paged VMA frames leak permanently. This is a prerequisite for userspace processes that need to exit cleanly and for any meaningful multi-process workload.

## Work
- [ ] Add \`task::exit() -> !\` that removes the current task from the scheduler and marks it for cleanup
- [ ] Walk the task's PML4 lower half, unmap all present pages, and return the backing frames to the global frame allocator
- [ ] Unmap all VMA-backed pages (iterate \`task.vmas\`) and free their physical frames before dropping the \`VmaList\`
- [ ] Free the PML4 frame itself after the lower-half walk
- [ ] Reclaim or track the stack VA slot (at minimum log the leak; ideally return the VA range to a free-list)
- [ ] Add an integration test: spawn a task that calls \`task::exit()\`, assert the scheduler keeps running without panic or leaked detectable corruption

## Context
- \`kernel/src/task/task.rs\`: \`NEXT_STACK_VA\` bump allocator, comment "task stacks are never freed (tasks don't exit yet)"
- \`kernel/src/mem/vma.rs\`: \`VmaList\` has no remove/drain; frames mapped by the \`#PF\` resolver are never returned
- \`kernel/src/mem/paging.rs\`: \`new_task_pml4\` allocates a frame that is never freed
- Related: #129 (demand paging adds VMA-backed frames that also need reclaim), #123 (process lifecycle depends on task exit)

dburkart:
Closes #132

## Summary
- Add `task::exit() -> !` that parks the current task in a new `Scheduler::pending_exit` slot and context-switches to the next ready task (never returns).
- Reap from the top of `preempt_tick`: unmap + free the victim's stack pages (shared upper-half), walk its `VmaList` to unmap + free per-PML4 private frames (`AnonZero` returns frames to the allocator; `Cow` only frees when the mapped frame differs from the shared source), free the PML4 frame, then drop the `Box<Task>`.
- Stack VA slots from the bump allocator stay leaked — logged per reap. Reclaiming the VA range would need a free-list on `NEXT_STACK_VA` and is out of scope.

## Test plan
- [x] `cargo xtask test` — new `task_exit` integration test passes
  - `exit_removes_task`: spawn → exit → task count returns to baseline.
  - `exit_loop_no_leak`: 64 spawn/exit cycles without allocator exhaustion.
- [x] `cargo xtask smoke` — all 22 markers present

---

> [!NOTE]
> **Medium Risk**
> Touches core scheduler and paging teardown logic and frees page-table frames; bugs could leak memory or free mappings still in use (TLB/aliasing), though changes are scoped and covered by a new integration test.
> 
> **Overview**
> Adds `task::exit() -> !` to terminate the current task by immediately context-switching to another runnable task and enqueueing the exiting `Task` for deferred cleanup.
> 
> On the next timer tick, `preempt_tick` now drains `Scheduler::pending_exit` and reaps each victim by unmapping/freeing its stack pages and VMA-backed frames (including special handling for `Cow` source vs private copies), then freeing the task’s PML4 frame.
> 
> Paging gains arbitrary-PML4 unmap helpers (`unmap_in_pml4`, `unmap_and_free_in_pml4`) with unconditional local TLB invalidation, plus `free_pml4_frame`; VMAs add `VmaList::iter()` to support the reaper. A new `task_exit` QEMU integration test validates tasks are removed and repeated spawn/exit cycles don’t exhaust frames, and `xtask test` now includes it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1aa505eb94659b89116ebe3b30aa7e7ef9c47efb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -127,3 +127,7 @@ harness = false
 [[test]]
 name = "cow_vma"
 harness = false
+
+[[test]]
+name = "task_exit"
+harness = false
--- a/kernel/src/mem/paging.rs
+++ b/kernel/src/mem/paging.rs
@@ -375,6 +375,53 @@ pub fn cow_copy_and_remap(
     Ok(new_frame)
 }
 
+/// Unmap `page` from `pml4_frame` and return its backing frame. Unlike
+/// [`unmap`], scopes to a specific PML4 — the reaper path in
+/// `task::exit` uses this to clean up a dead task's private mappings
+/// while running on a different (live) PML4.
+pub fn unmap_in_pml4(
+    pml4_frame: PhysFrame<Size4KiB>,
+    page: Page<Size4KiB>,
+) -> Result<PhysFrame<Size4KiB>, UnmapError> {
+    let hhdm = HHDM_OFFSET
+        .lock()
+        .expect("paging::init must run before unmap_in_pml4");
+    let l4 = unsafe { pml4_as_mut(pml4_frame, hhdm) };
+    let mut mapper = unsafe { OffsetPageTable::new(l4, hhdm) };
+    let (frame, flush) = mapper.unmap(page)?;
+    if Cr3::read().0 == pml4_frame {
+        flush.flush();
+    } else {
+        flush.ignore();
+    }
+    Ok(frame)
+}
+
+/// Unmap `page` from `pml4_frame` and return its backing frame to the
+/// global allocator. Mirrors [`unmap_and_free`] for the
+/// arbitrary-PML4 case.
+pub fn unmap_and_free_in_pml4(
+    pml4_frame: PhysFrame<Size4KiB>,
+    page: Page<Size4KiB>,
+) -> Result<(), UnmapError> {
+    let frame = unmap_in_pml4(pml4_frame, page)?;
+    // SAFETY: caller asserts the frame came from the global allocator.
+    unsafe { KernelFrameAllocator.deallocate_frame(frame) };
+    Ok(())
+}
+
+/// Return `pml4_frame` to the global frame allocator.
+///
+/// # Safety
+/// The caller must guarantee that no CPU is using `pml4_frame` as its
+/// active page table and that no outstanding reference (PTE alias,
+/// OffsetPageTable, cached `&mut PageTable`) still exists. Kernel
+/// upper-half entries reference shared L3 tables owned by the kernel
+/// PML4 — freeing this L4 frame does not affect them.
+pub unsafe fn free_pml4_frame(pml4_frame: PhysFrame<Size4KiB>) {
+    KernelFrameAllocator.deallocate_frame(pml4_frame);
+}
+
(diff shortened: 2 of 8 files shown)


dburkart:
Thanks Bugbot. Responses:

1. **Bootstrap task exit frees the kernel PML4** — fixed in the next commit with an explicit assertion in `task::exit` rejecting `id == 0`. The bootstrap inherits the kernel PML4 and the bootloader-owned boot stack; neither is the exiter's to reap, so failing loudly is the right behavior.

2. **Lower-half intermediate page-table frames leaked** — real, deferred to #146. Practical impact on this PR's tests is zero (the exit workers install no VMAs so no lower-half intermediates are ever allocated), but it's the right follow-up for when real userspace tasks with faulted VMA pages start exiting.

dburkart:
Bugbot second pass:

- **Stack reap in ISR deadlocks on global MAPPER lock** — real, fixed in fbe9d9c. Stack unmap now goes through `unmap_and_free_in_pml4(victim.cr3, page)` like the VMA cleanup; the shared upper-half L3 aliasing means the unmap still propagates to every PML4, and the transient `OffsetPageTable` never touches `MAPPER`.

- Intermediate-table leak (first review, re-flagged) — still deferred to #146 as explained in the previous comment.

CodeRabbit's `cargo fmt` comment was resolved by the earlier `fix: apply rustfmt` commit.

dburkart:
Auto-engineer paused — 3rd review-response pass hit the playbook's stop rule.

CI status: `test` ✅, `CodeRabbit` ✅, `Cursor Bugbot` NEUTRAL (1 new actionable finding + 1 deferred from the first pass).

Bugbot's newest finding (`ddf753ac-dd25-4a37-9035-983aed04d162`) flags a systemic issue: `reap_pending` still acquires blocking locks (frame allocator, `HHDM_OFFSET`) inside the timer ISR. The previous commit (fbe9d9c) fixed this for `MAPPER` but not for these other locks. The clean fix is to move the reaper off-ISR into a dedicated kernel worker task — tracked in #147.

Practical risk on single-CPU today: narrow, since it requires the timer to fire while interrupted code is mid-`allocate_frame` / mid-`HHDM_OFFSET.lock()`. The `task_exit` integration test has run 64 spawn-exit cycles across merges without wedging. But it's real and deserves a systematic fix rather than a spot patch under time pressure.

Leaving this PR open for you to merge (or defer) at your discretion. Branch `m132-task-exit` is in place; no force-pushes expected.

Issue #123, opened 2026-04-13
Process table + fork/exec/wait lifecycle


dburkart:
## Motivation

Deferred from #20 (userspace ring-3 entry) and #121 (PID 1 init). Once a single user process is running, the next step is a real multi-process model: a kernel-side process table and the system calls that drive process lifecycle.

## Scope

- **Process table**: a kernel structure keyed by PID mapping to task state (address space, fd table, parent PID, exit status). Monotonic PID allocator.
- **`fork()`**: duplicate the calling process — clone address space (COW via #51), fd table, register state. Child gets PID N, returns 0; parent returns N.
- **`exec()`**: replace the calling process image with a new ELF (via the loader from #44). Flush user mappings, load new segments, reset stack, jump to new entry.
- **`exit()`** / **`_exit()`**: tear down user address space, release fds, record exit status, wake any `wait()`-ing parent.
- **`wait()`** / **`waitpid()`**: block until a child exits, collect its exit status, reap the zombie entry.
- **Reparenting**: when a process exits with live children, reparent them to PID 1.

## Out of scope

- `vfork`, `clone` flags beyond basic fork semantics (threads land in a separate issue).
- `/proc` or any process-introspection interface.
- Process groups, sessions, job control.

## Dependencies

- #20 (ring-3 entry + syscall path)
- #121 (PID 1 / init, establishes first process)
- #51 (demand paging + COW, required for fork address-space duplication)
- Per-process fd table issue (fd inheritance across fork/exec)

## Done when

- `fork()` + `exec()` round-trip works: parent forks, child execs a new ELF, parent waits and collects exit status.
- Integration test asserts serial output from both parent and child sides.

PR #148, opened 2026-04-13
kernel: ring-0 ELF loader for the userspace_hello module

dburkart:
Closes #44

## Summary
- Give `userspace_hello` its own linker script placing it at `0xffffffffc0000000` — 1 GiB above the kernel text at `0xffffffff80000000`, still in the upper half so a pre-`task::init()` mapping propagates through `new_task_pml4`'s upper-half copy to every later-spawned task. Override the workspace's kernel-linker rustflags via `RUSTFLAGS` env so the two binaries can share the `x86_64-unknown-none` target.
- Replace the payload's spin-loop with a COM1 write of `"USRHELLO\n"` so the smoke test can observe end-to-end control transfer.
- Add `kernel::mem::loader::load()` that walks `PT_LOAD` segments, allocates + maps 4 KiB pages with each segment's final flags, copies `p_filesz` bytes through the HHDM window, and zero-fills `memsz - filesz` for `.bss`. Widen `LoadSegment` with `p_offset` / `p_filesz` and bump visibility from `pub(super)` to `pub(crate)`.
- Snapshot the module's raw `&'static [u8]` in `mem::init` before `reclaim_bootloader_memory` runs (module bytes live in `EXECUTABLE_AND_MODULES`, preserved; Limine's response metadata does not).
- Wire `loader::load()` into `_start` between `mem::init` and `task::init`, stash the entry in a static `AtomicU64`, and spawn a trampoline task that calls into it (still ring-0 — ring-3 is deferred).

## Test plan
- [x] `cargo xtask build` — clean
- [x] `cargo xtask test` — 49 host unit + all QEMU integration tests pass, including a new `kernel/tests/userspace_loader.rs` that asserts upper-half mapping and byte-for-byte parity at the entry page
- [x] `cargo xtask smoke` — all 26 markers present, including the new `userspace: loader mapping image` and `USRHELLO` markers proving the trampoline jumped into the payload and the payload reached COM1
- [x] `cargo fmt --all --check` — clean
- [x] `cargo clippy -p xtask --all-targets -- -D warnings` — clean (matches CI scope)

---

> [!NOTE]
> **High Risk**
> Adds a new ELF loader that maps arbitrary module segments into the kernel address space and transfers control to the loaded entry point, which is paging- and memory-safety critical. Failures could corrupt page tables or execute unintended code if validation/mapping assumptions are wrong.
> 
> **Overview**
[...]

PR #152, opened 2026-04-13
RFC 0001: Userspace Virtual Memory Subsystem

dburkart:
## Summary

RFC 0001 proposes a per-process userspace virtual memory subsystem for vibix:
a three-layer model (`AddressSpace` → `Vma` → `VmObject`), a redesigned `#PF`
resolver with CoW fast-path, a small POSIX-shaped syscall surface
(`mmap`/`munmap`/`mprotect`/`brk`), and a `Flusher` TLB-invalidation
abstraction that is a no-op on UP today but preserves the API shape needed
for SMP shootdown.

will be reviewed by four archetype reviewers (security researcher, OS
engineer, user space staff engineer, academic), revised until all blocking
findings are addressed, then merged. Implementation issues will be filed
from the roadmap section after merge, unblocking #123 / #132 / #133 on the
P0 critical path to PID 1.

## RFC sections

- Abstract
- Motivation
- Background (vibix today, Linux/Serenity/Redox prior art, x86-64 hardware
  contract, POSIX surface)
- Design (data structures, fault dispatch, fork PTE-W-strip, syscall API,
  stack growth, address-space teardown)
- Security Considerations (SMEP/SMAP/UMIP, kernel-half protection,
  CoW races, refcount overflow, fresh-frame zeroing, MAP_FIXED hijack)
- Performance Considerations (fault hot-path, TLB batch threshold, fork
  cost, refcount memory cost)
- Alternatives Considered (7 rejected approaches with rationale)
- Open Questions (PCID, file-backed, threads, zero-page opt, guard gap)
- Implementation Roadmap (13 dependency-ordered work items)

## Test plan

- [ ] All four archetype reviewers post a review comment.
- [ ] All blocking findings are addressed in at most 2 defense cycles.

---

> [!NOTE]
> **Low Risk**
> Documentation-only change that adds an accepted RFC; no runtime or behavioral impact until implemented.
> 
> **Overview**
> Adds **RFC 0001** (`docs/RFC/0001-userspace-vm.md`), an accepted design document specifying a new userspace VM architecture and planned syscall surface.
> 
> The RFC details the proposed `AddressSpace`/`Vma`/`VmObject` model, page-fault and `fork()` CoW handling, TLB-flush abstraction, syscall semantics (`mmap`/`munmap`/`mprotect`/`brk`/`madvise`), and an implementation roadmap, but includes **no kernel code changes**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ad6f4ce5c29278c607f873fae3eea5ba3324e12. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/RFC/0001-userspace-vm.md
+++ b/docs/RFC/0001-userspace-vm.md
@@ -63,6 +63,21 @@ swap, NUMA, or huge pages — but it leaves the type lattice room for them.
   current task's `VmaList`, allocates a frame for `AnonZero`, copies for `Cow`.
 - `kernel/src/task/task.rs` — `NEXT_STACK_VA` bump allocator, no exit path.
 
+### Foundational prior art
+
+This RFC is a faithful re-implementation of the **Mach VM model** as
+laid out in Rashid et al., *"Machine-Independent Virtual Memory
+Management for Paged Uniprocessor and Multiprocessor Architectures"*
+(ASPLOS-II, 1987), which introduced the `vm_object` / `vm_map_entry`
+split that maps directly onto our `VmObject` / `Vma`. The per-frame
+refcount discipline follows Bonwick's slab/vmem treatment
+(*"The Slab Allocator"*, USENIX 1994; *"Magazines and Vmem"*,
+USENIX 2001). The `MAP_FIXED_NOREPLACE` rationale follows Linux commit
+`a4ff8e8620d3` (Michal Hocko, 2018). Nothing in this RFC claims novelty
+beyond combining these well-studied pieces in idiomatic Rust; the
+Academic reviewer's framing as "duplicate, not novel" is correct and
+deliberate (Academic advisory A5, A6).
+
 ### What other kernels do
 
 **Linux** (per `Documentation/mm/process_addrs.html`,
@@ -175,18 +190,34 @@ pub struct AddressSpace {
 ```
 
 `AddressSpace` is wrapped in `RwLock<AddressSpace>` (vibix's existing
-sync primitive). The lock is a single rwsem-equivalent — read for fault
-resolution, write for `mmap`/`munmap`/`mprotect`/`fork`. SMP per-VMA
-sequence locks (Linux ≥ 6.4) are explicitly out of scope; the API leaves
-room (see `Vma::lock_seq` placeholder below).
+sync primitive). Because every published syscall and every fault that
+resolves into the address space *mutates* PTEs (CoW) or VMA structure
+(mmap/munmap/mprotect/fork), the page-fault path takes the **write** lock
+unconditionally — the read lock exists only for future read-only
+introspection (`/proc/self/maps`-equivalents). This is simpler and
+closes the OS-Engineer/Security finding that the original pseudocode took
+`read()` and then mutated. SMP per-VMA sequence locks (Linux ≥ 6.4) are
+explicitly out of scope; the API leaves room (see `Vma::_lock_seq` below).
+
+**IRQ-safety contract.** The `AddressSpace` lock is **never** taken from
+interrupt context. Every call site (page-fault handler, syscall entry,
+fork/exec/exit) runs with interrupts enabled in a task context. We
+enforce this with a debug assertion (`debug_assert!(!in_irq())`) inside
+the lock's read/write entry points. Async-completion and timer paths that
+need to touch user memory must defer to a task-context worker; this
+matches vibix's existing interrupt-safety contract documented in
+`docs/interrupts.md`. This invariant is what lets the lock remain a
+plain `RwLock` rather than a spin-lock-with-IRQ-disable.
 
 **`Vma`** (replaces today's `kernel/src/mem/vma.rs`):
 
 ```rust
+#[repr(C)]
 pub struct Vma {
-    pub start: VirtAddr,         // page-aligned
-    pub end:   VirtAddr,         // page-aligned, exclusive
-    pub prot:  Prot,             // R/W/X bitset
+    pub start: VirtAddr,         // page-aligned (4 KiB)
+    pub end:   VirtAddr,         // page-aligned (4 KiB), exclusive
+    pub prot_user: Prot,         // PROT_* as the user requested (W/O preserved)
+    pub prot_pte:  Prot,         // effective bits installed in PTEs (W/O => RW)
     pub share: ShareMode,        // Private | Shared
     pub object: Arc<dyn VmObject>,
     pub object_offset: usize,    // bytes into the VmObject this VMA starts at
@@ -195,6 +226,27 @@ pub struct Vma {
     _lock_seq: u32,
 }
 ```
+
+`prot_user` records exactly what userspace asked for; `prot_pte` is the
+hardware-installable subset (where `PROT_WRITE` without `PROT_READ` is
+mapped to `R|W` because x86 cannot separate them — see User-Space
+finding B2). `mincore` / `/proc/self/maps`-equivalents report
+`prot_user`, so the round-trip is preserved.
+
+**Alignment rule (single, repo-wide).** Every byte address that crosses
+the syscall boundary is rounded as follows, applied uniformly to `mmap`,
+`munmap`, `mprotect`, `madvise`:
+
+- `addr` must be page-aligned exactly (no rounding). A non-aligned `addr`
+  on `mmap` without `MAP_FIXED` is rounded **down** to a page; with
+  `MAP_FIXED` it returns `EINVAL`.
+- `len` is rounded **up** to a whole page (`(len + 4095) & !4095`),
+  matching POSIX "entire pages containing any part of the range." `len`
+  is rejected with `EINVAL` only when `len == 0` or when the rounded
+  result overflows.
+
+This single rule keeps `VmaTree` intervals and PTE presence in lockstep
+across all four syscalls (Security finding B3).
 
 **`VmaTree`**: a `BTreeMap<usize, Vma>` keyed by `start` for O(log n)
 lookup, split, and merge. `find(addr)` does `range(..=addr).next_back()` and
@@ -242,31 +294,47 @@ mirrors Redox's `PageInfo`. Refcount widening from 16-bit can happen later;
 **Page-fault dispatch**, replacing `idt.rs:page_fault` (signature unchanged):
 
 ```
-on #PF (cr2, error_code, in_kernel_mode):
-  if in_kernel_mode and error_code.user_bit() == 0:
-    // existing kernel-fault path (panic / IST handling) unchanged
+on #PF (cr2, error_code, cpl):
+  // 1. SMAP first, unconditionally — kernel mode touching a user page
+  //    with AC=0 is a violation no matter what the VmaTree says.
+  if cpl == 0 and error_code.user_bit() == 1 and !rflags_ac():
+    panic_smap_violation(cr2)              // never returns
+
+  // 2. Pure kernel fault (CPL=0, supervisor page) → existing handler.
+  if cpl == 0 and error_code.user_bit() == 0:
     return existing_kernel_fault(...)
 
-  let aspace = current_task.address_space.read();
+  // 3. Reserved-bit fault is always a kernel bug. Scrub PTE physaddr
+  //    from any user-reachable channel; log the VA only.
+  if error_code.rsvd():
+    panic_rsvd_corruption(cr2)             // PTE word logged to klog only
+
+  // 4. cr2 must be canonical and below USER_VA_END before we touch any
+  //    user-controlled state.
+  if !cr2.is_canonical() or cr2 >= USER_VA_END:
+    return deliver_sigsegv(SI_MAPERR)
+
+  // 5. Take the WRITE lock — every fault that survives may mutate PTEs.
+  //    See "AddressSpace lock semantics" above.
+  let mut aspace = current_task.address_space.write();
   let vma = match aspace.find(cr2) {
     Some(v) => v,
-    None    => return deliver_sigsegv(SI_MAPERR),
+    None    => return resolve_growsdown_or_segv(&mut aspace, cr2),
   }
 
   let access = Access::from_error_code(error_code);
 
-  // Permission check first — protection violation never falls through to
-  // the resolver.
-  if !vma.prot.allows(access):
+  // 6. Permission check uses prot_user, not prot_pte.
+  if !vma.prot_user.allows(access):
     return deliver_sigsegv(SI_ACCERR)
 
-  // Reserved-bit fault is always a kernel bug.
-  if error_code.rsvd() { panic("PTE reserved-bit corruption at {cr2}") }
-
   match (error_code.present(), access.is_write()) {
-    (false, _)    => resolve_demand_page(&aspace, vma, cr2, access),
-    (true, true)  => resolve_cow(&aspace, vma, cr2),
-    (true, false) => spurious_fault_retry(),  // widening race; just return
+    (false, _)    => resolve_demand_page(&mut aspace, vma, cr2, access),
+    (true, true)  => resolve_cow(&mut aspace, vma, cr2),
+    (true, false) => return,  // widening race; IRET re-walks TLB. Safe
+                              // because x86 auto-reloads on PTE present-
+                              // bit set; we did the user mutation under
+                              // the write lock and INVLPG'd already.
   }
 ```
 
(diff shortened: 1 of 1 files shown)


dburkart:

**Summary:** The POSIX-shaped surface (`mmap`/`munmap`/`mprotect`/`brk`/`madvise`) is well-chosen and the errno table is close to conformant, but a few behaviors diverge from POSIX in ways that will bite real libc callers, and several details needed by C/Rust userspace are under-specified.

**Blocking findings:**
- [B1] `munmap`/`mprotect` returning `ENOMEM` for holes breaks POSIX — POSIX.1-2017 explicitly requires `munmap` to succeed over a range containing unmapped pages (it is not an error to unmap an address that is not mapped), and `mprotect` returns `ENOMEM` only when *addresses within* the range are invalid for the address space, not merely unmapped-within-a-reserved-window. The RFC says "range covers unmapped pages — POSIX-conformant strictness"; that is the opposite of POSIX. musl and glibc free() paths routinely `munmap` across holes and will crash if we return an error. Either match POSIX (succeed on holes for `munmap`; only `ENOMEM` on `mprotect` when a sub-page is literally unmapped, per Linux) or document this as an intentional, named exception with a rationale.
- [B2] `PROT_WRITE` without `PROT_READ` silently upgraded to RW — POSIX says the behavior of write-only is implementation-defined, but silently mutating user-provided `prot` bits breaks programs that later query their mapping (future `/proc/self/maps`, `mincore`, or a `mprotect`-readback test) and is surprising to Rust wrappers that round-trip the value. The x86 hardware constraint is real, but the kernel should either (a) accept the request and record prot as W while installing R|W in the PTE (report original prot to userspace), or (b) reject with `EINVAL` and document. Silent upgrade with no record is the worst option and is not idiomatic for either C or Rust callers.
- [B3] `mmap` on a regular `fd` returns `ENOTSUP` — POSIX mandates `mmap` on a file descriptor; returning `ENOTSUP` (not in the POSIX error list for `mmap`) will confuse portable code. Use `ENODEV` or `EACCES` as Linux does for unsupported backings, or explicitly gate this behind a documented "file-backed not yet implemented" exception in the RFC's POSIX-deviation section. Also, the flags table lists `MAP_SHARED` as returning `ENOTSUP` "without a backing object" — anonymous `MAP_SHARED` is a valid, widely-used POSIX pattern (shared memory between fork children); refusing it wholesale blocks a legitimate use case that the design otherwise supports.

**Advisory findings:**
- [A1] Syscall numbers `TBD` — before merging, pin the numbers in the table so userspace headers (and any Rust `vibix-sys` crate) have a stable contract. Churn here is expensive once PID 1 lands.
- [A2] No mention of `select`/`poll`/`epoll` composability — not directly in scope for VM, but `mmap` of eventfd/signalfd-like objects is a future concern; call out explicitly that the `VmObject` trait is the extension point so userspace devs know where this is going.
- [A3] `brk` "never sets errno" — fine, but document the failure mode: on a request past `brk_max`, glibc expects `brk` to return the *old* break unchanged (so `sbrk` can detect failure by comparing). Make that behavior explicit in the RFC.
- [A4] `madvise` accepts-all as no-op — POSIX-conformant, but `MADV_DONTNEED` has load-bearing semantics for jemalloc/mimalloc (zero-on-next-fault). Treating it as a no-op is silently wasteful of RSS, not incorrect. Worth a note in Open Questions.
- [A5] Error reporting ergonomics — confirm the syscall ABI returns `-errno` in `rax` (Linux convention) so Rust `nix`-style and C wrappers translate cleanly; the RFC does not state the return convention.
- [A6] `MAP_FIXED` logging at info-level on overwrite — good, but please also surface this via a counter or tracepoint so userspace debuggers (future `strace`-equivalent) can observe it without parsing kernel logs.
- [A7] `PROT_EXEC` with no NX handling detail — confirm that clearing `PROT_EXEC` on a page actually sets XD in the PTE, and that W^X is either enforced or documented as not-enforced. Userspace JITs will ask.

dburkart:
### Review: Academic

**Summary:** The three-layer AddressSpace/Vma/VmObject design is well-grounded in published work (Linux mm, SerenityOS, Redox) and the CoW + fork invariants match the canonical Bonwick/Bovet-McKinley treatment. No published result contradicts the safety or liveness claims as stated; the remaining concerns are advisory refinements around formal invariants.

**Blocking findings:** none

**Advisory findings:**
- [A1] CoW race under a single rwlock is under-specified — The text says "Resolution holds `aspace.write()`, which serialises resolution within an address space; cross-address-space CoW state is protected by the per-frame refcount's atomic decrement." This is correct only if the fault path actually takes the write lock when it mutates a PTE; the pseudocode earlier reads `aspace = current_task.address_space.read()`. The classic double-CoW race (Bonwick & Adams 2001; see also Clements et al., "The Scalable Commutativity Rule", SOSP'13, on why naive per-address-space locks don't serialise shared-frame mutations) requires that either (a) the resolver upgrades to a write lock before `resolve_cow`, or (b) PTE CAS + refcount fetch_sub form a linearisable pair. Please state which and justify. Using `Acquire` on `fetch_sub` but `Release` on the fork-side `fetch_add` is the standard pattern (cf. Boost/C++ memory model for refcounted shared_ptr, Boehm "Threads cannot be implemented as a library", PLDI'05) — confirm the Acquire on the *last* decrement is paired with an Acquire-fence before freeing.
- [A2] Deadlock-freedom of the aspace lock under fork — `fork_address_space` walks the parent's PTEs while constructing a child; if a concurrent fault on the *child* (impossible today, but reachable once threads land) takes `child.read()` and then touches `parent` via a shared VmObject's internal `BTreeMap`, the lock order parent→child must be globally fixed. Linux formalises this via `mmap_lock` ordering by `mm` pointer address; worth adopting that invariant now (it's free) so the future threading RFC doesn't have to retrofit it.
- [A3] Liveness of the single-sharer fast path — `wp_page_reuse` in Linux has a documented TOCTOU window where refcount may transiently read 1 while another CPU is mid-`get_page`; this is why Linux uses `page_count(page) == 1 && PageAnon(page) && !PageKsm(page)` plus a trylock on the page. Your fast path `if rc == 1: pte.set_writable(true)` after `fetch_sub(1)` is safe *because* the decrement already removed this sharer, but the commentary should state the invariant: "after `fetch_sub` returns old value 1, no other address space can observe the frame as shared" — which is true only because VMAs hold strong refs via the VmObject, not via page_refcount directly. Please make that ownership invariant explicit; it is the keystone of the correctness argument.
- [A4] Per-frame 16-bit refcount vs. unbounded sharing — Appel & Li ("Virtual memory primitives for user programs", ASPLOS-IV, 1991) and the Mach external-pager literature flag refcount saturation as a perennial bug class. Your fallback ("on overflow copy eagerly") is the correct safe choice; consider documenting it as a formal invariant ("refcount is either exact or over-approximated; never under-approximated") so future readers don't "fix" it into an AtomicU32 CAS loop that re-introduces an overflow bug.
- [A5] Prior art citation is thin on the seminal sources — The Background cites implementations but not the foundational papers that formalise exactly this design: Rashid et al., "Machine-Independent Virtual Memory Management for Paged Uniprocessor and Multiprocessor Architectures" (ASPLOS-II, 1987) for the VmObject/pager split; Young et al. on Mach's `vm_map` for the interval-tree + share-map pattern; and Bonwick's slab/vmem papers for the refcount discipline. One paragraph acknowledging that this is a faithful re-implementation of the Mach/Rashid model (not a novel design) would sharpen the novelty-vs-duplication framing and help future reviewers.
- [A6] MAP_FIXED_NOREPLACE as default is well-chosen — worth a one-line citation to the rationale in Linux commit `a4ff8e8620d3f` (Michal Hocko, 2018) which argues MAP_FIXED is a footgun for exactly the hijacking reason you note. Advisory only.

**Verdict:** LGTM

dburkart:

**Summary:** The design is thoughtful about the big-ticket hardware boundary (SMEP/SMAP/NXE, kernel-half PML4, zero-on-demand) but several concrete data-flow and concurrency details create exploitable primitives for unprivileged code — in particular, the CoW write-fault sequencing, a sub-page `end` boundary gap, and handling of non-canonical/kernel-half addresses at the syscall layer.

**Blocking findings:**
- [B1] CoW refcount/PTE race is not atomic under `aspace.read()` — The fault dispatch takes `aspace.read()`, then `resolve_cow` does `page_refcount.fetch_sub(1, Acquire)` and mutates the PTE. Security Considerations claims "Resolution holds `aspace.write()`", but the pseudocode explicitly grabs a read lock. If two threads of one process (future) or two CPUs (future SMP, which this RFC is preparing for) hit the same PTE, two writers can each observe `rc==1` at different moments, or one upgrades in place while the other is mid-copy, leaking the pre-copy frame's stale contents or corrupting a sibling process's private page. Please specify: (a) the lock that serializes PTE mutation against refcount transitions, (b) the exact atomic ordering (the `fetch_sub` should almost certainly be `AcqRel`, not `Acquire`), and (c) what prevents a concurrent `munmap`/`mprotect` writer from freeing `old_frame` between the `fetch_sub` and the `pte.set_frame`. Under the current text, a user can craft a multi-threaded/forking workload that races write faults to read kernel heap residue out of a not-yet-reused frame.
- [B2] `len` overflow and kernel-half VA acceptance in mmap/munmap/mprotect — The syscall table lists `(addr, len, prot, flags, ...)` and only calls out `len == 0 → EINVAL` and "unaligned addr under MAP_FIXED". Nothing requires `addr + len` to be checked for wraparound, for canonical-form compliance, or for strict confinement to the user half (`0..0x0000_8000_0000_0000`). A user passing `addr = 0x0000_7fff_ffff_f000, len = 0x0002_0000_0000_0000` with MAP_FIXED (or even without, into `find_free_range`) could drive the VmaTree or page-table walker into the kernel half, or wrap `end` below `start` and pass later `start < end` checks trivially. Same hazard on `munmap`/`mprotect` — a careless range walk over a wrap-around could traverse kernel PTEs and strip `W` or free kernel frames via `unmap_range`'s refcount path. Please mandate: reject `addr + len` overflow, reject any range that touches `>= USER_VA_END`, reject non-canonical `addr`, and require all three to return `EINVAL` before touching the VmaTree.
- [B3] `Vma.end` not required page-aligned at the API boundary — The struct comment says `end` is "page-aligned, exclusive", but the syscalls take raw byte `len`. If `len` is rounded up silently (common in POSIX impls), a user who mmaps `4097` bytes at a fixed address and then `munmap(addr, 4096)` can leave a 1-byte VMA whose `end` is the next page boundary — but if the rounding is inconsistent between `mmap` (round up) and `munmap` (round down/up), the user can desynchronize VmaTree state vs. PTEs and leave stale writable PTEs pointing at freed frames. Please specify the exact rounding rule for every syscall and prove it keeps VmaTree intervals and PTE presence in lockstep.
- [B4] MAP_FIXED silent eviction is privilege-escalation-adjacent via the loader — MAP_FIXED "silently evicts overlapping mappings" is fine for same-process self-harm, but the ELF loader (issue #148 already in tree) runs in ring 0 and consumes user-provided segment addresses. If the ring-0 loader uses MAP_FIXED over user-controlled PT_LOAD `p_vaddr` values without a canonical/user-half check, a malicious ELF can request segments overlapping the canonical kernel PML4 entries or the growsdown-stack guard. The RFC says "default new code to MAP_FIXED_NOREPLACE" — please make this a hard invariant for any in-kernel ELF loader (MAP_FIXED-as-syscall may remain, but kernel-initiated mappings from untrusted binaries must fail closed on overlap), and require the loader to reject any `p_vaddr` outside `[mmap_base, USER_VA_END - guard)`.
- [B5] `resolve_cow` does not validate that the faulting PTE is user-owned — The resolver is entered whenever `error_code.user_bit() == 1` and dispatches on `(present, write)`. A write fault at an address whose `vma.prot.allows(write)` is true but whose PTE was installed by the kernel with `U/S=0` (e.g., a bug elsewhere that mapped a kernel frame into a user VMA) will still execute `pte.set_writable(true)` and flush. Please require an explicit check that the resolved PTE's U/S bit is 1 before any CoW mutation, and panic otherwise — this catches kernel bugs before they become user-writable kernel memory.

**Advisory findings:**
- [A1] SMAP check ordering — "The fault handler must reject any #PF whose CPL=0, U/S=1, AC=0 as an SMAP violation before consulting VMAs" is correct but the pseudocode's first branch is `if in_kernel_mode and error_code.user_bit() == 0: existing_kernel_fault`. The SMAP case is `in_kernel_mode && user_bit == 1` and falls through into the user resolver path where `aspace.find(cr2)` runs on a kernel-mode fault. Please re-order so the SMAP classification happens first, unconditionally, and never consults the VmaTree with kernel CPL.
- [A2] Reserved-bit panic leaks kernel state to serial on a user-reachable path — `panic("PTE reserved-bit corruption at {cr2}")` includes the faulting VA and PTE word. If a user can induce a reserved-bit fault (e.g., by racing `mprotect` against hardware A/D bit updates on some uarches, or via a future bug), the panic string exposes kernel-mapped physical addresses on the boot console. Keep the panic but scrub the PTE physical address from any user-visible channel.
- [A3] `page_refcount` sized by "usable 4 KiB RAM" — The array must be sized by the maximum frame number the allocator can hand out, not by "usable" RAM, or a hot-plugged / late-discovered MMIO frame could index out of bounds. Also please specify behavior on frames backing MMIO/reserved regions: they must never enter the refcount path, or an accidental `fetch_sub` on a reserved frame's slot (initialised to 0) underflows to 0xFFFF and pins the frame forever.
- [A4] 16-bit refcount saturation fallback is under-specified — "on overflow falls back to copying the frame eagerly" during fork. Please specify: (a) is this atomic relative to concurrent CoW write-faults on the same frame? (b) does the eager copy happen in parent or child? (c) what is the behavior when `checked_add` fails mid-fork — is the whole fork rolled back, or is the address space partially copied? A partially-copied fork on ENOMEM is a classic kernel-state-corruption vector.
- [A5] `madvise` accepts-all no-op — Accepting `MADV_DONTNEED` as a no-op is POSIX-permissible but a user program that relies on DONTNEED for zeroing (common pattern) will observe stale data persisting in its own pages; not a kernel-security issue but a userspace-security footgun. Consider at least honoring `MADV_DONTNEED` for anonymous private mappings by dropping PTEs, or returning `ENOTSUP` rather than silently lying.
- [A6] Stack growth heuristic is attacker-controllable — The stack resolver extends `vma.start` downward on any fault within `stack_guard_gap` of `vma.start`. A user thread can intentionally fault at `vma.start - 255*PAGE_SIZE` to force a 255-page stack extension in one shot, bypassing any per-page rlimit-style check. Please bound per-fault extension to one page (Linux does), and enforce RLIMIT_STACK before extending.
- [A7] TLB flush on permission narrowing in fork — `fork_address_space` strips `W` in the parent's PTEs then uses `Flusher`. The RFC notes `Flusher::finish` does local INVLPGs on UP. Confirm the parent is flushed *before* any userspace return from the fork syscall; otherwise the parent can briefly observe stale W=1 TLB entries and write to now-shared frames, silently corrupting the child's view. This is stated correctly in the Background (Linux prior art) but the roadmap step "implement `fork_address_space`" should explicitly require the flush-before-return invariant.
- [A8] `brk_max` derivation — "e.g., `mmap_base - guard`" — please make the guard non-zero and document its size. A zero guard lets a user `brk` up to the first mmap VMA and then any subsequent `mmap` with a hint finds no room, turning into `ENOMEM` DoS of the process's own loader.

dburkart:
### Review: OS Engineer

**Summary:** The three-layer shape (AddressSpace / Vma / VmObject) and the Flusher abstraction are sound and future-proof for SMP, but the CoW fast path and the page-fault locking protocol as written contain concurrency bugs that will bite the moment threads or a second CPU exist — and are subtly wrong even on UP if an IRQ ever touches an address space. The refcount semantics also conflict with the frame allocator's "0 == free" invariant.

**Blocking findings:**
- [B1] CoW single-sharer fast path decrements refcount to 0 — `resolve_cow` does `fetch_sub(1, Acquire)` unconditionally and, when the prior value was 1, "upgrades to writable in-place." The refcount is now 0. That is the same sentinel the frame allocator / VmObject drop paths use to free the frame, so a concurrent `fork()` path that observes rc==0 (or a later unmap that decrements from 0 and underflows AtomicU16) will free a frame that is still live in a PTE. Fix: either CAS 1→1 (no-op) on the fast path, or treat rc==1 as the "reuse" trigger without decrementing (decrement only on the copy path, where we're dropping our reference to the old frame).
- [B2] Page-fault locking protocol is inconsistent — the dispatch pseudocode acquires `aspace.read()` at entry, but `resolve_cow` mutates PTEs and the Security section claims CoW "holds `aspace.write()`." `std`-style `RwLock` cannot upgrade in place; the handler will either deadlock or race. Pick one: (a) always take the write lock in the fault path (simple, fine on UP), or (b) retry with the write lock after classifying the fault. Document it in the RFC; this is the project's core locking invariant.
- [B3] IRQ-safety of `RwLock<AddressSpace>` not addressed — #PF arrives with interrupts in whatever state the faulting context had them (user faults: IF=1). If any IRQ handler, softirq-equivalent, or timer callback ever touches an address space (e.g., future async I/O completion zeroing a user buffer, RCU-style reclaim, accounting), the fault path deadlocks against itself. The RFC must either (a) declare `AddressSpace` is never touched from IRQ context and enforce it with a debug assert, or (b) make the lock IRQ-safe (disable interrupts while held). Given the stated "interrupt-safety contract," option (a) with an explicit assertion is the minimum.
- [B4] Refcount memory ordering is inverted for the standard Arc-style pattern — fork does `fetch_add(1, Release)` and CoW does `fetch_sub(1, Acquire)`. The canonical pattern is `fetch_add(1, Relaxed)` on share and `fetch_sub(1, Release)` on drop with an `Acquire` fence on the last-reference branch. More importantly, the "I'm the last sharer, take the frame" path needs an Acquire *fence* (or Acquire on the decrement *that observes 1*) to synchronize-with all prior writes to the frame from other CPUs. As written, writes by the previous owner are not ordered before the new exclusive owner's reads. This is latent on UP but a data race under SMP and must be specified now since the RFC explicitly claims cross-address-space CoW "is protected by the per-frame refcount's atomic decrement."
- [B5] Fork W-strip TLB coherence not specified for the parent — `fork_address_space` strips W from every parent PTE but the pseudocode only says "queue local INVLPGs (or full TLB flush)." If the parent is ever multi-threaded (the RFC keeps the door open via `Arc<AddressSpaceLock>`) or preempted onto another CPU under SMP, a stale writable TLB entry lets the parent bypass CoW and silently scribble on a now-shared frame — classic fork-CoW race. Thread the `Flusher` through `fork_address_space` and require it to cover every CPU in `aspace.used_by`, not just local. This is the SMP-shootdown use case the Flusher exists to support; wire it at the API level now even if the implementation is a local INVLPG today.

**Advisory findings:**
- [A1] `PageRefcount` static sizing — 2 MiB static array sized to a 4 GiB hardcoded cap is fine for today but will break the moment `BitmapFrameAllocator` grows past 4 GiB. Either tie the size to the same `MAX_PHYS_BYTES` constant the allocator uses (single source of truth), or allocate the table dynamically once the memory map is parsed.
- [A2] 16-bit refcount overflow fallback — "on overflow fall back to copying eagerly" is reasonable, but the copy requires walking every PTE that points at the frame across every address space, which the current data model (refcount only, no reverse map) cannot do. Either commit to an rmap-equivalent in this RFC or widen to `AtomicU32` now; the "eagerly copy" sentence as written is not implementable.
- [A3] SMAP check placement — "reject CPL=0 + U/S=1 + AC=0 before consulting VMAs" is correct, but the dispatch pseudocode routes all `user_bit==0` faults to `existing_kernel_fault` before the SMAP check. Make the SMAP classification explicit in the top-level switch so the contract is visible in code review, not just prose.
- [A4] `spurious_fault_retry` on (present=1, write=0) — correct on x86 (TLB auto-reloads on IRET), but please add a comment explaining *why* it's safe to simply return; future readers will assume this is a bug.
- [A5] BTreeMap vs intrusive tree — BTreeMap allocates on every insert/split. `mmap` under memory pressure can now fail with ENOMEM from the kernel heap, not just from frame exhaustion. Document that `mmap` must handle `AllocError` from `VmaTree::insert` and not leave the tree in a partial state on OOM (split-then-insert is two steps).
- [A6] `Vma::_lock_seq: u32` reserved field — fine, but add `#[repr(C)]` to `Vma` or commit to a layout so the future SMP per-VMA seqlock doesn't force an ABI break in on-disk/debug tooling. Minor.
- [A7] Cache-line layout of `PageRefcount` — 2 MiB of `AtomicU16` means 32 refcounts per cache line; adjacent frame refcounts will false-share hard under SMP fork storms. Consider padding to cache-line granularity for hot frames (the zero page, loader segments) or accept the contention and note it as a known cost.

dburkart:
## Defense Cycle 1 — response to blocking findings

All 13 blocking findings addressed in commit `7e550ec`. Summary:

### Security Researcher

- **[B1] CoW refcount/PTE race** — addressed. The page-fault dispatch now takes `aspace.write()` unconditionally; the lock-semantics paragraph above the data-structure section makes this explicit. The `resolve_cow` pseudocode no longer decrements on the fast path (see [B1] under OS Engineer below) and uses explicit Acquire-load on the fast path / Release-decrement + Acquire-fence on the copy path.
- **[B2] addr+len overflow / canonical / user-half** — addressed. New "Address validation" subsection in the K-U Interface section enumerates a 6-step check (`len==0`, rounding overflow, canonical, `addr+len` overflow, `<= USER_VA_END`, alignment) that runs before any state mutation in every mmap-family syscall.
- **[B3] Vma.end / len rounding rule** — addressed. New "Alignment rule (single, repo-wide)" subsection in the Vma section pins one rule (`addr` aligned exactly, `len` rounded up to a page) that applies uniformly to all four syscalls.
- **[B4] MAP_FIXED loader hard invariant** — addressed. New "MAP_FIXED loader invariant" paragraph in the Flags subsection requires kernel-initiated mappings (today's #148 ELF loader) to use `MAP_FIXED_NOREPLACE` and reject any p_vaddr outside the user half / canonical range.
- **[B5] CoW PTE U/S=1 verification** — addressed. The `resolve_cow` pseudocode now `assert!(pte.user())` before any mutation; the Security Considerations section explains the rationale.

### OS Engineer

- **[B1] Refcount sentinel collision** — addressed. The fast path `load(Acquire)`s the refcount and *peeks* at it; only the copy path decrements (because only the copy path drops our reference to the old frame). The frame allocator's "0 == free" sentinel is preserved.
- **[B2] Inconsistent locking protocol** — addressed. Page-fault path takes the write lock unconditionally; the lock-semantics paragraph documents this as the project's core locking invariant.
- **[B3] IRQ-safety contract** — addressed. New "IRQ-safety contract" paragraph in the AddressSpace section declares the lock is never taken from interrupt context, enforced with a `debug_assert!(!in_irq())` inside the lock entry points.
- **[B4] Refcount memory orderings** — addressed. Adopted the canonical Arc-style pattern: `fetch_add(1, Relaxed)` on share, `fetch_sub(1, Release)` on drop, `Acquire` fence before `frame_free` on the last-ref branch. The fast path's `load(Acquire)` synchronizes with all prior owners' writes.
- **[B5] Fork W-strip TLB shootdown** — addressed. `fork_address_space` now takes a `&mut Flusher` parameter. `flusher.invalidate(va)` is called for every parent W-stripped PTE; `flusher.finish()` is mandatory before `fork` returns to userspace, asserted on drop.

### User Space Staff Engineer

- **[B1] munmap/mprotect ENOMEM on holes** — addressed. `munmap` over a range with holes now returns 0 (POSIX-conformant). `mprotect` returns `ENOMEM` only on Linux's `mprotect_fixup` failure mode (sub-page literally unmapped), not on whole-VMA holes.
- **[B2] PROT_WRITE silent upgrade** — addressed. `Vma` now carries both `prot_user` (the user's request, preserved verbatim) and `prot_pte` (the hardware-installable subset). Userspace observability paths report `prot_user`; mprotect round-trips preserve the original bits.
- **[B3] mmap fd!=-1 errno + MAP_SHARED rejection** — addressed. `mmap` with `fd != -1` now returns `ENODEV` (POSIX-listed, matching Linux), not the made-up `ENOTSUP`. Anonymous `MAP_SHARED` is now fully supported as a first-class case (shared memory between fork children); only file-backed `MAP_SHARED` returns `ENODEV` until the file-backed VmObject lands.

### Out-of-scope classifications

None — all 13 blockers were addressed in the RFC text. Several Academic advisory items were also incorporated (lock-ordering invariant for future threads, Mach/Rashid + Bonwick + Hocko prior-art citations) since they were cheap and improve defensibility.

Re-review requested from Security Researcher, OS Engineer, and User Space Staff Engineer.

dburkart:

**Summary:** All five prior blocking findings are genuinely closed in commit 7e550ec — the fault path now takes `aspace.write()` unconditionally, the CoW fast path no longer decrements (preserving the allocator's rc==0 sentinel), syscall address validation is pinned as a 6-step check, the alignment rule is unified across mmap-family syscalls, and the CoW resolver asserts PTE U/S=1 before mutation. No new security issues introduced by the changes.

**Blocking findings:** none

**Advisory findings:**
- [A1] The `MAP_FIXED` loader invariant now forbids out-of-range `p_vaddr`, but the RFC text says the check "lives in the loader itself." Please also have the VM subsystem reject kernel-initiated `MAP_FIXED[_NOREPLACE]` calls whose range overlaps the kernel half / non-canonical region unconditionally, as a defense-in-depth second line (the loader is untrusted-input-facing code; the VM layer is the last gate). Non-blocking — current text already rejects out-of-range addresses "unconditionally regardless," but making the two-layer contract explicit in the roadmap would help.
- [A2] `panic_rsvd_corruption` now scrubs the PTE word from serial and logs only the VA to the boot console — good. Consider also documenting that the klog ring buffer is not mmap-exposable to any future userspace without a capability gate; otherwise the scrub is defeated by a future `/dev/kmsg`-equivalent. Non-blocking, forward-looking only.
- [A3] Anonymous `MAP_SHARED` is now first-class — worth a follow-up issue to specify whether sibling-fork shared-anon mappings survive `execve` in the child (Linux: yes, via the shared `VmObject`; vibix has no exec yet). Not a security issue today.

**Verdict:** LGTM

dburkart:
### Review: OS Engineer

**Summary:** All five prior OS-Engineer blockers are substantively resolved in `7e550ec`. The refcount/CoW invariants, lock/IRQ contract, memory ordering, and fork TLB discipline now read as a coherent, defensible design.

**Blocking findings:** none

**Advisory findings:**
- [A1] Fork increment ordering comment. The fork path uses `fetch_add(1, Relaxed)` (lines ~424, ~435), and an inline comment says "a future CoW resolver in the parent (which loads refcount with Acquire) must observe rc>=2 once it sees the W-stripped PTE." Strictly speaking, a `Relaxed` increment alone does not establish a happens-before edge with an `Acquire` load on another CPU; what actually carries the ordering here is (a) holding `aspace.write()` during the fork, and (b) the `Flusher`'s TLB-shootdown IPI + ack, which is a release point that publishes the W-stripped PTE and the refcount bump together. The "Memory ordering" paragraph (lines ~397-403) already states the release-acquire chain is carried by the CoW-path `Release` decrement and the last-ref `Acquire` fence — consider tightening the inline fork comment to point at the shootdown/ack as the publication point rather than implying `Relaxed` alone suffices. No behaviour change required.
- [A2] `Flusher::finish()`-on-drop assertion. The RFC says `fork_address_space` "asserts on drop that `flusher` was finished." Worth stating explicitly whether the drop-assert is `debug_assert!` (panics only in debug builds, as with the `in_irq()` check) or a hard `assert!`. A missed `finish()` on a release kernel is exactly the class of bug that leaves stale W=1 TLB entries on parent CPUs — I'd argue hard-assert here, matching Linux's `tlb_finish_mmu` discipline.
- [A3] `Flusher` batch cap. Section 7 says the flusher "buffers up to 32 invalidations. Above that, flush the whole TLB (`mov cr3, cr3`)." For `fork_address_space` of a large W-stripped parent this threshold will be hit near-instantly, effectively always degrading fork to a full flush. That is fine as a v1 choice, but worth calling out in the "Alternatives considered" or "Future work" section so it's not an unstated pessimism.
- [A4] Minor: `_lock_seq: u32` reserved field. Fine as a forward-compat placeholder, but consider a `#[cfg(not(smp))]`-gated `PhantomData` or a comment noting it is deliberately wasted UP space to avoid a struct-layout churn when SMP lands.

**Verdict:** LGTM

dburkart:

**Summary:** All three prior blockers (B1 munmap/mprotect hole semantics, B2 PROT_WRITE silent upgrade, B3 mmap file-backed errno + blanket MAP_SHARED rejection) are cleanly resolved in commit 7e550ec with POSIX-grounded text and the right data-structure split (`prot_user` vs `prot_pte`). A few small libc-ergonomics advisories remain; none block.

**Blocking findings:** none

**Advisory findings:**
- [A1] `brk` failure convention. The RFC states "On failure ... returns the current (unchanged) break. No errno set." That matches glibc/musl's `sbrk` detection pattern, but a userspace caller invoking the raw syscall cannot distinguish `brk(cur)` (legitimate query, returns cur) from a failed grow that clamped back to cur. Linux's raw `brk(2)` has the same wart, so this is defensible — recommend calling it out explicitly in the syscall table as "Linux-compatible ambiguity; use sbrk-style compare-against-prior".
- [A2] `madvise` accepts-all semantics. Only `MADV_DONTNEED` on `MAP_PRIVATE|MAP_ANONYMOUS` is semantically implemented; `MADV_FREE`, `MADV_REMOVE`, `MADV_DONTFORK`, etc. are silent no-ops with return 0. jemalloc/mimalloc in particular lean on `MADV_FREE` where available and fall back to `MADV_DONTNEED` only if `MADV_FREE` returns EINVAL. Consider either (a) advertising only the `_DONTNEED` advice and returning EINVAL for the rest so libcs probe correctly, or (b) documenting in the RFC that the silent no-op is intentional and libcs that feature-detect via EINVAL on unknown advice will fall back. Today's text implies the latter; make it explicit.
- [A3] `mmap` flags table lists `MAP_GROWSDOWN` and `MAP_STACK` but the errno table does not say whether unknown bits in `flags` are rejected. The text says "EINVAL ... unknown bits in `flags`" — good. Worth an explicit note that `MAP_GROWSDOWN` from userspace (not just kernel-initiated stack setup) is permitted, since Linux historically restricts it; pick a policy and state it.
- [A4] Syscall numbers pinned to Linux x86-64 values (mmap=9, mprotect=10, munmap=11, brk=12, madvise=28). Matches Linux, which is exactly what a future `vibix-sys` crate wants. Non-blocking, just flagging that `madvise=28` is Linux's value for `madvise` and preserves that symmetry.
- [A5] PROT_EXEC / W^X: explicitly not enforced, which is correct for a kernel that wants to run JITs. The "revisit if a hardening RFC later adds opt-in W^X" line is the right posture.

**Verdict:** LGTM

PR #211, opened 2026-04-14
syscall: implement brk(12) and set heap base from ELF image_end

### LINKED ISSUE #163: syscall: brk over a pre-reserved heap region in AddressSpace
## Motivation
Implements RFC 0001 — Userspace Virtual Memory Subsystem. glibc/musl `malloc` use `brk` for small allocations; without it the libc allocator falls back to `mmap` per allocation, which is functional but blocks several common static-binary builds.

## Work
- [ ] Wire syscall at `brk = 12` (Linux-compatible)
- [ ] At `AddressSpace::new_empty`, set `brk_start = brk_cur = brk_end_of_image`, `brk_max = mmap_base - 1MiB` guard
- [ ] `sys_brk(addr)`: if `addr == 0`, return current `brk_cur`; if `addr > brk_max` or alloc fails, return current `brk_cur` unchanged (sbrk failure-detection contract); otherwise extend or shrink the heap VMA
- [ ] Heap is a single anonymous private VMA backed by `AnonObject`
- [ ] No errno set on failure (return-current-break is the failure signal)
- [ ] Integration test: glibc/musl-style `malloc(small)` succeeds and underlying `brk` advances

## Context
- RFC: `docs/RFC/0001-userspace-vm.md` — Kernel–Userspace Interface (`brk`)
- Merged via PR #152
- Depends on the mmap/AddressSpace work landing first

dburkart:
Closes #163

## Summary
- `AddressSpace::sys_brk(addr)`: manages a single anonymous-private VMA `[brk_start, brk_cur)` backed by an unbounded `AnonObject`; grow path clones the existing Arc so already-faulted frames survive the remove+reinsert cycle; shrink path unmaps PTEs before reinserting a smaller VMA; `addr=0` probes the current break
- `AddressSpace::set_brk_start(image_end)`: called after `load_user_elf_with_vmas` in both `init_process::launch` and the execve syscall so the heap starts immediately after the ELF image rather than at the `mmap_base` placeholder
- `LoadedImage` gains an `image_end` field (page-aligned end of the last `PT_LOAD` segment) so callers don't need to re-parse the ELF
- Wires syscall `nr=12` (Linux-compatible `brk`) in `syscall_dispatch`

## Test plan
- [x] `cargo xtask build` — clean build
- [x] `cargo xtask test` — 124 host unit tests pass (2 new: `set_brk_start_advances_heap_base`, `brk_zero_returns_current_break`)
- [x] `cargo xtask smoke` — all 29 markers present ✓

---

> [!NOTE]
> **Medium Risk**
> Touches core userspace memory management (new `brk` syscall, VMA/PTE unmapping, and VmObject cache trimming), where bugs can cause leaks or address-space corruption. Scope is contained to heap management and ELF load bookkeeping with added bounds checks and tests.
> 
> **Overview**
> Implements Linux-compatible `brk(2)` by wiring syscall `12` to a new `AddressSpace::sys_brk`, which grows/shrinks the heap as a single anonymous-private VMA and updates mappings (including unmapping freed pages on shrink).
> 
> Adjusts process startup and `execve` so the heap base (`brk_start`) is set to immediately after the loaded ELF image. To support this, the ELF loader now returns an `image_end` (page-aligned end of the last `PT_LOAD`), and `VmObject`/`AnonObject` gain a trimming hook used to drop cached heap frames when shrinking.
> 
> Adds unit tests around `set_brk_start` and brk constructor invariants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a96692d17b9d9f56508542b0263a2fd209b1277. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -286,6 +286,13 @@ pub unsafe extern "C" fn syscall_dispatch(nr: u64, a0: u64, a1: u64, a2: u64) ->
             }
         }
 
+        // brk(addr) — set the program break; returns the new break on success
+        // or the current (unchanged) break on failure.
+        12 => {
+            let addr = a0;
+            crate::task::current_address_space().write().sys_brk(addr) as i64
+        }
+
         // fork() — clone the calling process; parent returns child PID, child returns 0.
         57 => {
             let user_rip = FORK_USER_RIP.load(Ordering::Relaxed);
@@ -339,6 +346,12 @@ pub unsafe extern "C" fn syscall_dispatch(nr: u64, a0: u64, a1: u64, a2: u64) ->
                 Ok(img) => img,
                 Err(_) => return -8, // ENOEXEC
             };
+
+            // Position the heap start immediately after the ELF image so
+            // sys_brk allocates from the right base address.
+            aspace
+                .write()
+                .set_brk_start(x86_64::VirtAddr::new(image.image_end));
 
             // Map a fresh user stack page and register it as a VMA so the
             // exec'd process's stack frame is also reclaimed on exit.
--- a/kernel/src/init_process.rs
+++ b/kernel/src/init_process.rs
@@ -63,6 +63,10 @@ pub fn launch(bytes: &[u8]) -> usize {
         image.segments
     );
 
+    // Set the heap base to immediately after the ELF image so sys_brk
+    // starts the heap at the right address.
+    aspace.set_brk_start(VirtAddr::new(image.image_end));
+
     // 3. Allocate and map the user stack, then add it as a VMA so fork
     //    copies it too.
     let stack_page = Page::<Size4KiB>::containing_address(VirtAddr::new(USER_STACK_PAGE_VA));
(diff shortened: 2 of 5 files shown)


dburkart:
Round 2 fixes pushed in 9a96692:
- `sys_brk`: validate exact addr against `brk_start` before page-rounding (prevents `brk_cur` being set below heap base)
- `sys_brk` shrink: decrement `rss_pages` for each unmapped PTE

Deferred: changing `set_brk_start` to return `Result` — architectural change, out of scope for this PR. The current silent no-op is documented and safe for trusted ELFs (filed as follow-up #212).

dburkart:
Cursor (04:52): "Missing TLB invalidation in brk shrink path" — false positive. `unmap_in_pml4` (paging.rs line 447) unconditionally calls `flush.flush()` (`invlpg`) on every unmap. The comment in that function explicitly explains this is required: "leaving them behind lets freed frames be reallocated while stale translations still point at them." No fix needed.

PR #223, opened 2026-04-14
tests: add RFC 0001 VMA integration coverage (#165)

### LINKED ISSUE #165: tests: integration coverage for VMA split/merge, fork CoW, mprotect, brk, stack growth
## Motivation
Implements RFC 0001 — Userspace Virtual Memory Subsystem. Each implementation issue ships its own targeted test, but the cross-cutting interactions (split-then-fork-then-CoW, mprotect during a stack growth, MAP_FIXED_NOREPLACE collision detection) need integration coverage before we trust the subsystem under PID 1.

## Work
- [ ] Test: `mmap(8 pages) + munmap(middle 4 pages)` produces two VMAs split at the boundaries
- [ ] Test: `mmap + mprotect(downgrade) + mprotect(restore)` re-merges into one VMA
- [ ] Test: fork-CoW divergence — write in parent does not affect child and vice versa; both pages survive after both writes
- [ ] Test: partial `munmap` over a range that spans a hole returns 0 (POSIX-conformant)
- [ ] Test: `mprotect` on a downgrade triggers TLB flush and subsequent write SIGSEGVs
- [ ] Test: `brk` grow then shrink; `brk(0)` returns current; `brk(past_max)` returns prior break
- [ ] Test: `MAP_FIXED_NOREPLACE` over an existing mapping returns EEXIST without disturbing the existing mapping
- [ ] Test: stack `MAP_GROWSDOWN` extends one page per fault; multi-page jump into guard SIGSEGVs
- [ ] Test: refcount saturation rollback during fork (synthetic — manually crank refcount to MAX before forking a small VMA)

## Context
- RFC: `docs/RFC/0001-userspace-vm.md` — Implementation Roadmap (final item)
- Merged via PR #152
- All other roadmap issues should land before this; this is the regression-catch net

dburkart:
Closes #165

## Summary

- Three new host unit tests in `kernel/src/mem/vmatree.rs` exercise the
  split/merge machinery: middle-unmap splits into two, mprotect
  downgrade-then-restore re-merges through a 3-way split, and unmap
  across a hole is a no-op on the gap.
- New QEMU integration binary `kernel/tests/vm_integration.rs` drives
  the full `AddressSpace`: `sys_brk` grow/shrink/query and past-max
  rejection, `MAP_GROWSDOWN` one-page-per-fault resolver, guard-gap
  SIGSEGV past 256 pages, and fork CoW data divergence (writes
  distinct sentinels through the HHDM since we cannot switch CR3 to
  reach the user VA from kernel test context).

## Out of scope

Three items from the planning pass are explicitly deferred and called
out in the test file header:

- `mprotect` downgrade -> write SIGSEGV needs a ring-3 userspace
  harness and the `#PF` test-hook dance.
- `MAP_FIXED_NOREPLACE` EEXIST: no syscall plumbing yet.
- Refcount saturation rollback in fork: the current fork path
  silently pins at `u16::MAX` rather than erroring, so there is no
  rollback to observe yet.

Follow-up issues will be filed for each post-merge.

## Test plan

- [ ] `cargo xtask test` passes (exercises the three new host unit
      tests and the new `vm_integration` QEMU binary).
- [ ] `cargo xtask smoke` passes.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -167,3 +167,7 @@ harness = false
 [[test]]
 name = "syscall_open_mmap"
 harness = false
+
+[[test]]
+name = "vm_integration"
+harness = false
--- a/kernel/src/mem/vmatree.rs
+++ b/kernel/src/mem/vmatree.rs
@@ -746,19 +746,20 @@ mod tests {
     }
 
     #[test]
-    fn unmap_range_spanning_hole_is_noop_on_gap() {
-        // Partial unmap over a range that spans a hole (i.e. some of the
-        // range has no VMA to unmap) must not panic and must leave
-        // surrounding VMAs intact in their original shape.
+    fn unmap_range_spanning_hole_trims_both_sides() {
+        // Partial unmap over a range that spans a hole (some of the
+        // range has no VMA to unmap) must not panic, must leave the
+        // hole alone, and must correctly trim the neighbouring VMAs
+        // whose edges the range does overlap.
         let obj = stub();
         let mut t = VmaTree::new();
-        t.insert(vma(0, K, RW, Arc::clone(&obj), 0));
-        t.insert(vma(4 * K, 5 * K, RW, Arc::clone(&obj), 0));
+        t.insert(vma(0, 2 * K, RW, Arc::clone(&obj), 0));
+        t.insert(vma(3 * K, 5 * K, RW, obj, 0));
         assert_eq!(t.len(), 2);
 
-        // Unmap across the gap; only the outer edges of the two VMAs
-        // are touched and neither extends into the gap, so both survive.
-        t.unmap_range(2 * K, 4 * K);
+        // Unmap [K, 4*K): the gap at [2*K, 3*K) is left alone; the
+        // left VMA trims to [0, K) and the right VMA trims to [4*K, 5*K).
+        t.unmap_range(K, 4 * K);
         assert_eq!(t.len(), 2);
         let mut it = t.iter();
         let a = it.next().unwrap();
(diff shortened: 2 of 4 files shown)


dburkart:
auto-engineer pausing on this PR.

**State:** 4 commits on the branch. `tests: add RFC 0001 VMA integration coverage` (396ba69), `fix: apply rustfmt` (2acfed1), `tests: open the brk window before driving sys_brk` (1513a04), and `ci: re-trigger workflow` (8dd5103).

**What tried:**
1. Round 1 (fmt): round-1 CI failed on `cargo fmt --check` only. Applied `cargo fmt --all`. Next CI run still failed — but on a real test assertion in `brk_grow_shrink_query`.
2. Round 2 (brk test logic): the assertion showed `sys_brk(0x40002000)` returned `0x40000000`, i.e. every grow was refused. Reading `AddressSpace::new_empty` shows it leaves `brk_start == mmap_base` and `brk_max == mmap_base - DEFAULT_BRK_GUARD` — an inverted heap window that rejects all grows. The loader normally closes this by calling `set_brk_start(image_end)`. Factored `brk_grow_shrink_query` and `brk_past_max_returns_prior` onto a shared `make_brk_aspace()` helper that calls `set_brk_start(VirtAddr::new(0x1000_0000))` first.
3. Round 2b: pushed 1513a04 and after ~4 min observed that the `ci.yml` workflow had not re-triggered (only CodeRabbit updated). Pushed an empty commit (`8dd5103`) to nudge — CodeRabbit repicked it up, but `ci.yml` still has not run.

**Why stopping:** fix-round 2 is the playbook cap, and the blocker now is CI infra, not code — the two most-recent branch pushes (`1513a04`, `8dd5103`) have no `ci.yml` run attached even though webhooks fired (CodeRabbit did re-run). The `concurrency: cancel-in-progress` group may have eaten them, or a workflow file change on main is pending re-approval. Either way, I can't verify green CI without a human re-running the workflow or confirming the brk test fix.

**Local verification:** `cargo check -p vibix --test vm_integration --target x86_64-unknown-none` compiles cleanly. The brk fix is mechanical — `set_brk_start(image_end)` is what every production caller does before driving `sys_brk`.

PR #247, opened 2026-04-14
tests: cover MAP_FIXED_NOREPLACE success case and wire mmap-family suite

### LINKED ISSUE #226: tests: MAP_FIXED_NOREPLACE must return EEXIST on overlap
## Motivation

Discovered while merging #223. RFC 0001 test coverage (#165) explicitly deferred this scenario because the `MAP_FIXED_NOREPLACE` syscall plumbing does not exist yet. Linux-compatible semantics require `mmap` with `MAP_FIXED_NOREPLACE` to fail with `EEXIST` (rather than silently clobbering) when the target range overlaps any existing VMA. Without a test, the flag could be implemented incorrectly and silently regress.

## Work

- [ ] Implement `MAP_FIXED_NOREPLACE` in the `mmap` syscall entry point (`kernel/src/arch/x86_64/syscall.rs`) — check against the `VmaTree` before inserting and return `-EEXIST` on any overlap.
- [ ] Add an integration test (`kernel/tests/syscall_open_mmap.rs` or `kernel/tests/vm_integration.rs`) that: (a) pre-maps a VMA, (b) calls `mmap` with `MAP_FIXED_NOREPLACE` over an overlapping range, (c) asserts the return is `-EEXIST` and the original VMA is untouched.
- [ ] Also cover the success case: `mmap` with `MAP_FIXED_NOREPLACE` at an unused address must succeed and install the VMA at exactly that address.

## Context

File paths:
- `kernel/src/arch/x86_64/syscall.rs` (mmap entry, merged via #214)
- `kernel/src/mem/vmatree.rs` (overlap logic)
- `kernel/tests/syscall_open_mmap.rs` (existing mmap tests after #214)

Related issues:
- #165 (parent umbrella — RFC 0001 test coverage)
- #162 (mmap syscall tracking)
- #204 (mmap flag validation)

dburkart:
Closes #226.

## Summary

- The `mmap` / `munmap` / `mprotect` / `madvise` syscall plumbing landed in #214 and #228, including `MAP_FIXED_NOREPLACE` and its EEXIST-on-overlap semantics. The existing `syscall_mmap_family` integration suite already exercises the overlap path, but the suite was never added to the xtask `test_all` list, so CI was not actually running it.
- Adds `syscall_mmap_family` to `xtask test_all` so all thirteen cases (including the overlap → EEXIST assertion) run under QEMU on every CI build.
- Adds a new success-case test `mmap_fixed_noreplace_unused_succeeds` covering the other half of issue #226: a `MAP_FIXED_NOREPLACE` at an unused VA must honour the requested address exactly, actually demand-fault, and tear down cleanly.
- Drops the now-stale "no syscall plumbing yet" bullet from `vm_integration.rs`.

## Test plan

- [x] `cargo build --package vibix --test syscall_mmap_family --target x86_64-unknown-none -Z build-std=core,alloc,compiler_builtins` — integration test compiles clean
- [x] `cargo build --package xtask` — xtask changes compile clean
- [ ] `cargo xtask test` — exercises the new case + existing overlap test under QEMU (runs on CI; the local dev environment is aarch64 and cannot run the x86_64 host-lib pre-pass)
- [ ] `cargo xtask smoke` — no behavior change expected (runs on CI)

### DIFF
--- a/kernel/tests/syscall_mmap_family.rs
+++ b/kernel/tests/syscall_mmap_family.rs
@@ -63,6 +63,10 @@ fn run_tests() {
             &(mmap_fixed_noreplace_overlap_eexist as fn()),
         ),
         (
+            "mmap_fixed_noreplace_unused_succeeds",
+            &(mmap_fixed_noreplace_unused_succeeds as fn()),
+        ),
+        (
             "munmap_returns_zero_on_hole",
             &(munmap_returns_zero_on_hole as fn()),
         ),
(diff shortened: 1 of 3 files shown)


dburkart:
Superseded by #246, which was opened first for the same issue. Closing this duplicate — auto-engineer didn't see #246 when picking #226.

PR #254, opened 2026-04-14
fork: refuse saturated page refcount (#227)

### LINKED ISSUE #227: mem: fork should error and roll back on VMA refcount saturation (u16::MAX pin)
## Motivation

Discovered while merging #223. RFC 0001 test coverage (issue #165) deferred a "refcount saturation rollback in fork" test because the current `fork_address_space` path silently pins the VMA refcount at `u16::MAX` rather than returning an error. That is a behavioural gap: when the refcount is already at `u16::MAX` and fork tries to bump it, no error is surfaced to the caller, and the child's view of the shared-object refcount silently diverges from the actual parent chain length. A process accumulating many forks of the same CoW page could hit this silently, then UAF on the `u16::MAX`-th drop.

## Work

- [ ] Change the VMA refcount bump in `kernel/src/mem/addrspace.rs::fork_address_space` (and any shared helper) from a saturating add to a checked add that returns `ForkError::RefcountSaturated` (or reuses an existing error variant) on overflow.
- [ ] Roll back any partial clone state on failure: VMAs already inserted into the child `VmaTree`, PTE entries already mapped, PML4 frames allocated, etc. The existing `OutOfMemory` rollback path should already have most of the machinery.
- [ ] Add an integration test (`kernel/tests/fork_cow.rs` or a new `kernel/tests/fork_refcount.rs`) that primes a VMA's refcount near `u16::MAX`, calls fork, and asserts the error propagates and the parent aspace is unchanged afterwards.

## Context

File paths:
- `kernel/src/mem/addrspace.rs` (`fork_address_space`, currently saturating)
- `kernel/src/mem/vmatree.rs` (`Vma` refcount field)
- `kernel/tests/fork_cow.rs` (existing fork coverage)

Related issues: #165 (parent umbrella).

dburkart:
Closes #227

## Summary
- Add `refcount::try_inc_refcount` — checked variant that returns `Err(Saturated)` instead of pinning at `u16::MAX`. The saturating `inc_refcount` is kept for callers that honour the over-approximation.
- `fork_address_space` now uses `try_inc_refcount` for both Private-child and Shared-child PTE acquisitions and surfaces saturation as a new `ForkError::RefcountSaturated`. The child is rolled back by `Drop for AddressSpace` just like the OOM path.
- Drop the parent-side put/inc pair during the W-strip: the parent reference stays attached to the same frame across the unmap/remap, so the pair was a net no-op in the success case and a third saturation hazard otherwise. Removing it simplifies the code and eliminates that hazard outright.

## Test plan
- [x] New `fork_refcount` integration test: primes the slot to `u16::MAX` and asserts `Err(RefcountSaturated)` for Private, Shared, and a parent-still-usable variant.
- [x] Existing `fork_cow` integration test unchanged (no-prefault, W-strip, no-leak, isolation, nested-CoW cases still covered).
- [x] `cargo fmt --all --check` clean.
- [x] `cargo xtask build` clean (objcopy step is a pre-existing local env quirk; CI is authoritative).

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -153,6 +153,10 @@ name = "fork_cow"
 harness = false
 
 [[test]]
+name = "fork_refcount"
+harness = false
+
+[[test]]
 name = "tlb_flusher"
 harness = false
 
--- a/kernel/src/mem/addrspace.rs
+++ b/kernel/src/mem/addrspace.rs
@@ -718,6 +718,12 @@ pub enum ForkError {
     /// allocation needed to build the child's PML4. The child is rolled
     /// back automatically before this error is returned.
     OutOfMemory,
+    /// A page's refcount was already pinned at `u16::MAX` when the fork
+    /// tried to acquire a new reference for the child PTE. Honouring the
+    /// fork would require silently under-counting (a UAF on the
+    /// `u16::MAX`-th drop), so we refuse instead. The child is rolled
+    /// back automatically before this error is returned.
+    RefcountSaturated,
 }
 
 /// Clone the address space for a child task (fork CoW).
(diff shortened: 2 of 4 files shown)


dburkart:
Thanks for both findings.

**Finding 1 (parent PTEs left W-stripped on late failure)** — fixed in 187601c. Now tracks every W-stripped parent page in a list and walks it on any error path to restore the original flags. Added `fork_late_saturation_restores_earlier_pte` to cover the regression.

**Finding 2 (unchecked `inc_refcount` in `AnonObject::fault` and `insert_existing_frame`)** — agreed, real and the same RFC 0001 invariant, but it's outside the scope of #227 (which is fork-specific). Filing a follow-up issue to convert those call sites and any other unchecked `inc_refcount` users.

Issue #233, opened 2026-04-14
vfs: add MountTable with mount and unmount helpers


dburkart:
## Motivation

Implements RFC 0002 — Virtual Filesystem Layer (item 5/15). `MountTable` + mount/unmount + `SbActiveGuard` are required before any FS driver can be mounted; the two-phase unmount protocol is load-bearing for memory safety under concurrent syscalls.

## Work

- [ ] `MOUNT_TABLE: BlockingRwLock<Vec<Arc<MountEdge>>>` — single global, linear-scan (≤ 8 mounts expected).
- [ ] `mount(source, target_path, fs_type, flags)` — single `MOUNT_TABLE.write()` critical section across both `mount-table.push` and `mountpoint.mount = Some(edge)` mutations (RFC §Mount and unmount).
- [ ] `unmount(target_path, flags)` — two-phase protocol:
    - Phase A (under `MOUNT_TABLE.write()`): `mountpoint.upgrade()` → set `draining=true` → check `sb_active==0` (or MNT_FORCE) → unlink edge from both `MOUNT_TABLE` and `mountpoint.mount`.
    - Phase B (no VFS locks): `gc_drain_for(&sb)` → `ops.unmount()`.
- [ ] `SbActiveGuard::try_acquire(&sb)` RAII helper with commit-then-validate (SeqCst `fetch_add` → load `draining` → rollback on race).
- [ ] `impl Drop for OpenFile` performs the matching `sb.sb_active.fetch_sub(1, SeqCst)`.
- [ ] Host-side unit tests over stub filesystems: nested mounts, upward `..` crossing, `EBUSY` on in-use unmount, concurrent unmount vs. in-flight path-based syscall.

## Context

- RFC: `docs/RFC/0002-virtual-filesystem.md` — §Mount and unmount, §Locking discipline — Unmount barrier for path-based syscalls.
- Depends on: #230, #231, #232.
- Merged PR: #224.
- Approximate size: ~250 LOC.

PR #259, opened 2026-04-14
vfs: deferred-eviction GC queue (RFC 0002 item 3/15)

### LINKED ISSUE #231: vfs: implement deferred-eviction GC queue
## Motivation

Implements RFC 0002 — Virtual Filesystem Layer (item 3/15). Closes the Drop-reentrancy hazard [OS-B3]: `Drop` can fire while sibling locks are held (e.g., replacing a parent's child-map entry drops the old `Arc<Dentry>`), so inline eviction would re-enter the lock stack. `Drop` must only enqueue; teardown runs at well-defined drain points.

## Work

- [ ] Add `vfs::gc_queue` — per-CPU, bounded ring + spinlock.
- [ ] `impl Drop for Dentry`/`Inode` pushes `(fs_id, ino)` onto the queue; does NOT call `evict_inode` inline.
- [ ] Drain the queue at syscall-return and in the idle task. Eviction runs with NO VFS locks held.
- [ ] `sys_umount` Phase B calls `vfs::gc_drain_for(&sb)` before `ops.unmount()` (RFC §Mount and unmount).
- [ ] Unit tests force nested drops under a held parent lock to prove no re-entry.

## Context

- RFC: `docs/RFC/0002-virtual-filesystem.md` — §Inode lifecycle and §Locking discipline — Deferred eviction.
- Depends on: #230 (core VFS types).
- Merged PR: #224.
- Approximate size: ~200 LOC.

dburkart:
Closes #231

## Summary
- Adds `vfs::gc_queue`, a single-global FIFO of pending inode evictions guarded by `spin::Mutex` (cannot block — `Drop` runs with sibling locks held).
- `impl Drop for Inode` now enqueues `(Weak<SuperBlock>, ino)` instead of calling `SuperOps::evict_inode` inline. Closes the [OS-B3] Drop-reentrancy hazard from the RFC.
- Drain hooks: `gc_drain()` for syscall-return / idle-task sites; `gc_drain_for(&Arc<SuperBlock>)` for `sys_umount` Phase B.

## Design notes
- **Single global queue** instead of per-CPU. Per-CPU was the RFC target, but the kernel has no per-CPU infrastructure today; building it for GC alone is scope creep. Documented in the module header.
- **Spill instead of drop** when past `GC_RING_CAP = 256`. Losing eviction work is worse than a transient `VecDeque` allocation; the overflow counter (`gc_overflow_count()`) makes pressure observable.
- **`Weak<SuperBlock>` on the entry**: avoids depending on the unbuilt `MOUNT_TABLE` (#233). A failed upgrade at drain time is treated as \"SB already gone, skip.\"
- **No `impl Drop for Dentry`**: the `Dentry`'s `Option<Arc<Inode>>` field is what holds the Inode; its plain drop already releases the Arc, and the Inode's own Drop is the single enqueue point.

## Test plan
- [ ] `cargo xtask test` (host unit tests + QEMU integration) on CI.
- [ ] New host tests (in `gc_queue.rs`):
  - FIFO push/pop ordering
  - `Drop for Inode` enqueues, does not call evict inline; explicit `gc_drain()` then triggers it
  - Nested drop under a held parent dentry's `children.write()` lock — would have deadlocked with inline eviction
  - `gc_drain_for` filters by `fs_id`
  - Overflow path past `GC_RING_CAP` keeps every entry (no loss)
  - Dropped `Weak<SuperBlock>` entries are silently discarded at drain

### DIFF
--- a/kernel/src/fs/vfs/gc_queue.rs
+++ b/kernel/src/fs/vfs/gc_queue.rs
@@ -1,0 +1,326 @@
+//! Deferred-eviction GC queue (RFC 0002 item 3/15).
+//!
+//! `Drop for Inode` enqueues `(sb, ino)` rather than calling
+//! `evict_inode` inline. Eviction runs at well-defined drain points
+//! (syscall-return, idle task, `sys_umount` Phase B) with no VFS locks
+//! held. This closes [OS-B3]: an inline `evict_inode` invoked from
+//! `Drop` could re-enter sibling locks (e.g. the parent dentry's
+//! `children` rwlock that triggered the Arc replacement).
+//!
+//! ## Single global queue
+//!
+//! RFC 0002 calls for a per-CPU ring; v1 uses one global
+//! [`spin::Mutex`]-protected queue. The kernel has no per-CPU
+//! infrastructure today and building it for GC alone is scope creep.
+//! Follow-up issue captures the per-CPU upgrade.
+//!
+//! ## Why `spin::Mutex` (not `BlockingMutex`)
+//!
+//! `Inode::drop` may run while sibling [`crate::sync::BlockingRwLock`]s
+//! are held; a sleeping primitive would risk re-entry into the
+//! scheduler with locks held. The queue lock is held only across O(1)
+//! push/pop, never across `evict_inode`.
+
+use alloc::collections::VecDeque;
+use alloc::sync::{Arc, Weak};
+use core::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+
+use spin::Mutex;
+
+use super::super_block::SuperBlock;
+use super::FsId;
+
+/// Soft cap on the inline ring portion of the queue. Past this we
+/// spill into a heap `VecDeque` rather than panic or leak — the cost
+/// of a transient allocation is preferable to losing eviction work.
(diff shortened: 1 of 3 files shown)


dburkart:
Round 2 fixes pushed in 59592ac:

- **CRITICAL (DRAINING write-only / `gc_drain_for` quiescence)** — addressed. Introduced a `DrainGuard` that compare-exchanges `DRAINING` with Acquire/Release ordering, and `gc_drain_for` now loops until no entries match `target` so the documented Phase B "queue is quiescent for this fs_id on return" contract actually holds when `evict_inode` re-enqueues. New test `drain_guard_excludes_concurrent_drainer` proves the flag is observed, not write-only.
- **MAJOR (`GC_RING_CAP` doc/impl mismatch)** — addressed. Updated the doc comment to describe the actual single-`VecDeque`-with-watermark behavior and call out the per-CPU-ring + spill design as deferred along with per-CPU infrastructure.
- **MINOR (test parallelism on shared statics)** — deferred to follow-up issue #263. Today `cargo xtask test` serialises these in practice; the follow-up tracks adding an explicit test-only mutex so a future runner change can't surface flakes.

Issue #161, opened 2026-04-13
task: AddressSpace::drop reclaim (frees VMAs, page-table frames, PML4)


dburkart:
## Motivation
Implements RFC 0001 — Userspace Virtual Memory Subsystem. Without proper teardown, every spawned task leaks its PML4 + intermediate page tables + every VMA-backed frame (today's behaviour, called out in #132). Required before any multi-process workload can run without OOM.

## Work
- [ ] Implement `Drop for AddressSpace`: walk every VMA, call `unmap_range(vma.start, vma.end)` (decrements refcount, frees on hit-zero)
- [ ] Walk the lower half of the PML4 (entries 0..256) freeing every intermediate page-table frame whose only remaining presence is the empty walk
- [ ] Free the PML4 frame itself last
- [ ] Asserts: every freed frame's `page_refcount` reads exactly 0 at reclaim
- [ ] In-kernel test: spawn a task with several VMAs, exit, observe frame allocator's free-frame count returns to baseline

## Context
- RFC: `docs/RFC/0001-userspace-vm.md` — Address-space teardown
- Merged via PR #152
- Closes the leak documented in #132 (or supersedes that issue's mem-side work; coordinate with #132 owner)

PR #262, opened 2026-04-14
tests: assert task::exit reclaims intermediate page-table frames

### LINKED ISSUE #146: task: reap intermediate (L3/L2/L1) page-table frames on exit
## Motivation
Discovered while merging #145. `task::exit` / `reap_pending` unmaps leaf VMA pages from the victim's PML4 and frees the PML4 frame itself, but the L3/L2/L1 intermediate page-table frames allocated by `map_in_pml4` under the lower-half tree are never returned to `KernelFrameAllocator`. The `x86_64` crate's `unmap` clears only the leaf PTE.

Practical impact today is zero — #145's tests don't install VMAs on exiting tasks, so no lower-half intermediates are allocated — but a future test or real userspace process with even one faulted VMA page will leak up to three intermediate frames per exit.

## Work
- [ ] After unmapping all leaf pages in the victim's PML4 lower half, walk the L4 entries (lower half only) and for each present entry, recursively free L3→L2→L1 frames that have no remaining present entries.
- [ ] Add coverage to `kernel/tests/task_exit.rs` that installs at least one `AnonZero` VMA, faults it in, exits, and asserts the free-frame count returns to baseline across many cycles.

## Context
- `kernel/src/task/mod.rs` — `reap_pending` in PR #145.
- `kernel/src/mem/paging.rs` — `free_pml4_frame`, `unmap_in_pml4`.
- Cursor Bugbot flagged this on commit 89270f4 of PR #145.

## Labels
priority:P2, area:mem, area:userspace, track:userspace

dburkart:
Closes #146

## Summary
- Add `exit_loop_no_intermediate_leak` to `kernel/tests/task_exit.rs`. The new worker installs an 8-page `AnonZero` VMA at `0x0000_3000_0000_0000` (an L4 entry the bootstrap PML4 leaves untouched), faults every page so the resolver allocates L3 / L2 / L1 intermediate page-table frames under that subtree, then calls `task::exit()`.
- Across 32 spawn-exit cycles the free-frame count returned by `vibix::mem::frame::free_frames()` must not drop below the post-warm-up baseline.
- The existing walker (`free_user_page_tables` in `kernel/src/mem/paging.rs`, wired through `Drop for AddressSpace`) already does the reclaim — `addrspace_drop.rs` (#161) covers it via the `Arc<RwLock<AddressSpace>>` drop. This PR adds the coverage #146 explicitly asked for: the `task::exit` → `reap_pending` → `Drop` path that userspace actually triggers.

## Test plan
- [x] `cargo xtask build` clean (only the pre-existing `is_guard_hit` warning).
- [x] `cargo fmt --all` clean.
- [ ] CI host tests + `cargo xtask test` (QEMU) execute the new `exit_loop_no_intermediate_leak` integration test on x86_64 — local env is aarch64 so this can only run in CI.
- [ ] CI smoke.

### DIFF
--- a/kernel/tests/task_exit.rs
+++ b/kernel/tests/task_exit.rs
@@ -9,13 +9,18 @@
 extern crate alloc;
 
 use core::panic::PanicInfo;
+use core::ptr;
 use core::sync::atomic::{AtomicUsize, Ordering};
 
+use vibix::mem::frame;
+use vibix::mem::vmatree::{Share, Vma};
+use vibix::mem::vmobject::AnonObject;
 use vibix::{
     exit_qemu, serial_println, task,
     test_harness::{test_panic_handler, Testable},
     QemuExitCode,
 };
+use x86_64::structures::paging::PageTableFlags;
 
 #[no_mangle]
 pub extern "C" fn _start() -> ! {
@@ -35,6 +40,10 @@ fn run_tests() {
     let tests: &[(&str, &dyn Testable)] = &[
         ("exit_removes_task", &(exit_removes_task as fn())),
         ("exit_loop_no_leak", &(exit_loop_no_leak as fn())),
+        (
+            "exit_loop_no_intermediate_leak",
+            &(exit_loop_no_intermediate_leak as fn()),
+        ),
     ];
     serial_println!("running {} tests", tests.len());
     for (name, t) in tests {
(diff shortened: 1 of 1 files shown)


dburkart:
Both findings are non-blocking nits — deferring:

- **Docstring coverage 16% on the test file**: integration tests don't usually carry per-fn docstrings; the test name + the module-level comment carry the intent. Worth a project-wide threshold conversation rather than padding this PR.
- **Redundant `assert_eq!` after `wait_for_fault_worker`**: the assertion is harmless and gives a clearer panic message tied to iteration `i`. Refactoring `wait_for_fault_worker` to thread `i` through is a valid alternative but adds boilerplate for marginal value.

Merging since CI is green and there are no actionable bugs.

Issue #240, opened 2026-04-14
vfs: wire fs::vfs::init and default mounts into boot


dburkart:
Implements RFC 0002 item 12/15. Wires the VFS into vibix::init() and sets up the three initial mounts so the kernel has a filesystem before PID 1 runs.

Work:
- Call fs::vfs::init() from vibix::init() after mem::init().
- Mount tarfs at / from the Limine ramdisk module (MountSource::RamdiskModule).
- Mount devfs at /dev.
- Mount ramfs at /tmp.
- Panic with a clear message if any initial mount fails (these are all fatal pre-PID-1).
- QEMU smoke: boot log shows all three mounts succeed.

Context: docs/RFC/0002-virtual-filesystem.md §Initialization order. Depends on #233, #234, #235, #236. Merged PR #224. ~100 LOC.

Issue #237, opened 2026-04-14
vfs: wire VfsBackend into FileDescTable and fork path


dburkart:
Implements RFC 0002 item 9/15. VfsBackend adapts Arc<OpenFile> into the existing FileBackend trait so the VFS plugs into kernel/src/fs/mod.rs::FileDescTable without breaking SerialBackend-based stdio.

Work:
- Add VfsBackend { open_file: Arc<OpenFile> } implementing FileBackend (read/write delegates to OpenFile.ops).
- FileDescTable::clone_for_fork preserves Arc<OpenFile> sharing so fork()'d children share open-file description (offset, flags) with the parent.
- Verify close_cloexec still works with VfsBackend-backed FileDescriptions.
- Integration test on host exercising the fd-table → VfsBackend → OpenFile → FileOps chain.

Context: docs/RFC/0002-virtual-filesystem.md §Integration with the existing fd table. Depends on #230, #233. Merged PR #224. ~150 LOC.

PR #269, opened 2026-04-14
vfs: read-only tarfs (USTAR reader) over a byte slice

### LINKED ISSUE #235: vfs: implement tarfs (USTAR reader) over a byte slice
## Motivation

Implements RFC 0002 — Virtual Filesystem Layer (item 7/15). `tarfs` is the read-only initrd format; mounted at `/` from the Limine ramdisk module. This is what lets PID 1 load `/init` off disk (completes the P0 chain together with #121).

## Work

- [ ] Add `kernel/src/fs/vfs/tarfs.rs` implementing a read-only `FileSystem`.
- [ ] USTAR parser over `&[u8]`: header checksum, type flag, 100-byte name, size, symlink target (100-byte linkname field).
- [ ] Build an in-memory directory tree at mount time; serve lookups from it.
- [ ] Regular-file read memcpys from the ramdisk slice (no page cache in v1).
- [ ] `MountSource::RamdiskModule(*const u8, usize)` variant for the Limine ramdisk bootstrap.
- [ ] Host-side unit test mounting a small handcrafted USTAR archive and verifying lookup, read, readlink, and getdents.

## Context

- RFC: `docs/RFC/0002-virtual-filesystem.md` — §Concrete filesystems for day 1.
- Depends on: #230, #232, #233.
- Blocks: #121 (PID 1 launches /init).
- Merged PR: #224.
- Approximate size: ~400 LOC.

dburkart:
Closes #235

## Summary

- New `kernel/src/fs/vfs/tarfs.rs` implementing RFC 0002 item 7/15: a read-only USTAR parser that builds an in-memory directory tree at mount time and serves lookups, reads, readlinks, and getdents out of it. No page cache, no writes.
- USTAR header checksum validation accepts both unsigned and signed interpretations (GNU tar historical bug compat). Handles typeflags ``0``/``NUL`` (regular), ``5`` (directory), and ``2`` (symlink); other types are skipped so pax/GNU-extension blocks don't fail the mount.
- Auto-creates intermediate directories when an archive entry's path prefix doesn't carry its own dir entry (common for tarballs produced with ``tar --no-recursion`` or minimal headers).
- Adds ``MountSource::RamdiskModule(*const u8, usize)`` so the boot wiring (#240) can hand tarfs a Limine ramdisk module directly.

## Test plan

- [x] ``cargo xtask build`` passes
- [x] Unit tests (host-side) cover: checksum round-trip, mount + lookup of a handcrafted archive with dir/reg/link entries, file reads at head/tail/EOF, readlink target, getdents enumeration, malformed-header rejection.
- [x] Zero local changes outside the three targeted files; existing VFS tests unaffected.
- [ ] Integration into boot + syscalls is deferred to #237/#238/#240 per the issue scope.

### DIFF
--- a/kernel/src/fs/vfs/mod.rs
+++ b/kernel/src/fs/vfs/mod.rs
@@ -53,6 +53,7 @@ pub mod open_file;
 pub mod ops;
 pub mod path_walk;
 pub mod super_block;
+pub mod tarfs;
 
 pub use dentry::{ChildState, DFlags, Dentry, MountEdge, MountFlags};
 pub use gc_queue::{gc_drain, gc_drain_for, gc_overflow_count, gc_pending_count};
@@ -67,6 +68,7 @@ pub use path_walk::{
     path_walk, Last, LookupFlags, MountResolver, NameIdata, NullMountResolver, Path, PATH_MAX,
 };
 pub use super_block::{SbActiveGuard, SbFlags, SuperBlock};
+pub use tarfs::TarFs;
 
 /// POSIX-imposed cap on a single path component.
 pub const NAME_MAX: usize = 255;
--- a/kernel/src/fs/vfs/ops.rs
+++ b/kernel/src/fs/vfs/ops.rs
@@ -19,7 +19,6 @@ use super::{Access, Credential, InodeKind, Timespec};
 /// Source of a mount operation. Separated from the target path so
 /// future sources (block device, ramdisk module, network URL) can be
 /// added without breaking the trait signature.
-#[derive(Debug)]
 pub enum MountSource<'a> {
     /// No backing — the FS synthesises its own storage (ramfs, devfs).
     None,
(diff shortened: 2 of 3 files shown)


dburkart:
Addressed the two actionable correctness findings:
- **Dirent padding** (Major): now zeroes each record with `entry.fill(0)` before `write_dirent` — closes the nondeterminism/info-leak path.
- **sentinel in f_files** (Minor): changed to `saturating_sub(1)` so the reported inode count excludes `nodes[0]`.

Deferring to follow-up issues:
- **`RamdiskModule` unsafe boundary** — redesigning the enum variant to require an `unsafe` constructor is an API-level change beyond this PR's scope.
- **`mounted` one-shot flag** — enforcing single-use outside the FS type is a mount-table concern; deferred to a separate issue.

PR #279, opened 2026-04-14
ci: increase smoke timeout and pin syscall ABI numbers (RFC 0002 / issue #278)

### LINKED ISSUE #278: ci: smoke fork+exec+wait markers missing on abi commit 1f60f1d — fragile init path
## Motivation

CI on main failed at commit \`1f60f1d\` ("abi: pin POSIX type widths for the Linux x86_64 syscall boundary", merged ~09:47 UTC 2026-04-14) with: \`smoke: missing markers ["hello: hello from execed child", "init: fork+exec+wait ok"]\`. The kernel booted to ring-3 and printed \`init: hello from pid 1\`, but the \`fork()\` → \`execve(hello)\` → \`wait4()\` sequence never completed within the 15s smoke window — suggesting the POSIX type width changes broke the syscall ABI for \`fork\`/\`execve\`/\`wait4\` on that commit. The failure resolved itself by the next merge (\`2a175a1\`), but the root cause was never diagnosed, leaving the init process lifecycle path without a dedicated regression guard.

## Work

- [ ] Bisect which specific type width change in \`1f60f1d\` caused \`fork\`/\`execve\`/\`wait4\` to silently fail under QEMU
- [ ] Add a QEMU integration test (\`tests/init_lifecycle.rs\` or similar) that directly exercises fork → exec → wait in userspace and fails fast with a clear error (rather than relying on the smoke 15s timeout)
- [ ] Consider whether the 15s smoke timeout is sufficient given CI runner speed — the abi run shows the kernel boot alone takes ~7–8s, leaving only ~7s for the full fork+exec+wait round-trip
- [ ] Document the syscall ABI invariants (type widths, argument packing) that \`userspace/init\` and \`userspace/hello\` depend on, so future ABI pinning PRs know what to test

## Context

- Failed run ID: 24392203205 (workflow: CI, branch: main, commit: \`1f60f1d\`)
- Failure output: \`init: hello from pid 1\` printed, then immediately \`smoke: missing markers [...]\` — no \`hello:\` or second \`init:\` line
- Fixed by: commit \`2a175a17\` (CoW resolver), but root cause unknown
- Smoke timeout logic: \`xtask/src/main.rs\` lines 729–732 (15s sleep, then kill QEMU)
- Smoke markers: \`xtask/src/main.rs\` lines 74–78
- Related: \`userspace/init\` and \`userspace/hello\` binaries implement the fork+exec+wait test

dburkart:
Closes #278

## Summary

- Increases the `cargo xtask smoke` timeout from 15 s to 30 s. The kernel boot alone takes ~7–8 s on CI runners, leaving only ~7 s for the `fork+exec+wait` round-trip — a window that intermittently caused the smoke markers to go missing (confirmed on commit `1f60f1d`, run 24392203205).
- Adds a `syscall_nr` module in `kernel/src/arch/x86_64/syscall.rs` with the pinned Linux x86_64 syscall numbers that `userspace/init` depends on (`fork=57`, `execve=59`, `wait4=61`, and the full set). A companion host unit test asserts every constant matches its expected value — a renumbering now produces an immediate compile-time/test failure rather than a silent smoke regression.
- Expands the `userspace/init` module-level doc with a pinned syscall number table and detailed invariants for `fork`/`execve`/`wait4`, so future ABI changes know exactly what to test.

## Test plan

- [x] `cargo xtask build` — compiles clean (1 pre-existing `is_guard_hit` dead_code warning, unrelated)
- [x] `cargo fmt --all -- --check` — passes
- [x] New host unit test `syscall_numbers_match_linux_x86_64_abi` in `syscall.rs` passes on CI (x86_64); validates all 20 pinned syscall numbers
- [x] Smoke timeout raised to 30 s — verified in `xtask/src/main.rs` line ~734

### DIFF
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -904,3 +904,75 @@ syscall_entry:
     fork_rflags = sym FORK_USER_RFLAGS,
     fork_rsp = sym FORK_USER_RSP,
 );
+
+/// Pinned Linux x86_64 syscall numbers used by `userspace/init`.
+///
+/// These are the numbers that appear in the `match nr` arms of
+/// `syscall_dispatch`. A mismatch between this table and the match arms
+/// will silently cause the wrong kernel operation to run (or -ENOSYS),
+/// which manifests in CI as missing smoke markers. See issue #278.
+pub mod syscall_nr {
+    pub const READ: u64 = 0;
+    pub const WRITE: u64 = 1;
+    pub const BRK: u64 = 12;
+    pub const FORK: u64 = 57;
+    pub const EXECVE: u64 = 59;
+    pub const EXIT: u64 = 60;
+    pub const WAIT4: u64 = 61;
+    pub const SIGACTION: u64 = 13;
+    pub const SIGPROCMASK: u64 = 14;
+    pub const KILL: u64 = 62;
+    pub const SIGRETURN: u64 = 15;
+    pub const MMAP: u64 = 9;
+    pub const MUNMAP: u64 = 11;
+    pub const MPROTECT: u64 = 10;
+    pub const MADVISE: u64 = 28;
+    pub const GETDENTS64: u64 = 217;
+    pub const OPEN: u64 = 2;
+    pub const OPENAT: u64 = 257;
+    pub const LSEEK: u64 = 8;
+    pub const CLOSE: u64 = 3;
+    pub const FSTAT: u64 = 5;
+}
+
+#[cfg(test)]
+mod tests {
+    use super::syscall_nr;
+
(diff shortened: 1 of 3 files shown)


dburkart:
Two CodeRabbit nitpicks deferred to follow-up issues:

1. **Use `syscall_nr::*` in dispatch arms** — replacing raw literals in `syscall_dispatch` with named constants is a good improvement but is refactoring scope beyond issue #278's intent (doc + timeout fix). Filed as a separate issue.

2. **Marker-driven early exit in smoke** — polling for markers and exiting early rather than always sleeping 30 s is a valid CI latency improvement, also deferred as a follow-up issue.

The actionable bug (wstatus bits 8..15 vs low 8 bits) is fixed in the latest commit.

PR #289, opened 2026-04-14
tests: end-to-end VFS read of /hello under QEMU (#241)

### LINKED ISSUE #241: vfs: add QEMU integration test reading /hello through the VFS
Implements RFC 0002 item 13/15. End-to-end integration test that proves the full VFS stack (tarfs + path walk + FileOps) works under QEMU.

Work:
- Add a small xtask (or extend existing) that builds a USTAR archive containing /hello with contents 'Hello\n'.
- Boot vibix under QEMU with that archive as the Limine ramdisk module.
- Kernel test code opens /hello through the VFS, reads the content, and panics (with a distinctive string) if it does not match 'Hello\n'.
- Wire into CI as a separate test job (similar to existing QEMU smoke tests).

Context: docs/RFC/0002-virtual-filesystem.md Implementation Roadmap. Depends on #235, #240. Merged PR #224. ~100 LOC + test harness.

dburkart:
Closes #241

## Summary

- Adds `kernel/tests/vfs_hello.rs`: a QEMU integration test that synthesises a single-file USTAR archive in-test, mounts it via `TarFs` + `MountSource::Static`, walks `/hello` through the real `path_walk`, and reads the payload through `FileOps::read`. Proves the full VFS stack end-to-end.
- Wires the new test into `kernel/Cargo.toml` (`harness = false`) and the `xtask` QEMU integration runner list.

## Why `MountSource::Static` and not `::RamdiskModule`?

The production `MountSource::RamdiskModule` path is covered by the boot-time mount in `vfs::init` (#240). It differs from `::Static` only in where the byte pointer comes from — the parse / Inode / Dentry / FileOps chain is identical. Using an in-test archive avoids pulling Limine module wiring, a new cargo feature, and a separate CI job for what is functionally the same exercise.

## Test plan

- [x] `cargo build --target x86_64-unknown-none -Zbuild-std=... --test vfs_hello -p vibix` compiles cleanly
- [ ] CI: new `vfs_hello` integration test passes under QEMU
- [ ] CI: existing integration suite still passes

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -187,3 +187,7 @@ harness = false
 [[test]]
 name = "vfs_backend"
 harness = false
+
+[[test]]
+name = "vfs_hello"
+harness = false
--- a/kernel/tests/vfs_hello.rs
+++ b/kernel/tests/vfs_hello.rs
@@ -167,8 +167,12 @@ fn hello_read_through_vfs() {
         .file_ops
         .read(&of, &mut buf, 0)
         .expect("FileOps::read /hello");
-    assert_eq!(n, 6, "must read exactly 6 bytes");
-    assert_eq!(&buf[..6], b"Hello\n", "contents must match");
+    if n != 6 || &buf[..6] != b"Hello\n" {
+        panic!(
+            "vfs_hello: expected /hello == b\"Hello\\n\", got n={n}, bytes={:?}",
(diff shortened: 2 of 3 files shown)


dburkart:
The  allowlist drift (CodeRabbit finding on ) pre-dates this PR. Fixing it properly — deriving the list dynamically from the manifest or  — would materially expand scope beyond wiring . Deferring to a follow-up issue.

Issue #258, opened 2026-04-14
ci: `cargo xtask smoke` 15 s fixed wait is flaky on CI runners


dburkart:
## Motivation

Discovered while merging #254 — the initial CI run on commit 187601c failed with `smoke: missing markers ["hello: hello from execed child", "init: fork+exec+wait ok"]`, but a re-run of the same commit passed. The smoke harness uses `std::thread::sleep(Duration::from_secs(15))` after spawning QEMU (`xtask/src/main.rs:733`) before killing the process and scanning serial output for markers. On slower GitHub Actions runners the boot + fork+exec+wait round-trip can exceed 15 s, producing a false negative that masks real regressions and wastes auto-engineer cycles chasing non-bugs.

## Work

- [ ] Replace the fixed 15 s sleep with a marker-driven wait: stream QEMU's serial output line-by-line, track which `SMOKE_MARKERS` have been seen, and exit as soon as all are present (or after a generous hard cap, e.g. 60 s).
- [ ] Keep the hard cap so a truly hung kernel still fails in bounded time.
- [ ] Verify on CI that a healthy run now finishes in boot-time + epsilon instead of always waiting 15 s.

## Context

- `xtask/src/main.rs:718-740` is the smoke harness body.
- `SMOKE_MARKERS` is defined at `xtask/src/main.rs:48`.
- Flaky run: https://github.com/dburkart/vibix/actions/runs/24393986587 (failed on first attempt, passed on re-run of the same commit).
- Related to #254 / issue #227.

PR #294, opened 2026-04-14
xtask: simplify initrd code — redundant comments, magic check, test coverage

dburkart:
## Summary

Rebased onto current main. The original first bullet (dropping redundant comments from `ustar_dir_block`) became moot after main's refactor introduced a shared `ustar_header_block` helper. The two remaining changes are still useful:

- Strengthen `ensure_initrd`'s idempotency check to verify the USTAR magic bytes (`"ustar\0"` at offset 257), matching the pattern in `ensure_test_disk`. Size-only wasn't sufficient to detect a corrupt or stale file of the right length.
- Extend `ustar_dir_block_checksum_valid` to assert the NUL+space terminator at bytes 154–155, fully verifying the 8-byte checksum field format rather than just the numeric value.

## Test plan

- [x] `cargo test --package xtask` — 23 tests pass
- [x] `cargo clippy --package xtask --all-targets -- -D warnings` — clean

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -533,9 +533,22 @@ fn ensure_initrd() -> R<PathBuf> {
     // DIRS headers + 2 file headers + 2 padded data blocks + 2 end blocks.
     const EXPECTED_SIZE: u64 = ((DIRS.len() + 2 + 2 + 2) * 512) as u64;
 
+    // Size-only isn't sufficient — a file of the right length but wrong content
+    // would silently produce a bad rootfs. Also verify the USTAR magic at offset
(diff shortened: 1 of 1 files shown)


dburkart:
auto-engineer paused: CI `cargo xtask smoke` failed with missing markers `hello: hello from execed child` and `init: fork+exec+wait ok`. This is the known flaky smoke timeout (issue #258) — the same failure occurred on main at fa7a1f4 before this PR was created. My changes only touch `ensure_initrd` in xtask and cannot affect kernel execution. Pausing per policy since this is not a fmt/clippy failure I can auto-fix. The PR is ready to merge once the smoke flake resolves or is re-run.

PR #295, opened 2026-04-14
vfs: mount tarfs-on-ramdisk-module at / when rootfs.tar module is present

### LINKED ISSUE #287: vfs: swap ramfs-at-/ for tarfs-on-ramdisk-module once a rootfs tarball ships
## Motivation

Discovered while merging #280. RFC 0002 specifies that `/` should be a tarfs backed by a Limine ramdisk module, but the boot ISO currently ships no rootfs tarball (`kernel/limine.conf` only has `userspace_init.elf` and `userspace_hello.elf`), so #280 mounts ramfs at `/` as an interim. Once a rootfs tar module exists, swap the backing so `/` becomes read-only content from the tarball.

## Work

- [ ] Add a rootfs tarball build step in xtask that packs a minimal rootfs (e.g. `/bin`, `/etc`) into a USTAR tar.
- [ ] Declare the tar as a Limine module in `kernel/limine.conf`.
- [ ] Change `kernel/src/fs/vfs/init.rs` to build `MountSource::RamdiskModule(ptr, len)` from the Limine module response and mount `TarFs` at `/` instead of `RamFs`.
- [ ] Keep `/dev` and `/tmp` as currently implemented (devfs and ramfs).
- [ ] Update or add a QEMU integration test that opens a known path from the tarball.

## Context

- `kernel/src/fs/vfs/init.rs` (the existing interim comment explicitly flags this: "Once a tar module lands, swap the `/` backing to TarFs with MountSource::RamdiskModule").
- `kernel/src/fs/vfs/tarfs.rs` (USTAR reader already exists from #269).
- `kernel/src/fs/vfs/ops.rs:MountSource::RamdiskModule` (the source variant is already defined).
- `kernel/limine.conf` (where the module entry needs to land).
- Related PR: #280 (the interim mount). Related RFC: RFC 0002 item 12/15.

dburkart:
Closes #287

## Summary

- **xtask**: adds `build_rootfs_tar()` that creates a minimal USTAR archive (`etc/hostname = "vibix\n"`) at `target/rootfs.tar` using `tar --format=ustar`
- **xtask `make_iso()`**: copies `rootfs.tar` into `boot/` so Limine delivers it as a module
- **`kernel/limine.conf`**: declares `boot():/boot/rootfs.tar` as a third Limine module alongside the existing ELF binaries
- **`vfs::init`**: on bare-metal (`#[cfg(target_os = "none")]`), calls `find_rootfs_module()` to locate the module by path suffix; if found, mounts `TarFs` at `/` via `MountSource::RamdiskModule`; falls back to `RamFs` when the module is absent (host tests are unaffected)
- **`kernel/tests/rootfs_module.rs`**: new QEMU integration test that boots the real kernel, walks `/etc/hostname` through the global VFS namespace via `path_walk`, and verifies the content reads back as `"vibix\n"`

## Why the cfg split?

`find_rootfs_module()` calls `crate::boot::MODULE_REQUEST.get_response()` which is `#[cfg(target_os = "none")]`-gated (Limine isn't present in host test builds). The host-test path always takes the `RamFs` fallback, so the five existing `init.rs` unit tests pass unchanged.

## Test plan

- [x] `cargo build --package vibix --bin vibix --target x86_64-unknown-none -Zbuild-std=...` compiles cleanly
- [x] `cargo build --package vibix --test rootfs_module --target x86_64-unknown-none -Zbuild-std=...` compiles cleanly
- [x] `cargo build --package xtask` compiles cleanly
- [ ] CI: `rootfs_module` QEMU integration test passes (boots real kernel, tarfs at /, reads `/etc/hostname`)
- [ ] CI: existing integration tests unaffected

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -191,3 +191,7 @@ harness = false
 [[test]]
 name = "vfs_hello"
 harness = false
+
+[[test]]
+name = "rootfs_module"
+harness = false
--- a/kernel/src/fs/vfs/init.rs
+++ b/kernel/src/fs/vfs/init.rs
@@ -9,10 +9,10 @@
 //! | `/dev`     | devfs  | synthesised                   |
 //! | `/tmp`     | ramfs  | synthesised                   |
 //!
-//! The RFC names tarfs-on-a-ramdisk-module as the long-term backing
-//! for `/`, but the boot ISO does not yet ship a rootfs tarball, so
-//! ramfs is mounted there for now. Once a tar module lands, swap the
-//! `/` backing to [`TarFs`] with `MountSource::RamdiskModule`.
+//! On bare-metal, if a `rootfs.tar` module is present in the Limine
+//! config, TarFs is mounted at `/` instead of RamFs. `/dev` and `/tmp`
+//! are always overlaid with DevFs/RamFs respectively; their mountpoint
+//! dentries use synthetic bootstrap inodes when the root FS is read-only.
 //!
 //! ## Bootstrap: where does the namespace root come from?
 //!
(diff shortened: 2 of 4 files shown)


dburkart:
CodeRabbit finding on `rootfs_module.rs`: The suggestion to walk `/etc/hostname` and assert `b"vibix\n"` is a good idea for a deeper end-to-end test, but `ensure_initrd()` (merged in #293) populates only directory stubs (`dev/`, `tmp/`, `bin/`, `etc/`) — no files. Extending the test to read file content would require also updating `ensure_initrd()`, which is out of scope for this PR. Filed as a follow-up.

PR #296, opened 2026-04-14
m85: VFS multi-level directory support

### LINKED ISSUE #85: VFS: multi-level directory support
## Motivation

Issue #43 scopes the ramdisk filesystem to a single directory level (flat namespace). Real userspace programs expect to navigate a directory tree (e.g., `/usr/bin/sh`, `/etc/init.d/`).

## Scope

- Extend the VFS `open` path to resolve multi-component paths (e.g., split on `/`, walk directory entries recursively).
- Update the on-disk/in-memory ramdisk layout to represent nested directories.
- Ensure the shell `ls` and `cat` built-ins work with paths like `/usr/bin/foo`.

## Out of scope here

- Symlinks, hard links, mount points.
- Directory creation at runtime (writable fs is a separate concern).

## Notes

Explicitly deferred from #43. The flat namespace is sufficient to demonstrate `cat /hello.txt`. Multi-level paths are needed once userspace grows beyond a handful of top-level files.

dburkart:
Closes #85

## Summary

- Extends `xtask ensure_initrd` to emit an `etc/init/` directory entry and an `etc/init/hello.txt` file (`"nested\n"`) in the rootfs USTAR tarball, so the initrd actually carries a multi-level path.
- Refactors the USTAR header builder in xtask into a shared `ustar_header_block` with thin `ustar_dir_block` / `ustar_file_block` wrappers, keeping the checksum math in one place.
- Extends `kernel/tests/vfs_hello.rs` with a second test case `nested_read_through_vfs` that mounts a USTAR archive with explicit `etc/` and `etc/init/` directory entries, walks `/etc/init/hello.txt` through `path_walk`, and asserts the contents.
- Adds a host unit test in `tarfs::tests` covering the implicit-parent branch of `split_parent` (a deep file entry with no explicit parent-dir headers).

The VFS path walker at `kernel/src/fs/vfs/path_walk.rs` already resolved multi-component paths — the gap closed by this PR was coverage: the rootfs had no nested entries and no integration test exercised more than one component. Per the issue, symlinks, hard links, mount points, and writable directory creation remain out of scope.

## Test plan

- [ ] `cargo xtask build` — kernel compiles
- [ ] `cargo xtask test` — host unit tests (including new `implicit_parent_dirs_are_auto_created`) and QEMU integration tests (including new `nested_read_through_vfs`) pass
- [ ] `cargo xtask smoke` — boot markers still present

(Local host is aarch64; CI exercises the full x86_64 toolchain.)

Verified locally:
- `cargo build -p xtask` succeeds.
- `cargo xtask initrd` produces a 4608-byte archive whose `tar -tvf` listing shows `dev/`, `tmp/`, `bin/`, `etc/`, `etc/init/`, `etc/init/hello.txt` (7 bytes) in order.
- Extracting the archive yields `etc/init/hello.txt` containing `nested\n`.

### DIFF
--- a/kernel/src/fs/vfs/tarfs.rs
+++ b/kernel/src/fs/vfs/tarfs.rs
@@ -969,6 +969,35 @@ mod tests {
             .expect("remount after unmount must succeed");
     }
 
+    /// A deep file entry whose intermediate directories are absent from
+    /// the archive must still be resolvable — `split_parent` auto-creates
+    /// the missing `a/` and `a/b/` directory nodes.
+    #[test]
+    fn implicit_parent_dirs_are_auto_created() {
+        let mut archive: Vec<u8> = Vec::new();
+        // Only a file entry; no explicit "a/" or "a/b/" headers.
+        archive.extend_from_slice(&make_header(b"a/b/c.txt", 3, b'0', b"", 0o644));
+        archive.extend_from_slice(&pad_block(b"hey"));
+        archive.extend_from_slice(&[0u8; BLOCK * 2]);
+
+        let fs = TarFs::new_arc();
+        let sb = fs
+            .mount(
+                MountSource::Static(leak_archive(archive)),
+                MountFlags::default(),
+            )
+            .expect("mount deep-file archive");
+
+        let root = sb.root.get().unwrap().clone();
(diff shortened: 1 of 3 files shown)


dburkart:
@coderabbitai review

PR #320, opened 2026-04-14
loader: reject PT_LOAD whose page-aligned end reaches USER_VA_END

### LINKED ISSUE #212: loader: image_end computation can produce non-canonical USER_VA_END address
## Motivation

The `image_end` calculation in `load_user_elf` / `load_user_elf_with_vmas` can produce exactly `USER_VA_END` (0x0000_8000_0000_0000) when a `PT_LOAD` segment ends right at the canonical boundary. Callers that pass `image_end` to `VirtAddr::new` (e.g. `set_brk_start`) would then panic. Currently only trusted ramdisk ELFs are loaded, limiting practical risk, but the loader-level invariant should be enforced regardless.

## Work

- [ ] In `loader.rs`, cap `image_end` at `USER_VA_END - 1` (or skip the brk setting when `image_end >= USER_VA_END`)
- [ ] Add a test ELF-builder fixture that produces a segment ending at the canonical boundary and verify `load_user_elf_with_vmas` returns `image_end < USER_VA_END`
- [ ] Update `set_brk_start` to handle the case gracefully (already done in PR #211 — just needs the loader-level fix)

## Context

- `kernel/src/mem/loader.rs` — `load_user_elf` and `load_user_elf_with_vmas` `image_end` computation
- `kernel/src/mem/addrspace.rs` — `set_brk_start`
- Discovered during Cursor Bugbot review of PR #211 (commit 6d14787)

dburkart:
Closes #212

## Summary

\`LoadedImage::image_end\` is eventually passed to \`VirtAddr::new\`
(\`kernel/src/init_process.rs:68\` and \`kernel/src/arch/x86_64/syscall.rs:500\`).
\`x86_64::VirtAddr::new\` panics on non-canonical inputs, and
\`USER_VA_END = 0x0000_8000_0000_0000\` is the first non-canonical
lower-half address. A PT_LOAD segment whose page-aligned end lands
exactly at \`USER_VA_END\` therefore turned a malformed user ELF
into a kernel panic.

Add \`LoadError::SegmentEndsAtUserVaEnd\` and check
\`seg_end >= USER_VA_END\` in \`load_user_elf\` **before** mapping the
segment, so the offending page is never installed in the per-process
PML4. Prior segments from the same ELF may already be mapped on this
error path — same contract as existing \`MapFailed\` /
\`SegmentNotLowerHalf\` branches; callers drop the partial PML4.

## Test plan

- [x] \`cargo build --target x86_64-unknown-none --test userspace_loader\` compiles clean
- [ ] CI's QEMU run of \`userspace_loader\` exercises the new
      \`load_user_elf_rejects_segment_ending_at_user_va_end\` case

### DIFF
--- a/kernel/src/mem/loader.rs
+++ b/kernel/src/mem/loader.rs
@@ -25,6 +25,7 @@ use x86_64::VirtAddr;
 
 use alloc::sync::Arc;
 
+use super::addrspace::USER_VA_END;
 use super::elf;
 use super::paging;
 use super::tlb::Flusher;
@@ -52,6 +53,12 @@ pub enum LoadError {
     /// Wraps the underlying `paging::map` error so callers can tell
     /// `FrameAllocationFailed` from `PageAlreadyMapped` etc.
     MapFailed(MapToError<Size4KiB>),
+    /// A `PT_LOAD` segment's page-aligned end lands at or past
+    /// `USER_VA_END` (`0x0000_8000_0000_0000`). `LoadedImage::image_end`
+    /// is eventually wrapped in `VirtAddr::new`, which panics on the
+    /// first non-canonical lower-half address. Reject at load time so
+    /// the panic can't be reached from a malformed user ELF.
+    SegmentEndsAtUserVaEnd,
 }
 
 /// Result of a successful load: the entry point, segment count, and the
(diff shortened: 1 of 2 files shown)


dburkart:
@coderabbitai review

PR #321, opened 2026-04-14
test(rootfs): assert /etc/hostname content end-to-end

### LINKED ISSUE #297: vfs: extend rootfs_module integration test to assert /etc/hostname content
## Motivation

The rootfs_module QEMU integration test (merged in #295) only checks that /etc is a directory — it doesn't read any file content. CodeRabbit flagged this gap in PR #295 review. To make the test actually verify the tar payload end-to-end, ensure_initrd() should write etc/hostname with content b"vibix\n", and the test should walk /etc/hostname and assert those bytes.

## Work

- [ ] In `xtask/src/main.rs`, update `ensure_initrd()` to also write `etc/hostname` with content `b"vibix\n"` inside the staging USTAR archive
- [ ] In `kernel/tests/rootfs_module.rs`, after verifying `/etc` is a `Dir`, add a second `path_walk` to `/etc/hostname`, open the inode for reading, and assert the content equals `b"vibix\n"`

## Context

- Test file: `kernel/tests/rootfs_module.rs`
- Initrd builder: `xtask/src/main.rs` — `ensure_initrd()` around line 524
- Related PR: #295 (squash SHA 23a6e03)
- Related issue: #287
- Discovered while merging #295; CodeRabbit flagged the missing content assertion in the PR review

dburkart:
Closes #297

## Summary

- Add `etc/hostname` (6 bytes: `vibix\n`) to the USTAR archive built by `xtask::ensure_initrd()`.
- Extend `kernel/tests/rootfs_module.rs` with a second test case `rootfs_etc_hostname_content` that path-walks to `/etc/hostname`, opens the inode through `FileOps`, and asserts the read-back bytes match.
- The existing `rootfs_etc_is_directory` case is preserved unchanged.

Completes the end-to-end assertion CodeRabbit flagged as a gap on PR #295: the test now exercises `boot module → TarFs::mount → SuperBlock → path_walk → Inode (Reg) → FileOps::read → bytes`, not just path resolution to a directory.

## Test plan

- [x] `cargo check --target x86_64-unknown-none -Z build-std=core,alloc,compiler_builtins --test rootfs_module -p vibix` succeeds.
- [x] `cargo xtask initrd` rebuilds `target/rootfs.tar` to 5632 bytes (11 × 512: 5 dirs + 2 file headers + 2 data blocks + 2 end blocks).
- [x] `tar xOf target/rootfs.tar etc/hostname | od -c` prints exactly `v i b i x \n`.
- [ ] CI runs `rootfs_module` under QEMU and both test cases pass.

### DIFF
--- a/kernel/tests/rootfs_module.rs
+++ b/kernel/tests/rootfs_module.rs
@@ -4,14 +4,13 @@
 //! After `vibix::init()`, `vfs::init::root()` should point to a TarFs root
 //! populated from the `rootfs.tar` module declared in `limine.conf`. This test
 //! walks `/etc` through the global VFS namespace and verifies it resolves to a
-//! directory, exercising the full
+//! directory, then opens `/etc/hostname` and reads back the payload bytes —
+//! proving the tar file content survives end-to-end through the full
 //!
 //!   boot module → TarFs::mount (MountSource::RamdiskModule) → SuperBlock
-//!     → path_walk("/etc") → Inode (Dir)
+//!     → path_walk("/etc/hostname") → Inode (Reg) → FileOps::read → bytes
 //!
-//! chain end-to-end. The initrd produced by `cargo xtask initrd` contains
-//! `etc/` as a directory stub (no files inside), so we assert directory kind
-//! rather than reading file content.
+//! chain.
 
 #![no_std]
 #![no_main]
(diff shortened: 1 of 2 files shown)


dburkart:
Good catch — 
running 5 tests
test tests::demangles_legacy_rust_symbol ... ok
test tests::passes_through_non_rust_symbol ... ok
test tests::ensure_initrd_creates_valid_archive ... ok
test tests::ustar_dir_block_checksum_valid ... ok
test tests::demangles_v0_rust_symbol ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s was failing locally on the stale assertions. Fix pushed in 2e35c8b: archive now asserts 11 * 512 and the zero-tail assertion uses a negative offset so it survives future layout changes.

Issue #60, opened 2026-04-12
Demangle Rust symbols in the embedded ksymtab


dburkart:
Follow-up from #39. The panic backtrace currently prints raw mangled names:

```
#0  0xffffffff8000c75b _RNvNtCs1LVtGmr8r9Y_4core9panicking9panic_fmt+0x2b
#1  0xffffffff800013a8 _start+0x338
```

Readable but noisy. We should run names through a demangler so frames land as `core::panicking::panic_fmt+0x2b`.

## Proposal

Demangle at build time in xtask, not in the kernel. `ksymtab::embed_ksymtab` already owns the blob format — have it feed each name through `rustc-demangle` (which has a host-std API) before serializing into the string table. Zero kernel code change, zero runtime cost, names are stable post-link anyway.

Demangled names are ~2-3× longer than mangled. Current usage is ~28 KiB of the 256 KiB reservation, so there's room; if it ever gets tight, we can drop the hash suffix (`rustc-demangle` can format without it).

## Why not in-kernel

`rustc-demangle` does support `no_std`, but:
- Adds code size for a feature that runs once per panic.
- Names are frozen at link time — no reason to defer the work to runtime.
- Keeps the kernel's `ksymtab` reader trivial (it already just emits UTF-8 bytes).

## Acceptance

- [ ] xtask adds `rustc-demangle` as a build-time dep.
- [ ] Each name passes through `rustc_demangle::demangle(..).to_string()` before going into the strtab.
- [ ] `cargo xtask run --panic-test` prints readable symbol names.
- [ ] No change to `kernel/src/ksymtab.rs` runtime parser.

dburkart:
@claude could we also translate addresses to lines & columns?

dburkart:
@claude move that technical analysis into it's own issue

PR #351, opened 2026-04-15
debug: embed address→file:line:col line table for backtraces

### LINKED ISSUE #62: Translate backtrace addresses to file/line/column
Follow-up from #60. Once symbol names are demangled, the next step toward readable backtraces is mapping each frame's PC to a source location — `file.rs:line:col`.

## Proposal

Do the work at build time in xtask (same philosophy as the demangling work), not at runtime.

### Option A — Build-time line table (recommended)

This is the natural continuation of the issue's philosophy ("do the work at link time, zero runtime cost").

1. Build the kernel with `debug = 1` (line tables only, no variable info — keeps DWARF small).
2. In xtask, after linking, parse `.debug_line` + `.debug_info` from the ELF using the [`gimli`](https://github.com/gimli-rs/gimli) crate (host-only, no `no_std` needed here).
3. For each symbol address range in the ksymtab, extract the line number program rows that fall in that range.
4. Serialize into a compact sorted table — call it `lntab`:
   ```
   [(pc: u64, file_idx: u32, line: u32, col: u16), ...]
   ```
   plus a filename string table (same layout as the existing strtab). Binary-search by PC at runtime.
5. Embed `lntab` in a new ELF section alongside the existing ksymtab blob.

Kernel side: a small binary search in the backtrace printer — no DWARF parser, no new crate. Result:
```
#0  core::panicking::panic_fmt at library/core/src/panicking.rs:100:5
#1  _start at kernel/src/main.rs:42:9
```

**Size**: typical Rust function has ~5–15 line-table rows; with ~300 symbols that's ≈3 000 entries × 16 bytes ≈ **48 KiB** — comfortably within the 256 KiB reservation. `debug = 1` adds no variable info bloat to the ELF.

**xtask new dep**: `gimli` (build-host only).

### Option B — Keep DWARF sections, query at runtime (`addr2line` crate)

The [`addr2line`](https://github.com/gimli-rs/addr2line) crate is `no_std`-compatible and would let the kernel walk DWARF directly during a panic. Complete coverage, nothing to pre-compute.

Downsides: `.debug_*` sections add **~200–400 KiB** to the kernel image, and the `addr2line` + `gimli` crates bring ~50 KiB of code that only runs once per panic. Overkill for a panic-only path.

## Why not in-kernel

Same reasoning as #60 for demangling — names (and addresses) are frozen at link time, no reason to defer the work to runtime.

## Acceptance

- [ ] Enable `debug = 1` for the dev profile (or the xtask build step).
- [ ] Add `gimli` to `xtask/Cargo.toml`.
- [ ] `embed_ksymtab` also emits a `lntab` section: sorted `(pc, file_idx, line, col)` array + filename strtab.
- [ ] Kernel backtrace printer binary-searches `lntab` for each frame's PC.
- [ ] `cargo xtask run --panic-test` prints `symbol at file.rs:line:col`.
- [ ] No change to `kernel/src/ksymtab.rs` runtime parser beyond the backtrace printer.

> This can share the same xtask code path as the demangling work in #60 — `embed_ksymtab` would build both tables in one ELF parse pass.

dburkart:
Closes #62

## Summary

Follow-up to #60. ksymtab already maps return addresses to symbol names; this extends the backtrace printer to also print source coordinates (`file.rs:line:col`) when line information is available.

- Philosophy matches #60: do the work at build time in xtask, not at runtime.
- Turn on `debug = 1` for the `vibix` package (line tables only, no variable info).
- Post-link, xtask parses `.debug_line` via `gimli` and patches a compact `(pc, file_off, file_len, line, col)` blob into a reserved `LNTAB_RESERVATION` section (mirrors the ksymtab reservation trick).
- `strip_debug` then drops the DWARF, so the shipped kernel carries only the 1 MiB blob plus a small binary search in the backtrace printer — no runtime DWARF parser, no `addr2line`.

On a debug build today: ~40k rows after dedup, ~988 KiB — fits comfortably in the 1 MiB reservation. If it overflows, xtask aborts with a clear `bump LNTAB_BYTES` error (same failure mode as ksymtab today).

### Kernel-side (`kernel/src/lntab.rs`, new)

- `LNTAB_RESERVATION` fixed at 1 MiB, non-zero magic forces SHT_PROGBITS.
- `parse_blob`, `resolve_line`, `is_populated`, `len`; layout pinned by `size_of::<Header>() == 20` / `size_of::<Entry>() == 24` asserts that xtask's constants must match.
- `ksymtab::format_addr` calls `lntab::resolve_line` and appends ` at file:line:col` when a row is present; falls through cleanly when no info is available.
- Module is visible under host `cfg(test)` so the decoder has unit tests.

### xtask-side (`xtask/src/lntab.rs`, new)

- `collect_rows` walks every unit's line program, `sort_and_dedup` collapses adjacent rows with identical source triples keeping the lowest pc, `build_blob` writes the on-wire format, `embed` locates the reservation symbol and overwrites its bytes in place.
- Filenames are normalized workspace-relative when possible.
- `gimli = "0.31"` added to xtask deps (host-only, `read-all` feature).
- `embed_lntab` runs *before* `strip_debug` in both `build()` and `test_runner()` — once the DWARF sections are gone there is nothing left to extract.

### Risks & mitigations

- **Inlined functions**: `.debug_line` lists only outer-frame rows, so inline callees don't get their own entry. Out of scope for v1 — the call-site file:line is still a big readability win.
- **Generic monomorphizations**: dedup by `(pc)` plus the `(file, line, col)` run collapse drops row count ~5–10× and keeps the strtab small via file-string dedup.
- **PCs outside any row**: binary-search miss returns `None`; `format_addr` omits the `at …` suffix. No panic path.
- **Image size**: `debug = 1` inflates the pre-strip ELF but `strip_debug` runs after `embed_lntab`, so the shipped image only grows by the embedded blob.

## Test plan

- [x] `cargo test -p xtask` — 23 passed (includes 6 new `lntab::tests::*`: sort/dedup ordering, duplicate pc handling, adjacent-same-source collapse, on-wire roundtrip, strtab dedup, u16 column saturation, workspace-prefix normalization).
- [x] `cargo clippy -p xtask -- -D warnings` — clean.
- [x] Standalone unit-test compile of `kernel/src/lntab.rs` — 9 passed (bad magic/version/truncated rejection, exact-pc hit, largest-pc-not-exceeding, before-first-None, past-last-returns-last, empty-table-None, 1000-row binary-search spot-check).
- [x] Full `cargo xtask build` path runs `embed_lntab` successfully on this branch and reports `→ lntab: 40104 rows, 987700/1048576 bytes`.
- [ ] CI: `cargo xtask build` + `cargo xtask test` + `cargo xtask smoke` on x86_64. Local aarch64 host blocks the full pipeline (objcopy lacks x86_64 ELF support), so the post-objcopy stages are gated on CI.
- [ ] Manual: `cargo xtask run --panic-test` on an x86_64 host should print `symbol at file.rs:line:col` — out of scope for CI, noted in the acceptance criteria.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,15 @@ panic = "abort"
 lto = true
 codegen-units = 1
 
+# Line tables only — xtask parses .debug_line post-link to emit an
+# embedded address→file:line:col table (issue #62), then strip_debug
+# drops the DWARF from the shipped image.
+[profile.dev.package.vibix]
+debug = 1
+
+[profile.release.package.vibix]
+debug = 1
+
 # panic="abort" for the test/bench profiles is enforced via the
 # `-Z panic-abort-tests` flag in .cargo/config.toml; setting it here
 # directly is ignored by cargo.
--- a/kernel/src/ksymtab.rs
+++ b/kernel/src/ksymtab.rs
@@ -193,9 +193,22 @@ pub fn len() -> usize {
 
 /// Format a single return address the way a human wants to read it:
 /// `0x<addr> <name>+0x<off>` when a symbol is found, else `0x<addr> ?`.
+/// If a line-table entry is available, append ` at <file>:<line>:<col>`.
 pub fn format_addr<W: Write>(w: &mut W, addr: u64) -> fmt::Result {
     match resolve(addr) {
-        Some((name, off)) => write!(w, "{addr:#018x} {name}+{off:#x}"),
+        Some((name, off)) => {
+            write!(w, "{addr:#018x} {name}+{off:#x}")?;
+            if let Some((file, line, col)) = crate::lntab::resolve_line(addr) {
+                if line != 0 {
+                    if col != 0 {
+                        write!(w, " at {file}:{line}:{col}")?;
+                    } else {
+                        write!(w, " at {file}:{line}")?;
+                    }
+                }
+            }
+            Ok(())
+        }
         None => write!(w, "{addr:#018x} ?"),
     }
 }
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -41,6 +41,7 @@ pub mod hpet;
 pub mod init_process;
 #[cfg(target_os = "none")]
 pub mod ksymtab;
+pub mod lntab;
 #[cfg(target_os = "none")]
 pub mod process;
 #[cfg(target_os = "none")]
--- a/kernel/src/lntab.rs
+++ b/kernel/src/lntab.rs
@@ -1,0 +1,319 @@
+//! Kernel line-number table embedded in the final ELF.
+//!
+//! Mirrors the shape of `ksymtab`: a fixed-size reservation patched in
+//! place by xtask after linking. xtask parses `.debug_line` from the
+//! kernel's own DWARF, emits a compact `(pc → file:line:col)` blob, and
+//! overwrites `LNTAB_RESERVATION`'s bytes. At runtime the backtrace
+//! formatter binary-searches the table to annotate each frame.
+//!
+//! ## Binary format
+//!
+//! ```text
+//! struct Header {
+//!     magic:     [u8; 4],   // b"LNTB"
+//!     version:   u8,        // 1
+//!     _reserved: [u8; 3],
+//!     count:     u32 LE,    // number of entries
+//!     str_off:   u32 LE,    // byte offset from blob start to strtab
+//!     str_len:   u32 LE,    // strtab length
+//! }
+//! struct Entry {
+//!     pc:       u64 LE,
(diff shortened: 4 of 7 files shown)


dburkart:
Thanks for the review. Addressed the empty-path issue in 7771ca4 — unresolvable file_idx entries now get skipped instead of serializing as empty strings.

The other two findings are deferred to a follow-up issue:
- Shortening non-workspace path strings: real concern, but the 2 MiB reservation (bumped in ff46c21) gives ~900 KiB headroom for now. Worth addressing before the table grows further.
- Keeping the last row for equal-PC duplicates: judgment call. Worth benchmarking against real DWARF data before picking a policy.

Issue #360, opened 2026-04-15
ramdisk: bundle prebuilt musl ld.so and add dynamic-binary smoke test


dburkart:
## Motivation

Follow-up of #127 (Dynamic ELF linking). Once PT_INTERP detection (#358) and auxv setup (#359) land, the final piece is including a real dynamic linker in the ramdisk and verifying end-to-end that a dynamically-linked musl binary runs. This is the third independently-landable chunk of dynamic linking support.

## Work

- [ ] Download or cross-compile a static musl `ld-musl-x86_64.so.1` (the dynamic linker binary) and commit it to `userspace/bin/` (or a new `userspace/lib/`) in the repo alongside the existing `userspace_init` binary.
- [ ] Update `xtask/src/main.rs` `ensure_initrd()` to include the interpreter at the path expected by the PT_INTERP segment (e.g. `/lib/ld-musl-x86_64.so.1`), consistent with what the test binary's PT_INTERP will say.
- [ ] Produce a dynamically-linked test ELF (a tiny musl C `hello world`, or a Rust binary linked against musl) and include it in the ramdisk as the `userspace_init.elf` entry.
- [ ] Add a smoke marker (or QEMU integration test) that boots with the dynamic binary and verifies serial output (e.g. `hello from dynamic init`), confirming ld.so ran and the binary executed to completion.

## Context

- `xtask/src/main.rs` — `ensure_initrd()` builds the ramdisk; currently bundles `userspace_init.elf` from `userspace/bin/`
- Issue #127 — parent issue
- Issue #358 — PT_INTERP detection (prerequisite)
- Issue #359 — auxv setup (prerequisite)

dburkart:
Comment: Should we bundle musl? Or roll our own linker in rust? If this is a one-way door, we should file a bug to track rolling our own rust-based dynamic linker in the future.

Issue #359, opened 2026-04-15
exec: write ELF auxv vector to initial user stack before ring-3 entry


dburkart:
## Motivation

The dynamic linker (ld.so) reads the auxiliary vector (auxv) from the initial stack to locate the main binary's program headers, the interpreter base, and other metadata. Without a correctly-populated auxv, ld.so cannot initialize the process. This is the second independently-landable chunk of dynamic linking support, after PT_INTERP detection (#358), and a follow-up of #127 (Dynamic ELF linking).

## Work

- [ ] In `kernel/src/init_process.rs` (or a new `kernel/src/exec.rs`), before calling `syscall::jump_to_ring3`, write the System V AMD64 initial stack layout to the user stack: argc, argv[], envp[], auxv[]
- [ ] Required auxv entries: AT_NULL (0), AT_PHDR, AT_PHENT, AT_PHNUM, AT_BASE (interpreter load base), AT_ENTRY (main binary entry point), AT_RANDOM (16 random bytes), AT_PAGESZ
- [ ] Update `syscall::jump_to_ring3` signature to accept the initial stack pointer (after writing argc/argv/envp/auxv), so the `rsp` register on ring-3 entry points to argc
- [ ] Adjust `USER_STACK_TOP` / stack allocation in `init_process.rs::launch()` to reserve enough space for the auxv region below the stack canary
- [ ] Integration test (or smoke marker): after entry, verify rsp points at argc=0 and that AT_PAGESZ=4096 is in the auxv table

## Context

- `kernel/src/init_process.rs` — `launch()` builds the address space; `init_ring3_entry()` calls `syscall::jump_to_ring3(entry, USER_STACK_TOP)`
- `kernel/src/arch/x86_64/syscall.rs` — `jump_to_ring3(entry: u64, stack_top: u64)`
- Issue #127 — parent issue (Dynamic ELF linking)
- Issue #358 — PT_INTERP detection (prerequisite; provides AT_BASE and AT_ENTRY values)
- System V AMD64 ABI §3.4 "Process Initialization" defines the initial stack layout

PR #361, opened 2026-04-15
elf: detect PT_INTERP segment and load interpreter at INTERP_LOAD_BASE

### LINKED ISSUE #358: elf: detect PT_INTERP segment and load interpreter into user address space
## Motivation

Follow-up of #127 (Dynamic ELF linking). Before a dynamically-linked binary can run, the kernel ELF loader must detect the PT_INTERP segment that names the dynamic linker, then load that interpreter ELF into the user address space in addition to the main binary. This is the first independently-landable chunk of dynamic linking support.

## Work

- [ ] Extend `kernel/src/mem/elf.rs` `ParsedElf` to expose the PT_INTERP segment (segment type 3): add a method like `interp_path()` that returns the null-terminated interpreter path string (e.g. `/lib/ld-musl-x86_64.so.1`).
- [ ] In `kernel/src/mem/loader.rs` `load_user_elf_with_vmas()`: after loading PT_LOAD segments of the main binary, check for PT_INTERP. If present, look up the interpreter path in the ramdisk/initrd, load its PT_LOAD segments into the user address space at a fixed base offset (e.g. `0x4000_0000` to avoid collision with the main binary at low addresses), and return the interpreter entry point and base alongside the main binary's metadata.
- [ ] Update `kernel/src/init_process.rs` `launch()` to pass interpreter base/entry back through `LoadedImage` so the auxv setup (next sub-issue) can use them.
- [ ] Unit test: ELF bytes with a PT_INTERP segment → `interp_path()` returns the correct path string.

## Context

- `kernel/src/mem/elf.rs` — ELF64 program-header walker (currently only handles PT_LOAD, type 1)
- `kernel/src/mem/loader.rs` — user ELF loader (`load_user_elf_with_vmas`); `LoadedImage` struct
- `kernel/src/init_process.rs` — calls `loader::load_user_elf_with_vmas`
- Issue #127 — parent issue (decomposed into sub-issues)

dburkart:
Closes #358

## Summary

- Add `PT_INTERP` and `ET_DYN` constants to `elf.rs`; store full image `bytes` in `ParsedElf` so the new `interp_path()` method can extract the null-terminated interpreter path string from the PT_INTERP segment's file content. Returns `None` if no PT_INTERP is present or the segment is malformed (no null terminator, out-of-bounds).
- Relax the `EntryOutsideLoadSegment` validation for `ET_DYN` binaries — a dynamic linker may have `e_entry == 0` because the OS supplies the load base at runtime.
- Add `module_bytes_for_path(suffix)` in `elf.rs` to look up any Limine module by path suffix; simplify the two existing per-module helpers to delegate to it.
- Extend `LoadedImage` with `interp_entry: Option<VirtAddr>` and `interp_base: Option<u64>` (both `None` for static binaries).
- Add `INTERP_LOAD_BASE = 0x4000_0000`, `LoadError::InterpNotFound`, and `LoadError::InterpLoadFailed`.
- Refactor `map_user_segment` to accept a `base_offset` parameter so ET_DYN segments (vaddr near 0) can be mapped at `INTERP_LOAD_BASE` without code duplication.
- Extend `load_user_elf_with_vmas`: after loading the main binary, check for PT_INTERP, look up the interpreter in Limine modules, load its PT_LOAD segments at `INTERP_LOAD_BASE` with VMA tracking, and record the adjusted interpreter entry and base in `LoadedImage`.
- Update `init_process.rs` and `syscall.rs` to use `interp_entry.unwrap_or(entry)` so the OS jumps to the dynamic linker rather than the main binary when a PT_INTERP is present.

## Test plan

- [x] `cargo xtask build` — kernel compiles cleanly for `x86_64-unknown-none`, no errors or new warnings
- [x] `cargo test --package xtask` — 23 xtask tests pass
- [x] `cargo clippy -p xtask --all-targets -- -D warnings` — clean
- New unit tests in `elf.rs`: `interp_path_returns_correct_string`, `interp_path_none_when_no_interp`, `interp_path_none_when_no_null_terminator` — verified compile; will run in CI on x86_64

### DIFF
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -561,7 +561,12 @@ pub fn exec_atomic(elf_bytes: &[u8]) -> Result<core::convert::Infallible, i64> {
     // interpreter's entry point; otherwise jump directly to the binary's entry.
     let effective_entry = image.interp_entry.unwrap_or(image.entry);
     // Never returns.
-    unsafe { jump_to_ring3(effective_entry.as_u64(), crate::init_process::USER_STACK_TOP) }
+    unsafe {
+        jump_to_ring3(
+            effective_entry.as_u64(),
+            crate::init_process::USER_STACK_TOP,
+        )
+    }
 }
 
 /// Public wrapper around `copy_path_from_user` for use by the VFS
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -557,8 +557,11 @@ pub fn exec_atomic(elf_bytes: &[u8]) -> Result<core::convert::Infallible, i64> {
     // Now safe to release the old PML4 + its frames.
     drop(old_aspace);
 
+    // If the binary has a dynamic interpreter (PT_INTERP), jump to the
+    // interpreter's entry point; otherwise jump directly to the binary's entry.
+    let effective_entry = image.interp_entry.unwrap_or(image.entry);
     // Never returns.
-    unsafe { jump_to_ring3(image.entry.as_u64(), crate::init_process::USER_STACK_TOP) }
+    unsafe { jump_to_ring3(effective_entry.as_u64(), crate::init_process::USER_STACK_TOP) }
 }
 
 /// Public wrapper around `copy_path_from_user` for use by the VFS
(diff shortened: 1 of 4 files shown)


dburkart:
Addressed findings 2 (basename PT_INTERP lookup) and 3 (interpreter cleanup on partial failure) in commit ceb87cd.

Finding 1 (initial user stack ABI layout for ld.so) is intentionally deferred to issue #359 — this PR's scope is PT_INTERP detection and interpreter loading only. The `interp_entry` path is wired through so the plumbing is in place, but a dynamically-linked binary won't actually run correctly until #359 (auxv setup) and #360 (ld.so in ramdisk) land. The 30-minute elapsed rule will apply for merge once CI is green.

PR #362, opened 2026-04-15
task: migrate SCHED from spin::Mutex to IrqLock

### LINKED ISSUE #313: sync: migrate SCHED to IrqLock once block_current/exit/wake sequencing is reworked
## Motivation

Discovered while merging #311 (initial IrqLock audit). The scheduler lock `SCHED` in `kernel/src/task/mod.rs` is the remaining high-value ISR-reachable static that was excluded from the IrqLock migration because migrating it naively would invert the interrupt-flag save/restore ordering that `block_current`, `exit`, and `wake` rely on around `context_switch` — those paths currently drop the `SCHED` guard BEFORE the context switch and manually restore IF AFTER, whereas `IrqLock` would restore on guard drop (before `context_switch`). Until `SCHED` is migrated, one of the highest-contention statics in the kernel still relies on convention (manual `interrupts::disable`/`enable` bookends) rather than being encoded in the type. Proceeding now-that-the-primitive-exists is the natural next step.

## Work

- [ ] Audit every `SCHED.lock()` / `SCHED.try_lock()` call site in `kernel/src/task/mod.rs` for the IF lifecycle it assumes (particularly `block_current`, `exit`, `wake`, `preempt_tick`, `spawn`).
- [ ] Decide whether `IrqLock` can be adopted as-is, or whether `SCHED` needs a bespoke wrapper that drops the lock before `context_switch` without restoring IF.
- [ ] Migrate `SCHED` to the chosen primitive; remove the hand-rolled `interrupts::disable()`/`enable()` bookends.
- [ ] Rerun the existing `task_exit` and `addrspace_drop` integration tests to confirm no regression.

## Context

- `kernel/src/task/mod.rs`: `SCHED: Lazy<Mutex<Scheduler>>` around line 65; call sites at approximately lines 241, 269, 304, 721, 1011, 1112.
- `kernel/src/sync/irqlock.rs`: the primitive merged in #311.
- Related issue: #49 (original "IrqLock primitive and tree-wide audit"), closed by #311.

dburkart:
Closes #313

## Summary

- Replace `static SCHED: Lazy<Mutex<Scheduler>>` with `Lazy<IrqLock<Scheduler>>` — the scheduler lock is ISR-reachable (via `preempt_tick`'s `try_lock`) and should encode that invariant in its type rather than relying on manual `interrupts::disable/enable` bookends at every call site.
- Remove the hand-rolled IF save/disable/restore patterns from four task-context helpers: `set_priority`, `adjust_nice`, `set_affinity`, and `wake`. The `IrqLock` guard lifecycle handles save-and-disable on acquire and restore on drop automatically.
- `exit` and `block_current` retain their pre-lock `interrupts::disable()` calls intentionally: both need IRQs masked across `context_switch`, which fires *after* the SCHED guard drops. The manual `disable()` before `lock()` ensures `IrqLock` saves `false` and restores `false` on drop, keeping IRQs masked through `context_switch`.
- `preempt_tick` uses `try_lock` from inside the timer ISR (IF already clear); `IrqLock::try_lock` saves and restores `false` — a no-op, preserving identical semantics.

## Test plan

- [x] `cargo xtask build` — kernel compiles without errors or new warnings
- [ ] Host unit tests (`cargo xtask test --lib`) — pre-existing failure due to `pic8259` not compiling on aarch64 CI host; not introduced by this PR (confirmed via `git stash` baseline check)
- [ ] CI smoke/integration on x86_64 runners

### DIFF
--- a/kernel/src/task/mod.rs
+++ b/kernel/src/task/mod.rs
@@ -43,6 +43,8 @@ use core::sync::atomic::Ordering;
 use spin::{Lazy, Mutex};
 use x86_64::instructions::interrupts;
 
+use crate::sync::IrqLock;
+
 use crate::sync::WaitQueue;
 
 pub use priority::{
@@ -62,7 +64,7 @@ use crate::time::TICK_MS;
 /// single PIT tick — every tick is a rescheduling opportunity.
 pub(crate) const DEFAULT_SLICE_MS: u32 = 10;
 
-static SCHED: Lazy<Mutex<Scheduler>> = Lazy::new(|| Mutex::new(Scheduler::new()));
+static SCHED: Lazy<IrqLock<Scheduler>> = Lazy::new(|| IrqLock::new(Scheduler::new()));
 
 /// Victims produced by [`exit`] waiting for the reaper task to reclaim
 /// them. Holds the `Box<Task>` so the task's stack, address space, and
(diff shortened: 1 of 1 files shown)


dburkart:
@coderabbitai review

PR #364, opened 2026-04-15
RFC 0003: Pipes, Poll, and TTY Line Discipline

dburkart:
## Summary

RFC 0003 proposes three tightly-coupled subsystems for POSIX-capable userspace: anonymous pipes + FIFOs, the `poll(2)` / `select(2)` readiness syscall, and an N_TTY-compatible line discipline with job-control signal generation. It also introduces one novel primitive — **deadline-aware poll readiness with kernel-side aggregation (DAPRA)** — that extends `poll` with a per-fd soft-deadline hint so the scheduler can batch wake-ups across a poll group within a latency budget.

This PR contains a design document only — no kernel code changes. The RFC will be reviewed by the scoped archetype panel (security researcher, OS engineer, user space staff engineer, academic, and performance engineer), revised until all blocking findings are addressed, then merged. Implementation issues will be filed from the roadmap section after merge.

## RFC sections

- Abstract
- Motivation
- Background (Linux, POSIX, Serenity/Redox, academic, x86-64 hardware)
- Design
  - Overview
  - Key Data Structures (Pipe, PollTable, Tty, NTty)
  - Algorithms and Protocols (including the novel DAPRA design)
  - Kernel–Userspace Interface (syscalls, ioctls, termios bits, errno table)
- Security Considerations
- Performance Considerations
- Alternatives Considered
- Open Questions
- Implementation Roadmap (13 items)

## Test plan

- [ ] All blocking findings are addressed within the scoped 2 defense-cycle cap.
- [ ] RFC frontmatter updated to \`status: Accepted\` before merge.

### DIFF
--- a/docs/RFC/0003-pipes-poll-tty.md
+++ b/docs/RFC/0003-pipes-poll-tty.md
@@ -1,7 +1,7 @@
 ---
 rfc: 0003
 title: Pipes, Poll, and TTY Line Discipline
-status: In Review
+status: Accepted
 created: 2026-04-15
 ---
 
@@ -1155,26 +1155,33 @@ would have leaked `ERESTARTSYS`.
 
 ## Open Questions
 
-1. **Pipe SPSC vs locked ring, revisited.** If initial benchmarking shows
-   lock contention is significant even in the SPSC case, swap the inner
-   mutex for true SPSC atomics. Defer until measurement.
-2. **PTY scope.** This RFC mentions `Pty{Master,Slave}` as TTY-driver
-   implementers but does not design `devpts` or the `/dev/ptmx` path.
-   File a follow-on RFC for PTY once the base TTY lands.
-3. **`sigaltstack` for SIGPIPE delivery.** vibix's signal delivery does
-   not yet support alternate stacks; SIGPIPE is delivered on the default
-   stack only. POSIX permits this; flag as an `area:security` follow-up.
-4. **Canonical VMIN/VTIME timer source.** MIN=0,TIME>0 needs a per-read
-   one-shot timer; we assume `hpet.rs` gives us ≤ 100 µs resolution.
-   Verify before implementing non-canonical reads.
-5. **DAPRA default deferral via libc.** Whether a future `libc.so` should
-   set a conservative default `deferral_ns` on normal `poll(3)` calls is
-   deferred to userspace policy. Kernel defaults to 0.
-6. **POLLRDBAND / POLLWRBAND.** Accept the flags (POSIX requires the
-   symbols) but never set them (no STREAMS priority bands). Documented
-   as a known POSIX-conforming no-op.
-7. **select exceptfds semantics.** Linux maps `exceptfds` to POLLPRI. We
-   do the same — documented here to avoid surprise.
+All questions below are **deferred to implementation** — they are measurement-
+or policy-driven and do not block the design. Implementation issues filed
+from the Implementation Roadmap below will pick them up as they arise.
+
+1. **Pipe SPSC vs locked ring, revisited.** *Deferred to implementation.* If
+   initial benchmarking shows lock contention in the SPSC case, swap the
+   inner mutex for true SPSC atomics. Measurement gate is the `pipe-bulk`
+   bench committed in §Performance Considerations.
+2. **PTY scope.** *Out of scope; future RFC.* This RFC mentions
+   `Pty{Master,Slave}` as TTY-driver implementers but does not design
+   `devpts` or the `/dev/ptmx` path. A follow-on RFC will cover PTY once
+   the base TTY lands.
+3. **`sigaltstack` for SIGPIPE delivery.** *Deferred to implementation.*
+   vibix's signal delivery does not yet support alternate stacks; SIGPIPE
+   is delivered on the default stack only. POSIX permits this; captured
+   as an `area:security` follow-up issue.
+4. **Canonical VMIN/VTIME timer source.** *Deferred to implementation.*
+   MIN=0,TIME>0 needs a per-read one-shot timer assuming `hpet.rs` gives
+   us ≤ 100 µs resolution. Verify during non-canonical read implementation.
+5. **DAPRA default deferral via libc.** *Deferred to userspace policy.*
+   Whether a future `libc.so` should set a conservative default
+   `deferral_ns` on `poll(3)` is deferred. Kernel defaults to 0.
+6. **POLLRDBAND / POLLWRBAND.** *Resolved — documented no-op.* Accept the
+   symbols (POSIX requires them) but never set them (no STREAMS priority
+   bands).
+7. **select exceptfds semantics.** *Resolved — Linux-compatible.* Linux
+   maps `exceptfds` to POLLPRI; we do the same.
 
 ## Implementation Roadmap
 
--- a/docs/RFC/0003-pipes-poll-tty.md
+++ b/docs/RFC/0003-pipes-poll-tty.md
@@ -1,7 +1,7 @@
 ---
 rfc: 0003
 title: Pipes, Poll, and TTY Line Discipline
-status: Draft
+status: In Review
 created: 2026-04-15
 ---
 
--- a/docs/RFC/0003-pipes-poll-tty.md
+++ b/docs/RFC/0003-pipes-poll-tty.md
@@ -1,0 +1,801 @@
+---
+rfc: 0003
+title: Pipes, Poll, and TTY Line Discipline
+status: Draft
+created: 2026-04-15
+---
+
+# RFC 0003: Pipes, Poll, and TTY Line Discipline
+
+## Abstract
+
+This RFC proposes three tightly-coupled subsystems required for a POSIX-capable
+userspace on vibix: anonymous pipes (`pipe(2)` / `pipe2(2)` / FIFOs), the
+level-triggered readiness syscall `poll(2)` (with `select(2)` layered on top), and
+an N_TTY-compatible line discipline that handles canonical-mode input, echo,
+termios flag processing, and job-control signal delivery. Together they are the
+minimum surface a shell, `sh`-style job pipeline, and an interactive editor
+(`vi`/`readline`) need in order to run on vibix. The RFC also introduces one
+novel primitive — **deadline-aware poll readiness with kernel-side
+aggregation** (DAPRA) — that extends `poll` with a per-fd soft-deadline hint so
+the scheduler can batch wake-ups across a poll group within a latency budget.
+
+## Motivation
+
+vibix today has a serial-backed stdio (`SerialBackend`) and a VFS with `devfs`,
+but has no mechanism for:
+
+1. **Inter-process byte streaming.** Shell pipelines (`ls | grep | wc`) require
+   an anonymous pipe whose read end and write end can be installed into
+   different fd tables after `fork`.
+2. **Multiplexed readiness.** Any non-trivial userspace (the shell's prompt
+   loop, a network daemon, a supervisor waiting on stdin + timer + child) needs
+   `poll`/`select` to wait on multiple descriptors. Without it, either every
+   descriptor must be made `O_NONBLOCK` + spun on, or the process blocks
+   against one fd and starves the rest.
+3. **Line-edited terminal input.** Today every read from `/dev/ttyS0` returns
+   raw bytes. A shell cannot implement Ctrl-C (SIGINT), Ctrl-Z (SIGTSTP),
+   backspace, or line commit without duplicating a canonical-mode state
+   machine in userspace — and even then it cannot correctly generate
+   job-control signals to its own process group, because that requires kernel
+   cooperation with the process table.
+
+Delivering all three in one RFC reflects their data-flow coupling: the TTY
+read buffer is a pipe-shaped producer/consumer queue whose producer is the
+line discipline; `poll` is the unified readiness mechanism across pipes, TTYs,
+and eventual socket fds. Designing them together prevents the classic trap of
+shipping pipes and poll with a waitqueue convention that the TTY can't reuse.
+
+## Background
+
+### Prior art in vibix
+
+- **`FileBackend` trait** (`kernel/src/fs/mod.rs:30-42`) already defines a
+  per-fd `read`/`write` dispatch. We extend it with a `poll` method and
+  plumb a line-discipline layer between the existing PS/2/serial input path
+  and the fd-readable buffer.
+- **RFC 0002 VFS** (`docs/RFC/0002-virtual-filesystem.md`) established `devfs`
+  as the mount point for character-device fds. Pipe file descriptors are *not*
+  VFS-backed (they have no path); they are first-class `FileBackend`
+  implementations alongside `SerialBackend`.
+- **Signal delivery** (`kernel/src/signal/mod.rs`) gives us per-process
+  signal queues. This RFC consumes but does not extend that API.
+
+### Linux reference points
+
+- `fs/pipe.c` — `struct pipe_inode_info` with a power-of-two ring of
+  `struct pipe_buffer` slots, `PIPE_DEF_BUFFERS = 16` (`include/linux/pipe_fs_i.h:5`),
+  `PIPE_BUF = 4096` (`include/uapi/linux/limits.h`). Atomic writes below
+  `PIPE_BUF` are **implemented by buffer merging**, not by locking: if the
+  tail buffer has `PIPE_BUF_FLAG_CAN_MERGE` set and offset+chars ≤ PAGE_SIZE,
+  the writer appends into the existing slot. Packet mode (`O_DIRECT`) clears
+  the merge flag so each write becomes its own slot (`fs/pipe.c:415-418`).
+- `fs/select.c` — the `poll_table_struct { _qproc, _key }` (`include/linux/poll.h:37-40`)
+  and the two-pass pattern inside each driver's `.poll`: register wait
+  callback first, *then* `READ_ONCE` the state, so any concurrent state
+  transition either (a) sets the state before the load and is seen, or
+  (b) fires the registered callback.
+- `fs/eventpoll.c` — epoll's `ep_poll_callback` appends to `rdllist` in O(1);
+  level vs edge differs only in whether the epitem is re-queued after delivery
+  (`fs/eventpoll.c:1835-1851`). vibix does **not** ship epoll in this RFC —
+  only poll. Epoll is a future extension.
+- `drivers/tty/n_tty.c` — three 4 KiB ring buffers (`read_buf`, `echo_buf`,
+  driver `write_buf`), `N_TTY_BUF_SIZE = 4096` (`drivers/tty/n_tty.c:59`).
+  ICANON gates the line-commit path; ISIG dispatches `VINTR/VQUIT/VSUSP` to
+  `kill_pgrp(tty->ctrl.pgrp, sig, 1)` (`drivers/tty/n_tty.c:1044-1050`).
+- `drivers/tty/tty_jobctrl.c` — `__tty_check_change` returns `-ERESTARTSYS`
+  after delivering SIGTTOU to background writers (lines 33-67), and
+  `TIOCSPGRP` demands session-matching (lines 493-531).
+
+### POSIX reference points
+
+- XBD §11 — General Terminal Interface (termios: c_iflag / c_oflag / c_cflag /
+  c_lflag / c_cc[NCCS]).
+- XBD §11.1.7 — non-canonical MIN/TIME four-case semantics.
+- XBD §11.1.4 — SIGTTIN/SIGTTOU/SIGTSTP generation rules.
+- XSH `pipe()` — O_NONBLOCK/FD_CLOEXEC **shall be clear** on both new fds.
+- XSH `write()` — atomicity guaranteed up to `{PIPE_BUF}`, which must be ≥
+  `_POSIX_PIPE_BUF = 512` (XBD §13).
+- XSH `poll()` — `struct pollfd{int fd; short events; short revents;}`;
+  negative `fd` is ignored (revents=0, not POLLNVAL); timeout in ms with
+  −1 = infinite, 0 = poll, >0 = minimum ms. POLLERR/POLLHUP/POLLNVAL
+  are always returned in `revents` regardless of `events` mask.
+- `pipe2(2)` is **not** in POSIX.1-2017 base (it arrives in Issue 8 draft);
+  we ship it anyway because it is mandatory on modern Linux and allows
+  O_CLOEXEC-atomic creation — a small deviation from strict POSIX for a
+  critical race-freedom property.
+
+### SerenityOS / Redox patterns
+
+- **Serenity FIFO** wraps a `DoubleBuffer` (two 64 KiB flip-flop `KBuffer`s)
+  and calls `evaluate_block_conditions()` on every state change; EOF-on-no-
+  writer and EPIPE-on-no-reader are expressed as "always-ready" so the
+  subsequent read/write delivers the right error/short-count. PTYs live
+  in-kernel (`MasterPTY`/`SlavePTY`/`PTYMultiplexer`).
+- **Redox pipes** use `Mutex<VecDeque<u8>>` with `MAX_QUEUE_SIZE = 65536` and
+  encode read-vs-write endpoint in the low bit of the handle ID
+  (`WRITE_NOT_READ_BIT = 1`). Poll is expressed via the `event:` scheme +
+  per-scheme `fevent`. PTYs are a userspace daemon in Redox drivers, not
+  in-kernel.
+- **Critical architectural rule from Serenity** (`VirtualConsole::on_key_pressed`
+  → `Processor::deferred_call_queue`): the line discipline must **never** run
+  in ISR context. Keystrokes from the PS/2 or UART ISR are posted to a
+  deferred-call queue, and ldisc work runs in the kernel's soft-IRQ/worker
+  thread. We adopt the same rule.
+
+### Academic reference points
+
+- Banga, Mogul, Druschel — "A scalable and explicit event delivery mechanism
+  for UNIX" (USENIX ATC 1999). Canonical paper on why `select()` scales poorly;
+  motivates the eventual epoll/kqueue, but also motivates the batched-wake
+  design we propose in the novel idea below.
+- Lamport — "Specifying concurrent program modules" (TOPLAS 1983). SPSC
+  lock-free ring correctness proof; justifies our choice of a per-pipe SPSC
+  ring over `Mutex<VecDeque>`.
+- Ritchie — "A Stream Input-Output System" (Bell Labs TJ 1984). STREAMS
+  origin of pluggable line discipline. We adopt the trait-seam in spirit
+  (see `LineDiscipline` trait below) without the full STREAMS stack.
+- CVE-2022-0847 (Dirty Pipe) — cautionary tale for page-donation semantics
+  (splice `SPLICE_F_GIFT`). We deliberately **do not** support page donation
+  in this RFC; pipes are byte-copy.
+
+### x86-64 hardware considerations
+
+- TSO guarantees store-store and load-load ordering. An SPSC ring's producer
+  (`data[tail] = v; tail = tail + 1;`) and consumer
+  (`h = head; v = data[h];`) need *no* fence between the two stores
+  (producer) or the two loads (consumer) on x86-64 — only a compiler barrier.
+  (Intel SDM Vol. 3A, Memory Ordering chapter.)
+- A StoreLoad fence is needed only for wakeup-elision ("publish data, then
+  check if consumer is sleeping"). Linux uses `lock addl $0,(%rsp)` rather
+  than MFENCE; we follow suit.
+- Cache line is 64 B on all x86-64 parts (CPUID leaf 1 EBX[15:8] × 8).
+  Head and tail indices live on separate `#[repr(align(64))]` cache lines.
+- `rep movsb` with FSRM (CPUID.(EAX=7,ECX=0):EDX[4], Ice Lake+) is
+  competitive with SIMD for short copies; pipe small-message traffic
+  benefits. We expose a single `copy_to_user` / `copy_from_user` that
+  dispatches to `rep movsb` under STAC/CLAC (SMAP bracketing).
+
+## Design
+
+### Overview
+
+Four new kernel modules land together:
+
+```
+kernel/src/ipc/pipe.rs     — anonymous pipe + FIFO backend
+kernel/src/poll/mod.rs     — poll_table + sys_poll + sys_select
+kernel/src/tty/mod.rs      — Tty type, session/pgrp binding
+kernel/src/tty/n_tty.rs    — the N_TTY line discipline
+```
+
+Plus changes to:
+
+- `kernel/src/fs/mod.rs` — add `fn poll(&self, pt: &mut PollTable) -> PollMask`
+  default method on `FileBackend` returning `DEFAULT_POLLMASK` (ready for
+  read+write). Pipe, tty, and future socket backends override it.
+- `kernel/src/process/mod.rs` — session ID, process-group ID, controlling-tty
+  fields on the PCB. (Signal delivery is already there.)
+- `kernel/src/arch/x86_64/syscalls/` — six new syscall numbers
+  (see §Kernel–Userspace Interface).
+
+### Key Data Structures
+
+#### Pipe (anonymous + FIFO)
+
+```rust
+/// Capacity: 16 pages = 65_536 bytes, matching Linux default & Serenity
+/// DoubleBuffer. Power of two so head/tail wrap by mask.
+pub const PIPE_CAPACITY: usize = 64 * 1024;
+pub const PIPE_BUF: usize = 4096;   // POSIX atomic-write bound
+
+/// SPSC-ish byte ring. "-ish" because pipes permit multiple readers
+/// and multiple writers (POSIX); the ring itself is protected by a
+/// single `IrqSafeMutex`, and the head/tail pair is also exposed as
+/// two `AtomicUsize` on separate cache lines for a lock-free poll.
+#[repr(C)]
+pub struct PipeRing {
+    // Line 0: written by writer, read by writer + poll fast-path.
+    #[repr(align(64))] tail: AtomicUsize,
+    // Line 1: written by reader, read by reader + poll fast-path.
+    #[repr(align(64))] head: AtomicUsize,
+    // Bulk state.
+    buf: Box<[UnsafeCell<u8>; PIPE_CAPACITY]>,
+    lock: IrqSafeMutex<()>,
+}
+
+pub struct Pipe {
+    ring: PipeRing,
+    rd_wait: WaitQueue,
+    wr_wait: WaitQueue,
+    readers: AtomicUsize,
+    writers: AtomicUsize,
+    flags: AtomicU32,   // O_NONBLOCK, O_DIRECT (packet mode)
+    // Each record is (start, len); in packet mode this carries boundaries.
+    packets: IrqSafeMutex<Option<VecDeque<PacketHdr>>>,
+}
+
+pub struct PipeReadEnd(Arc<Pipe>);
+pub struct PipeWriteEnd(Arc<Pipe>);
+```
+
+Two fd backends (`PipeReadEnd`, `PipeWriteEnd`) share an `Arc<Pipe>`. On
+close, the relevant refcount decrements; when `readers == 0` a pending
+writer sees EPIPE + SIGPIPE. When `writers == 0` a reader drains the ring
+then sees EOF (`read` returns 0).
+
+The ring holds raw bytes, not `pipe_buffer`-style page references. This
+deliberately gives up splice/vmsplice/page-donation in the initial
+implementation (see Alternatives). A future RFC can layer `pipe_buffer`-
+style scatter/gather on top once we have a splice syscall use case.
+
+**Why bytes, not pages:** Linux's `pipe_buffer` was chosen because splice
+moves pages; without splice there is no value to the extra indirection, and
+a byte ring is simpler, has half the metadata, and avoids the CVE-2022-0847
+family entirely (no `PIPE_BUF_FLAG_CAN_MERGE` + stolen-page interactions).
+
+**Atomic-write guarantee for `n ≤ PIPE_BUF`.** A write of `n` bytes with
+`n ≤ PIPE_BUF` holds `lock` for the entire copy *only if* it cannot
+complete without blocking. The fast path (enough free bytes) takes `lock`,
+copies, updates `tail`, drops `lock` — a single critical section
+producing the POSIX non-interleaving guarantee. A blocking write of
+`n ≤ PIPE_BUF` waits for **≥ n free bytes** before acquiring the final
+copy-lock, so no partial write is ever visible. Writes of `n > PIPE_BUF`
+may interleave, matching POSIX XSH `write`.
+
+#### Poll table
+
+```rust
+pub type PollMask = u16;   // POLLIN|POLLOUT|POLLHUP|POLLERR|POLLNVAL|...
+
+/// Per-syscall scratch. Either `Scan` (non-blocking probe) or
+/// `Wait` (registration pass).
+pub struct PollTable {
+    mode: PollMode,
+    entries: ArrayVec<PollEntry, POLL_STACK_ENTRIES>,
+    spill: Option<Vec<PollEntry>>,
+    // Deadline-aware extension (§Deadline-aware poll readiness).
+    deadlines: Option<DeadlineSet>,
+}
+
+enum PollMode {
+    /// First pass: non-blocking probe, do not register waits.
+    Probe,
+    /// Second pass: register waits on each fd's WaitQueue.
+    Wait,
+}
+
+struct PollEntry {
+    waitq: *const WaitQueue,   // back-pointer into pipe/tty
+    token: WaitToken,          // handle for de-registration
+}
+```
+
+Driver `.poll` methods register themselves exactly as in Linux:
+
+```rust
+fn poll(&self, pt: &mut PollTable) -> PollMask {
+    pt.register(&self.pipe.rd_wait);   // no-op in Probe mode
+    pt.register(&self.pipe.wr_wait);
+    self.current_mask()                 // READ_ONCE of the atomics
+}
+```
+
+The **two-pass pattern** lives in the syscall body, not the driver:
+
+```rust
+fn sys_poll(fds: &mut [PollFd], timeout: Timeout) -> Result<usize> {
+    loop {
+        let mut pt = PollTable::probe();
+        let ready = scan(fds, &mut pt)?;        // Probe pass
+        if ready > 0 || timeout.is_zero() { return Ok(ready); }
+
+        let mut pt = PollTable::wait();
+        scan(fds, &mut pt)?;                    // Wait pass: register
+        // All waits are registered; re-check once more to close the race.
+        let ready2 = recheck(fds, &mut pt);
+        if ready2 > 0 { pt.cancel_all(); return Ok(ready2); }
+
+        match pt.sleep_until(timeout) {
+            Slept::Signal => return Err(EINTR),
+            Slept::Timeout => return Ok(0),
+            Slept::Woken => continue,   // loop and rescan
+        }
+    }
+}
+```
+
+#### TTY
+
+```rust
+pub struct Tty {
+    ldisc: Box<dyn LineDiscipline>,   // default: NTty
+    driver: Arc<dyn TtyDriver>,       // SerialTty, VirtualConsoleTty, Pty{Master,Slave}
+    termios: IrqSafeMutex<Termios>,
+    ctrl: IrqSafeMutex<JobControl>,
+    read_wait: WaitQueue,
+    write_wait: WaitQueue,
+}
+
+pub struct JobControl {
+    session: Option<SessionId>,       // set when a session leader acquires
+    pgrp: Option<ProcessGroupId>,     // foreground pgrp (TIOCSPGRP target)
+}
+
+/// Seam for pluggable line disciplines. N_TTY is the only one shipped here;
+/// future raw-mode, SLIP, or N_BPF (speculative) slot in behind this trait.
+pub trait LineDiscipline: Send + Sync {
+    /// Called from the driver side (ISR-deferred) with one raw byte.
+    fn receive_byte(&self, tty: &Tty, byte: u8);
+    /// Called from the userspace `read(2)` side; fills user buffer per
+    /// termios MIN/TIME and ICANON rules.
+    fn read(&self, tty: &Tty, buf: &mut [u8]) -> Result<usize, i64>;
+    /// Called from the userspace `write(2)` side; applies OPOST.
+    fn write(&self, tty: &Tty, buf: &[u8]) -> Result<usize, i64>;
+    /// Readiness predicate for poll.
+    fn poll(&self, tty: &Tty, pt: &mut PollTable) -> PollMask;
+}
+```
+
+```rust
+/// Termios bit layout matches Linux x86_64 ABI verbatim (see
+/// <asm-generic/termbits.h>): 4 × u32 flags + cc_t c_cc[NCCS] with
+/// NCCS = 19, c_line = 1 byte, ispeed/ospeed = u32. This is 36 bytes,
+/// and we preserve the Linux layout so a userspace built against Linux
+/// headers needs no shim.
+#[repr(C)]
+pub struct Termios {
+    pub c_iflag: u32,
+    pub c_oflag: u32,
+    pub c_cflag: u32,
+    pub c_lflag: u32,
+    pub c_line:  u8,
+    pub c_cc:    [u8; 19],   // NCCS
+    pub c_ispeed: u32,
+    pub c_ospeed: u32,
+}
+```
+
+#### N_TTY
+
+The canonical-mode state machine lives in `kernel/src/tty/n_tty.rs`. We
+deliberately diverge from Serenity's "parallel bitmask" trick and use an
+explicit `VecDeque<Line>`:
+
+```rust
+pub struct NTty {
+    state: IrqSafeMutex<NTtyState>,
+}
+
+struct NTtyState {
+    /// Incomplete line being accumulated in ICANON mode.
+    line_buf: ArrayVec<u8, 4096>,
+    /// Completed lines ready to read. A line ends at '\n', VEOF, or VEOL.
+    committed: VecDeque<Line>,
+    /// Raw-mode ring (ICANON off). Flat byte buffer.
+    raw: Ring<u8, 4096>,
+    /// Bytes queued to echo back to the driver.
+    echo: VecDeque<u8>,
+}
+
+struct Line { bytes: Box<[u8]>, contains_eof: bool }
+```
+
+**Justification for committed-line-queue over Linux's byte+bitmap:** in
+canonical mode, the boundary between "committed bytes readable now" and
+"line-in-progress not yet readable" is the dominant piece of state. Linux
+uses parallel bits per byte (`read_flags` bitmap) because its buffer is
+byte-flat; with a Rust `VecDeque<Line>` the boundary is structural and the
+readiness predicate is `!committed.is_empty()`. Memory overhead is one
+`Line` struct (24 B) per outstanding uncommitted line — bounded by the
+typical 1 or 2 lines in flight during interactive editing.
+
+### Algorithms and Protocols
+
+#### Pipe read / write fast paths
+
+Both operate under `pipe.lock`. Poll readiness checks do NOT take the
+lock: they read `head`/`tail` via `Ordering::Acquire` and use the ring's
+invariant `fill = tail - head (wrapping)` to compute readiness. This
+matches Linux's `READ_ONCE(pipe->head_tail)` trick (see Background).
+
+```
+Writer:
+  acquire pipe.lock
+  loop:
+    fill = tail - head
+    free = CAPACITY - fill
+    if free >= min_needed: break
+    if O_NONBLOCK: return EAGAIN
+    if readers == 0:
+       signal current with SIGPIPE (deferred past lock release)
+       return EPIPE
+    drop lock; wait on wr_wait; acquire lock
+  copy chunk into buf[tail..tail+n]
+  tail.store(tail + n, Release)
+  release lock
+  rd_wait.wake_one()          // directed wake, not wake_all
+```
+
+`wake_one` is a deliberate divergence from Serenity and Redox, both of
+which wake_all on every transition. Wake-all is a thundering-herd
+hazard for pipes with many blocked readers and no correctness benefit;
+the Linux pipe wait queues use `wake_up_interruptible_sync_poll(...,
+EPOLLIN)` which conceptually wakes one waiter plus any epoll watchers.
+
+#### TTY input path (ISR → ldisc → ring)
+
+```
+UART/PS-2 ISR:
+  byte = read_data_register()
+  deferred_call_queue.push(|| tty.driver.recv_byte(byte))
+  eoi()
+
+Soft-IRQ worker:
+  drain deferred_call_queue
+  for each byte:
+    tty.ldisc.receive_byte(tty, byte)
+
+NTty::receive_byte(tty, b):
+  t = tty.termios.lock()
+  // Input flags (c_iflag)
+  if ISTRIP: b &= 0x7f
+  if INLCR and b == '\n': b = '\r'
+  else if IGNCR and b == '\r': return
+  else if ICRNL and b == '\r': b = '\n'
+  // ISIG (before ICANON processing)
+  if ISIG in c_lflag:
+    if b == c_cc[VINTR]: send_signal(SIGINT); flush(); echo_control(); return
+    if b == c_cc[VQUIT]: send_signal(SIGQUIT); flush(); return
+    if b == c_cc[VSUSP]: send_signal(SIGTSTP); flush(); return
+  if ICANON in c_lflag:
+    handle_canonical_byte(b)   // VERASE, VKILL, VEOF, line commit on '\n'
+  else:
+    state.raw.push(b)
+    tty.read_wait.wake_one()
+  if ECHO: queue_echo(b)
+```
+
+`send_signal` resolves to `kill_pgrp(tty.ctrl.pgrp, sig)` with the same
+orphaned-pgrp check Linux performs (`drivers/tty/tty_jobctrl.c:33-67`).
+
+#### TTY write path (userspace → OPOST → driver)
+
+```
+NTty::write(tty, user_buf):
+  let t = tty.termios.lock();
+  // Background-pgrp check (SIGTTOU if TOSTOP).
+  if current.pgrp != tty.ctrl.pgrp
+     && t.c_lflag & TOSTOP
+     && !current.blocks(SIGTTOU)
+     && !current.orphaned_pgrp():
+      kill_pgrp(current.pgrp, SIGTTOU)
+      return ERESTARTSYS
+  drop t
+  for b in user_buf:
+    if OPOST:
+      if ONLCR and b == '\n': driver.put(b'\r'); driver.put(b'\n')
+      else if OCRNL and b == '\r': driver.put(b'\n')
+      else: driver.put(b)
+    else: driver.put(b)
+  return user_buf.len()
+```
+
+#### Job control
+
+- `TIOCSCTTY` (ioctl) — if caller is session leader and tty has no session,
+  bind tty.ctrl.session = caller.session.
+- `TIOCSPGRP` — tty.ctrl.session must equal caller.session AND target pgrp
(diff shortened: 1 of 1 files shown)


dburkart:

**Summary:** The RFC is well-researched and the syscall surface is largely faithful to the Linux x86_64 ABI, which is the right adoption path for vibix. However, several POSIX-visible corners — O_NONBLOCK/FD_CLOEXEC semantics on `pipe(2)`, FIFO open rendezvous, O_NOCTTY, the `ERESTARTSYS` leak into userspace, and the termios layout — are underspecified or technically wrong as written and need to be nailed down before this is implementable from a libc author's perspective.

**Blocking findings:**

- [B1] `pipe(2)` flag semantics not explicit — POSIX XSH requires that `pipe()` (no flags) leave both fds with O_NONBLOCK **and** FD_CLOEXEC **clear**. The RFC mentions `pipe2()` takes flags but never states this baseline for `pipe()`. Userspace will assume POSIX defaults; please state it in the Kernel–Userspace Interface section so a libc author doesn't have to guess whether `pipe()` is `pipe2(fds, 0)` or something else.
- [B2] `ERESTARTSYS` must not reach userspace — The errno table lists `ERESTARTSYS` as the return for background TTY writes (SIGTTOU path), and the pseudocode in the write path returns it literally. `ERESTARTSYS` is a Linux *kernel-internal* sentinel; userspace must see `EINTR` (or have the syscall transparently restarted when the installed signal handler has `SA_RESTART`). The RFC must document that the syscall trampoline translates this to EINTR / restart per the process's sigaction flags, and the errno table entry should be `EINTR` (restartable) from the userspace view. Also please specify whether vibix honors `SA_RESTART` on the EINTR paths for pipe/poll/tty — that is a program-visible contract.
- [B3] FIFO open rendezvous unspecified — FIFOs are listed in the roadmap and referenced throughout, but POSIX `open(2)` on a FIFO has very specific semantics: O_RDONLY blocks until a writer opens (or returns immediately with O_NONBLOCK); O_WRONLY blocks until a reader opens (or fails with ENXIO under O_NONBLOCK); O_RDWR is implementation-defined but conventionally non-blocking. None of this is described. Without it, libc's `fopen("fifo", "r")` behaviour is undefined on vibix. Please add a FIFO open-state subsection.
- [B4] `termios` layout does not match glibc's `struct termios` — The RFC claims "Linux x86_64 ABI verbatim" and then shows `c_iflag, c_oflag, c_cflag, c_lflag, c_line, c_cc[19], c_ispeed, c_ospeed`. That is the kernel's `struct termios2` layout (or `ktermios`), not the `struct termios` that glibc exposes to userspace via `<termios.h>`. The userspace glibc `struct termios` on Linux x86_64 is `{tcflag_t c_iflag; tcflag_t c_oflag; tcflag_t c_cflag; tcflag_t c_lflag; cc_t c_line; cc_t c_cc[NCCS]; speed_t c_ispeed; speed_t c_ospeed;}` only when `__USE_MISC` is set, and the `c_ispeed`/`c_ospeed` fields are actually populated by `TCGETS2` / `TCSETS2` ioctls, not `TCGETS`. If the RFC wants `TCGETS` to fill them, that is a real ABI deviation, not "verbatim." Either (a) drop `c_ispeed`/`c_ospeed` and expose them via `TCGETS2`/`TCSETS2` like Linux, or (b) document this as an intentional simplification and note that the struct is closer to `termios2`. `cfsetispeed`/`cfsetospeed` behaviour depends on which one you pick.
- [B5] O_NOCTTY and controlling-terminal acquisition on `open()` are missing — Linux's convention is: opening a TTY by path without O_NOCTTY *and* the opener is a session leader *and* the session has no controlling tty, then the tty is acquired as the session's ctty. The RFC only documents `TIOCSCTTY` as the acquisition path, which silently changes POSIX-shell behaviour (`sh` relies on the open-path acquisition on many systems). Either adopt the Linux convention or explicitly document that vibix requires explicit `TIOCSCTTY` and that O_NOCTTY is a documented no-op — userspace programs ported from Linux will otherwise quietly lose their ctty.

**Advisory findings:**

- [A1] `struct pollfd` alignment — The RFC defines `PollFd { fd: i32, events: i16, revents: i16 }` which is 8 bytes packed. On Linux x86_64 `nfds_t` is `unsigned long` and `struct pollfd` is 8 bytes with natural alignment 4, so this is fine. Please add a `static_assert!(size_of::<PollFd>() == 8)` and a note that the struct is `#[repr(C)]` + naturally 4-aligned so arrays marshalled from userspace match bit-for-bit. Worth a one-line comment because several BSDs pad this to 12.
- [A2] DAPRA userspace adoption path — DAPRA is genuinely novel and the additive design is the right call, but the RFC never answers "how does a program actually use this?" Options: (a) vibix ships a `<vibix/poll.h>` header and a thin wrapper in a `libvibix`; (b) a future libc patches `poll(3)` to transparently use `sys_poll_deadline` with `deferral_ns=0` (POSIX-identical) and exposes `poll_deadline(3)` as an extension; (c) it is strictly a kernel-native API with no libc shim. Please pick one in the Open Questions or a new "Userspace adoption" subsection. Without a story, DAPRA risks being a kernel-only feature no real program ever uses.
- [A3] `PollFdDeadline.deferral_ns: u32` cap — Max deferral ≈ 4.29 s is almost certainly enough (any program that wants to defer longer than that is really wanting a timer, not a poll batch). Please document the cap explicitly in the struct comment so userspace sees it as a contract rather than a discovered limitation.
- [A4] `isatty(3)` contract — Not mentioned, but libc will implement it as `ioctl(fd, TCGETS, &t)` → `ENOTTY` on non-TTY. Your errno table already lists `ENOTTY` for "fd is not a TTY", so `isatty()` works by construction. Worth a one-liner confirming this intended idiom.
- [A5] `SIGPIPE` default action — The RFC says "signal current with SIGPIPE" on a broken-pipe write but never states what happens when the process has not installed a handler. Default action is process termination with the signal; most shells `SIG_IGN` SIGPIPE explicitly or install handlers. Acknowledge this in the errno table or Security Considerations — it is standard but surprising to authors of small C programs.
- [A6] `POLLRDBAND` / `POLLWRBAND` accept-but-no-op — Good, POSIX-legal, documented in Open Questions. No action, just confirming.
- [A7] `eventfd` / `signalfd` / `timerfd` — Not mentioned. They're Linux extensions and out of scope is fine, but please add a one-line "explicitly out of scope; future RFC" so a reader does not assume they come with poll.
- [A8] Per-process `PollGroupToken` semantics on `execve` — Stated not to survive `fork`; say what happens on `execve`. Presumably tokens are dropped (the fd table is not inherited for group tokens because they aren't fds). Document to avoid a leak class.
- [A9] `F_GETFL` / `F_SETFL` on pipe fds — The RFC adds O_NONBLOCK and O_DIRECT on `pipe2()` but doesn't say whether `fcntl(F_SETFL, O_NONBLOCK)` post-creation is honored. Userspace expects it to be. Please confirm the flag is per-fd-description (Linux semantics) and mutable via fcntl.
- [A10] `select()` fd_set size — Classic `select` caps at `FD_SETSIZE` (1024 on Linux). If vibix keeps the 4096 cap from `nfds`, `select` needs to either enforce `FD_SETSIZE` or document the divergence.

dburkart:
### Review: OS Engineer

**Summary:** The RFC is architecturally sound and cites the right prior art, but its concurrency specification has several unresolved issues (incoherent pipe-ring model, missing global lock order, wake-one edge semantics, ISR back-pressure) plus at least one invalid-Rust snippet that will mislead implementers. These must be nailed down before code lands.

**Blocking findings:**

- [B1] PipeRing concurrency model is incoherent — The RFC says poll reads `head`/`tail` "lock-free via Acquire/Release" while writers update them under `IrqSafeMutex<()>`. That is only race-free if every mutator publishes `tail` (producer) / `head` (consumer) with `Release` *as the last action inside* the critical section and nothing else re-orders across lock release. The current pseudocode (`tail.store(tail + n, Release); release lock`) is fine, but the RFC never states this as an invariant, never forbids mid-critical-section partial updates of the indices, and never says what poll is allowed to observe between "bytes copied" and "tail published". Without a written rule of the form "indices are only ever advanced as the single publishing store of a completed chunk, under the lock, with Release ordering; no torn updates," the lock-free poll path is not provably safe. Spell the rule out and prove that `fill = tail - head` observed by poll is always a consistent snapshot (tail monotonically increases, head monotonically increases, both 64-bit on x86-64 so loads are atomic). Also note that a `Mutex<()>` does not by itself order the `Release` atomic store relative to the mutex release on weakly-ordered arches — the RFC leans on TSO without saying so.

- [B2] `#[repr(align(64))]` on struct fields is not valid Rust — The struct definition `#[repr(align(64))] tail: AtomicUsize` in `PipeRing` will not compile. `repr(align(N))` applies to type definitions (structs/unions/enums), not fields. The idiomatic fix is a wrapper newtype `#[repr(align(64))] struct CachePadded<T>(T)` or equivalent (e.g. `crossbeam_utils::CachePadded`) and store `CachePadded<AtomicUsize>`. This is also repeated in Performance Considerations ("enforced by `#[repr(align(64))]`"). Fix the RFC so implementers are not chasing an invalid syntax; otherwise the cache-line isolation claim is aspirational, not enforced.

- [B3] Missing lock ordering — Four sleepable/IRQ-safe locks interact: `pipe.lock`, `tty.termios`, `tty.ctrl`, and the per-`WaitQueue` internal lock. The RFC does not declare a total order. `receive_byte` takes `tty.termios` and then calls `tty.read_wait.wake_one()` (WaitQueue lock). `sys_poll`'s wait registration takes each fd's WaitQueue lock; if the driver's `.poll` also re-enters `termios` or `pipe.lock`, we have a lock-order inversion waiting to happen on SMP. A deadlock-free ordering must be stated explicitly — a reasonable candidate is `pipe.lock` / `tty.ctrl` / `tty.termios` → WaitQueue internal lock (wait queue is always innermost, never held across another lock acquisition), with a hard rule that driver `.poll` methods do not acquire any sleepable/IRQ-safe lock above the WaitQueue. The RFC must also state whether `IrqSafeMutex` disables IRQs on acquire (it should, given the name) and whether wake-from-ISR paths are therefore legal (they must not block on a locked waitqueue, so the waitqueue internal lock is presumably also IRQ-safe). Document this.

- [B4] Wake-one + edge-triggered correctness gap — The RFC prescribes `rd_wait.wake_one()` on every writer publish. For plain blocking `read(2)` + condvar discipline this is fine. But the RFC explicitly plans a future epoll with edge-triggered (EPOLLET) semantics, and DAPRA already relies on poll-group aggregation. With multiple blocked POLLIN waiters (e.g. epoll ET with several epitems, or multiple threads polling the same pipe), `wake_one` can wake a waiter that chooses not to drain — and because ET fires only on the 0→1 transition, remaining waiters miss the edge and deadlock until the next transition. Linux addresses this with `wake_up_interruptible_sync_poll(..., EPOLLIN)` which wakes one *blocking* waiter but still runs the poll callback for all epoll watchers. vibix's current text says "wake_one" unconditionally. Either (a) wake one classic waiter + run poll callbacks for all poll_table registrants, or (b) wake_all on pipes. Decide and write it down; "standard condvar discipline" is not sufficient once epoll/DAPRA are in the mix.

- [B5] `sys_poll` sleep-after-recheck race is not actually closed — The two-pass pattern text says "re-check once more to close the race," but the pseudocode is `recheck; if ready2==0 { sleep_until }`. Between `recheck` returning 0 and `sleep_until` actually parking the task, a wake can fire on the already-registered WaitQueue entries and be lost if the wait primitive is not wake-latching. The RFC must specify that `sleep_until` is a condition-wait primitive that *remembers* a wake occurring after registration (equivalent to `prepare_to_wait` + `schedule` with `TASK_INTERRUPTIBLE` on Linux — the wake flips the task state to RUNNING and the subsequent `schedule` returns immediately). Without that guarantee from `WaitQueue`/`sleep_until`, the two-pass pattern has a latent lost-wakeup. Add an invariant section for the wait primitive: "a wake after `register` and before or during `sleep_until` causes `sleep_until` to return `Woken` without sleeping."

- [B6] `readers == 0` SIGPIPE check is unsynchronised with respect to writes — The writer pseudocode does `if readers == 0 { ... return EPIPE }` after acquiring `pipe.lock`, but `readers` is declared `AtomicUsize` outside the lock, and the last reader's drop path decrements `readers` *outside* the pipe lock (the RFC does not say otherwise) and then calls `wr_wait.wake_all()` per Security Considerations. If the reader drop decrement is not serialised against the writer's observation + the subsequent copy, a writer can observe `readers > 0`, begin copying into a ring whose only reader just vanished, and no EPIPE is delivered until the next iteration — producing a silent "write to pipe that just lost its reader" that blocks forever on the next round if `wake_all` was issued before the writer parked. The fix (well-known from Linux) is: reader drop decrements `readers` *under* `pipe.lock` (or with Release + pipe.lock-acquired fence), then wakes; writer rechecks `readers` *under* `pipe.lock` after a wake. The RFC must pick one discipline (lock-protected count, or `Acquire`-load + recheck-after-wake) and commit to it; the current text is ambiguous and the bug is real.

- [B7] Deferred-call queue back-pressure is unspecified — The ISR enqueues one closure per keystroke; the ldisc worker drains them. If the worker is preempted or starved (long-running kernel work, scheduler latency), the deferred-call queue grows unboundedly — an attacker can hammer input and OOM the kernel with queued closures, or, if the queue is fixed-size, silently lose keystrokes with no signal. The RFC must specify: (a) is the deferred-call queue bounded? (b) what is the drop policy on overflow — newest, oldest, or hard panic? (c) is there any flow-control signal back to the driver (RTS on UART, PS/2 inhibit)? Serenity's pattern that the RFC cites handles this, and the RFC should either mirror it or justify deviating.

**Advisory findings:**

- [A1] WaitQueue-entry lifetime outliving PollTable — The RFC acknowledges this under Security Considerations (Pin<&PollEntry> stack-allocated, `PollTable::drop` cancels, debug-assert on wake). That is the right shape but the text should explicitly forbid the ISR from dereferencing a `PollEntry` across a kernel-preemption boundary without holding the WaitQueue's own lock — i.e. `wake_one` must hold the WaitQueue lock while dispatching, and `PollTable::drop` (cancel) must also take that lock, so the cancel-vs-fire race is ordered. Without stating this, the "debug-assert" is a symptom detector, not a correctness proof.

- [A2] `MFENCE avoidance` claim is right but under-cited — The text says Acquire/Release on indices is sufficient on x86-TSO. That is true for the SPSC data/index ordering, but the wakeup-elision pattern ("publish data, then check if consumer is sleeping") still requires a StoreLoad fence between the `tail` publish and the read of the waitqueue-empty flag. Linux does `smp_mb()` there, not just Release. The RFC's current pseudocode always wakes (`rd_wait.wake_one()` unconditionally), so there is no elision and no fence needed — but if anyone optimises this later to "skip wake when waitqueue empty," they will introduce a lost wakeup. Add a comment to the pseudocode that unconditional-wake is load-bearing, or explicitly call out the fence requirement for elision.

- [A3] `IrqSafeMutex<Option<VecDeque<PacketHdr>>>` on the hot path — In packet mode every write allocates a `PacketHdr` and pushes to a VecDeque under a second mutex, which means packet-mode writes take two locks in sequence (pipe.lock then packets lock). State the order (pipe.lock outer, packets inner) and consider folding the packet deque into the inner state protected by pipe.lock to avoid the second lock entirely.

- [A4] `PollFdDeadline` has `_pad: i16` but the layout is not size-annotated — On x86-64 with `#[repr(C)]` this lays out as i32+i16+i16+i16+i16(pad)+u32 = 16 bytes, which happens to be fine, but add a `static_assert_size!` (or equivalent) and a comment stating the exact expected size+alignment so a future field addition doesn't silently break the ABI.

- [A5] `O_NONBLOCK` is an `AtomicU32` on the `Pipe` itself — That conflates per-fd flag state with per-pipe state. Linux keeps `O_NONBLOCK` per-fd (in the `file` struct), not per-pipe — otherwise `dup2`+`fcntl(F_SETFL)` on one fd flips behaviour for the other end. Either move `O_NONBLOCK` to the `PipeReadEnd`/`PipeWriteEnd` (per-fd) or document the deliberate divergence from POSIX/Linux semantics.

- [A6] Default Termios `c_cc` values are listed but there is no statement that `c_ispeed`/`c_ospeed` have sensible defaults (e.g. B38400) — Some userspace checks these. Pin a default.

- [A7] DAPRA group tokens are process-scoped and "not inheritable across fork" — Good, but say what happens on `execve`: are existing group tokens revoked? If not, the token-namespace becomes unclear when libc-internal state is reset. Spell it out.

- [A8] The soft-IRQ worker is described as "the kernel's soft-IRQ/worker thread" — vibix does not yet have a distinct soft-IRQ context. If this lands before that infrastructure exists, the ldisc will run on some other deferred-work context (kernel thread? timer bottom half?) — name it, and ensure the chosen context permits the `tty.termios` lock acquisition pattern.

dburkart:
### Review: Performance Engineer

**Summary:** The RFC is thoughtful about cache layout and the lock-free poll fast path, but several performance claims are either unsupported, inconsistent with modern hardware baselines, or undermined by the pseudocode itself. The hot-path remains mutex-bound with no cited evidence for the throughput/latency targets.

**Blocking findings:**
- [B1] Unsupported pipe throughput target (≥ 2 GB/s single-core) — Linux's byte-ring pipe under ideal cache-resident conditions achieves roughly 5-10 GB/s single-core with FSRM `rep movsb`. A `Mutex<>`-protected byte copy plus SMAP STAC/CLAC bracketing plus wakequeue bookkeeping is strictly below that. 2 GB/s may actually be reachable, but the RFC offers no measurement, no cited baseline, and no breakdown (copy cycles vs lock overhead vs wake cost). Please either cite a reference benchmark (e.g., `pipe-test` from Linux self-tests, or a published lmbench number) or mark the target speculative pending measurement.
- [B2] Invalid Rust alignment in the one structure the perf claim depends on — `PipeRing` places `#[repr(align(64))]` on *fields* (`tail`, `head`) of a struct. That attribute is not valid on struct fields in Rust; it applies to type definitions only. The correct pattern is a wrapper type such as `#[repr(align(64))] struct CacheLine<T>(T);` or `#[repr(C, align(64))]` on a whole newtype struct. As written, the compiler will reject (or, worse, silently not pad) the layout, which directly invalidates the "only two cache lines bounce per round-trip" claim and the false-sharing mitigation. This is a performance-correctness issue, not just a style nit.
- [B3] `< 1 µs RTT` pipe-ping-pong target is aggressive without budgeting — two full context switches dominate. On modern x86-64, a kernel-entry + schedule + kernel-exit pair is ~1-2 µs *each* (see LWN "The cost of a context switch" measurements; Linux futex ping-pong is typically 1.5-3 µs RTT). The RFC acknowledges ctx switch is the limit but still targets <1 µs. Please either raise the target to a defensible 2-5 µs, or cite a measured vibix baseline showing a ctx switch fits inside ~400 ns. Also not budgeted: `wake_one` (~200-500 cycles + IPI if cross-CPU), SMAP STAC/CLAC (~5-20 cycles per boundary, non-trivial when the payload is only 8 bytes), and the termios/pipe mutex acquire/release.

**Advisory findings:**
- [A1] FSRM CPUID bit — RFC says "CPUID.(EAX=7,ECX=0):EDX[4], Ice Lake+". Please double-check against Intel SDM Vol. 2A (CPUID) and Intel Architecture Instruction Set Extensions Programming Reference. FSRM (Fast Short REP MOV) enumeration is easy to confuse with ERMS (bit 9 of leaf 7 EBX) and with the newer FSRS/FSRC variants. Some sources place broad FSRM availability at Sapphire Rapids and Alder Lake rather than Ice Lake client. Worth a citation directly from the SDM; otherwise the FSRM feature gate may be wrong on the hardware we care about.
- [A2] Mutex on the pipe hot path — every `read`/`write` takes `IrqSafeMutex`. The lock-free poll fast path is good, but the common case is read+write not poll. Under heavy multi-threaded userspace writers (e.g., a shared stderr from a threaded program) this serializes every byte. The RFC defers true SPSC to Open Question #1 "pending measurement" — acceptable as deferral, but please add a concrete measurement trigger (e.g., "if contended-write throughput < 50% of uncontended, swap in SPSC") so it doesn't get lost.
- [A3] DAPRA "≥ 30% fewer schedule events" baseline uncited — the number is plausible but speculative. Strengthen with a reference to io_uring multishot-poll's published gains (Axboe's LPC talks, or the Netflix io_uring study) so reviewers have an analogous datapoint rather than a bare target.
- [A4] TTY deferred-call-queue overflow behaviour undefined — RFC discusses the 4 KiB raw-ring overflow (silent drop, matches N_TTY) but says nothing about the deferred-call queue between ISR and soft-IRQ worker. Under a 1 MB paste into a non-draining terminal at e.g. 115200 baud UART burst, or a flood of PS/2 scancodes, what happens when `deferred_call_queue.push` finds the queue full? Performance cliff, silent drop, backpressure into ISR? This is a predictable pathological case worth specifying.
- [A5] STAC/CLAC cost on 8-byte pipe-ping-pong — at ~5-20 cycles per SMAP boundary and two crossings per round-trip, this is ~40 cycles baseline overhead on a sub-microsecond target. Negligible at 64 KiB bulk; non-trivial at 8 B. Consider mentioning in the performance section so the <1 µs target accounts for it.
- [A6] Cross-CPU wake cost — `wake_one` across cores pays an IPI (several hundred cycles, plus remote-CPU interrupt entry). vibix is UP today so this is latent, but the moment SMP lands the wake/sched budget in the RFC's latency math changes materially. Add a one-line note that targets are UP-measured and will be re-baselined on SMP.

dburkart:
### Review: Academic

**Summary:** The RFC is well-grounded in the canonical systems literature (Banga/Mogul/Druschel, Lamport SPSC, Ritchie STREAMS, Dirty Pipe) and its core pipes/poll/TTY design is sound. The "novel" DAPRA primitive, however, overstates its novelty: readiness-notification batching under a latency budget has close analogs in NIC interrupt moderation and soft-timers, and the prior-art survey in the Deadline-aware poll readiness section should be expanded before the novelty claim stands.

**Blocking findings:**
- [B1] Missing prior-art citations for DAPRA — The RFC claims "Academic search found no prior art" for attaching a soft deadline to readiness events and batching wakes. This is too strong. The closest analog at the hardware level is **adaptive NIC interrupt moderation / interrupt coalescing** (cf. Mogul & Ramakrishnan, "Eliminating Receive Livelock in an Interrupt-Driven Kernel," TOCS 1997; Intel's InterruptThrottleRate; DPDK's adaptive polling), which trades a bounded latency budget for wake/IRQ amortization — the same trade DAPRA makes, one layer up. Aron & Druschel's **soft timers** (SOSP 1999 / TOCS 2000) provide the exact mechanism DAPRA needs for cheap per-group deadline firing and should be cited as both prior art and an implementation primitive. Pariag et al. (EuroSys 2007) on connection-handling accept strategies and the EPOLLEXCLUSIVE literature are also relevant to the wake-one pipe design. Deadline-I/O scheduler work from the early 2000s (Linux deadline/CFQ I/O schedulers) is a looser analog but worth naming to delineate scope. None of these are identical to DAPRA — DAPRA is about **readiness-notification batching** with per-fd budgets composed across a poll group, not scheduler deadlines per se — but the current framing ("unclaimed territory") will not survive peer review. Add a prior-art paragraph that (a) names these works, (b) articulates precisely what DAPRA adds on top (per-fd deferral hint composed across a poll group, rather than a single device-wide moderation knob), and (c) downgrades the novelty claim from "no prior art" to "a new point in a known design space."

**Advisory findings:**
- [A1] Liveness under DAPRA — adversarial `deferral_ns` vs. `timeout` — The Invariants section states "Deferral never extends `timeout` — the waiter is woken at the earlier of (timeout, batch deadline)." This is the correct rule and blocks the obvious livelock (waiter with `deferral_ns > timeout` never woken), but the invariant should be stated more formally: for all waiters w in group g, wake_time(w) <= min(arrival_time(w) + timeout(w), first_ready(g) + min_deferral(g)). Also state explicitly that POLLERR/POLLHUP/POLLNVAL (already exempted), signal delivery (already exempted), and timeout expiry (this advisory) form the complete set of deferral-exempt wake sources. Without the enumeration, a future reader cannot verify closure of the exemption set.
- [A2] Two-pass poll as optimistic concurrency / barrier synchronization — A one-paragraph gesture at the formal invariant would strengthen the Algorithms section. The pattern is essentially: `register(W); publish_state(S); release` on the producer side and `subscribe(W); observe(S)` on the consumer side, with the property that any state transition either precedes the subscribe (and is seen by `observe`) or follows it (and is delivered via `W`). This is the classic "publication with double-check" / seqlock-adjacent invariant and is expressible in TLA+ as a two-state safety property over (`registered`, `state_observed`). Suggest citing Lamport's TLA+ work alongside the 1983 SPSC paper, and including a short prose invariant: "for every state transition t, either (a) t is observed in the Probe pass, or (b) t's wake arrives after the Wait-pass registration — never neither." A proof sketch is not needed for this RFC but the invariant should be named.
- [A3] LineDiscipline vs. STREAMS divergence — The RFC cites Ritchie 1984 for the `LineDiscipline` trait seam, but STREAMS' defining feature is **stacked** push/pop modules (one-at-a-time composition via `I_PUSH`/`I_POP`); the proposed `LineDiscipline` trait is single-object polymorphism, not a stack. This is fine — Linux's `tty_ldisc` is the same simplification — but the RFC should say so in one sentence ("We adopt STREAMS' pluggability but not its compositionality") to avoid a reviewer claiming the cite is inapt.
- [A4] Novelty is conditional, not strict — The Performance Considerations section honestly marks the 30% wake reduction as "speculative pending measurement." Good. But the Deadline-aware poll readiness section still asserts DAPRA is "better than the state of the art" before any measurement. Please soften the Design text to match the Performance honesty: "we conjecture DAPRA reduces wake count for workloads with >= 2 fds ready within the deferral window; this is subject to measurement in the `pipe-broadcast` bench." Strictly beneficial vs. conditionally beneficial is an important distinction in a peer-review setting.
- [A5] Wake-one pipe design citations — Pariag, Brecht, Harji et al. (EuroSys 2007, "Comparing the Performance of Web Server Architectures") and the EPOLLEXCLUSIVE kernel-list discussion are worth one additional citation in the Pipe read/write fast paths section alongside the existing Banga/Mogul/Druschel reference, since the wake-one decision is load-bearing and the thundering-herd literature is directly on point. The current Background coverage of these is adequate but the in-design justification would be stronger with an explicit cite.
- [A6] Dirty Pipe (CVE-2022-0847) citation — correctly attributed and correctly used as justification for rejecting page-donation semantics in v1. Flagging as positive; no change needed.
- [A7] Formal specification as future work — Given the two-pass poll pattern and DAPRA's group-aggregation both have clean safety/liveness statements, an Open Question entry on "lightweight TLA+ spec of the poll readiness protocol and DAPRA deferral bound" would be a cheap credibility win. Not blocking; propose as Open Question #8.

dburkart:

**Summary:** The RFC's security posture is generally sound — SMAP bracketing, nfds caps, session-leader gating of TIOCSCTTY, and pipe-ring zeroing are all spelled out. However, two load-bearing invariants (PollEntry lifetime and the DAPRA "no cross-process DoS" claim) have concrete defects I can demonstrate, and there are several unbounded kernel-side allocations an unprivileged process can trigger. The design is recoverable, but the blocking issues must be addressed in the RFC text before implementation.

**Blocking findings:**

- [B1] PollEntry holds a raw `*const WaitQueue` with no lifetime tie to the fd — use-after-free on close-during-poll. The struct is literally `waitq: *const WaitQueue, token: WaitToken`. The RFC's Security section only reasons about the reverse direction ("a wait queue that fires after the PollTable has dropped is a kernel bug"), but the real hazard is the other way: Thread A calls `sys_poll` on fd X and is in Wait mode with a registered `PollEntry` pointing into `Pipe::rd_wait`. Thread B `close(X)` drops the last reference to the `Arc<Pipe>` (nothing in the raw pointer keeps it alive), `Pipe` is freed, `rd_wait` is freed, Thread A's `PollTable::drop` then calls `cancel_all()` which dereferences a dangling pointer — or worse, the ISR-side `wake` deref fires first. Fix: hold `Arc<WaitQueue>` (or `Arc<Pipe>`) in `PollEntry`, or model this explicitly as a pinned intrusive list where `close` drains and cancels all registered entries under the WaitQueue's own lock before the Arc drops. The RFC should state the lifetime rule explicitly and the test matrix should include close-during-poll, not just drop-before-wake.

- [B2] DAPRA can be weaponized across process boundaries via back-pressure, contrary to the Security claim. The RFC asserts "groups are per-process; its deadlines only defer its own wakes" — this is true for the direct wake, but ignores that a deferred wake holds the reader process in kernel sleep, and *other processes blocked against that reader's progress are transitively deferred*. Concrete attack: attacker A reads a pipe from victim B with `deferral_ns = 1_000_000_000` (u32 ns max is ~4.29 s, more than enough). B's `write(2)` on a full 64 KiB pipe blocks until A drains. A's `poll` is ready immediately but the wake is deferred for up to 1 s. A's `read` therefore runs ~1 s later; B is starved for that second per full-buffer round. The same vector works against any TTY/ldisc output path where a slow reader blocks a writer. Mitigations to spec now: (a) hard cap `deferral_ns` at something interactive (e.g., 1 ms), (b) document that a process synchronously blocked waiting for the deferred reader's progress promotes the wake to immediate, or (c) require the deferral to be capped by the group's own timeout so starvation is bounded by the waiter's patience, not the attacker's. The RFC currently lists only POLLERR/POLLHUP/POLLNVAL/signals as deferral-promoting events; peer-blocked-on-me should also promote.

- [B3] PIPE_BUF atomicity is achievable, but the pseudocode admits a livelock-style DoS against a co-writer. The PIPE_BUF-≤-n atomicity claim itself is defensible under a single mutex because the writer re-checks `free >= n` inside the lock after each wake (the pseudocode loop is correct). However, the combination of `wake_one` on `wr_wait` plus an unfair `IrqSafeMutex` means two concurrent writers can starve one another: writer W1 waits for `n1` free bytes, writer W2 is woken ahead of W1, consumes the freed space with `n2`, and W1 re-sleeps; if W2 holds a tight write loop (either buggy or malicious, `dup(2)`-sharing the pipe), W1 never makes progress. This is not memory-unsafety, but it is an unprivileged DoS of one writer by another writer on the same pipe — relevant whenever a pipe fd is passed across a trust boundary (e.g., a privileged logger writing into a pipe also held by a sandboxed child). Either (a) FIFO-fair the `wr_wait` and/or the mutex, (b) require the newly-ready writer to be granted the lock next (ticket handoff), or (c) acknowledge explicitly in Security that co-writers on a pipe are assumed to be mutually trusting.

**Advisory findings:**

- [A1] Deferred-call queue from ISR appears unbounded. §TTY input path: "UART/PS-2 ISR: ... deferred_call_queue.push(...)". A keystroke storm (hostile serial peer, crafted VMM input, or IRQ storm) can push unbounded bytes into the deferred queue because the soft-IRQ worker runs at lower priority. The RFC's DoS section only discusses the 4 KiB raw ring overflow; the deferred-call queue sitting *in front of* the ring is the earlier bottleneck. Spec a bound (e.g., 64 entries per device) and document the drop policy on overflow (match the raw-ring "drop newest" policy for symmetry).

- [A2] `sys_poll_group_create` has no documented per-process cap. A process calling `poll_group_create` in a loop can allocate unbounded kernel `PollGroupToken` state plus associated deadline timers. Add an explicit cap (e.g., 64 groups/process) and return EMFILE past it; this is distinct from `RLIMIT_NOFILE` because groups are not fds.

- [A3] Termios `c_cc` is user-writable including VINTR/VQUIT/VSUSP, with no privilege boundary inside a session. A foreground process can set `VINTR = 0x00` so every NUL byte on the terminal line kills the foreground pgrp with SIGINT, and can reprogram `c_cc` while another pgrp is about to become foreground. This is POSIX-conformant (Linux does the same) and not a vulnerability per se, but the implicit trust model — "session leader trusts every member of every pgrp in its session with respect to termios" — deserves one sentence in Security for completeness.

- [A4] DAPRA `deferral_ns` as `u32` allows ~4.29 s deferrals, far exceeding the stated design targets (200 µs network, 1 ms work-stealing). A hard ceiling (e.g., 10 ms) eliminates the need for defensive reasoning about long deferrals and removes most of B2's severity. Prefer capping in the kernel on entry and returning EINVAL, rather than silently clamping.

- [A5] Error-path information disclosure: EPERM from TIOCSPGRP/TIOCSCTTY reveals whether the target pgrp is in the caller's session. This is POSIX-mandated behaviour and Linux-identical, so not a new surface, but worth noting as an information flow out of session/pgrp tables to any process with a tty fd. No mitigation requested; flagging for the threat model doc.

- [A6] `PollEntry.waitq` as `*const WaitQueue` also means the two-pass `recheck` in `sys_poll` must tolerate a fd being closed between `register` and `recheck`. The pseudocode does not specify what happens if `close` races with the recheck; combined with [B1], this should have its own test and a documented invariant (either closed fd yields POLLNVAL on recheck without dereferencing the stale pointer, or close blocks until all pollers have drained — Linux does the former via a per-file refcount dance).

dburkart:
## Defense cycle 1 — findings addressed

Updated RFC 0003 pushed as commit `b2943d2`. Every `[Bx]` blocking finding from the five reviewer archetypes is addressed below with a pointer to the specific change.

### Security Researcher

- **[B1] PollEntry UAF on close-during-poll** — `PollEntry` now stores `Arc<WaitQueue>` instead of `*const WaitQueue`. Added explicit lifetime-rule paragraph covering cancel-vs-fire ordering and close-vs-recheck invariant.
- **[B2] DAPRA cross-process DoS via peer-blocked-on-me** — Added (a) hard cap `DAPRA_MAX_DEFER_NS = 10_000_000` ns, (b) peer-blocked-on-me promotion rule (complementary waiter on same fd cancels the deferral timer), (c) reformulated liveness bound.
- **[B3] Co-writer starvation (wake_one + unfair mutex)** — RFC commits to FIFO-ordered `wr_wait` and ticket-handoff semantics on `IrqSafeMutex`. Recorded in §Pipe ring concurrency rules and §Security Considerations.

### OS Engineer

- **[B1] Invalid `#[repr(align(64))]` on struct fields** — Introduced `#[repr(C, align(64))] struct CachePadded<T>(pub T)` wrapper newtype; `head`/`tail` now wrapped instead of field-aligned.
- **[B2] Same `#[repr(align(64))]` issue in PipeRing** — Same fix applied; `const _: () = assert!(align_of::<CachePadded<AtomicUsize>>() == 64);` added.
- **[B3] Missing formal two-pass invariant** — Added formal either-Probe-observes-readiness-or-Wait-receives-wake invariant, proven against wait-latching contract.
- **[B4] `wake_one` breaks EPOLLET-style / DAPRA poll-group readers** — Replaced with `wake_poll_then_one` semantics (fires all poll callbacks + exactly one blocking waiter).
- **[B5] Latent lost-wakeup between recheck and sleep_until** — Added explicit wait-latching invariant to the `WaitQueue` primitive contract in §Algorithms.
- **[B6] readers==0 SIGPIPE race** — Reader drop path decrements `readers` under `pipe.ring.lock`; writer observation also under lock. Spelled out in §Pipe ring concurrency rules.
- **[B7] Unbounded deferred-call queue** — Bounded 128-entry ring, drop-newest policy, UART RTS deassertion when ≥ 75% full, soft-IRQ drains under bounded work.

### User Space Staff Engineer

- **[B1] pipe(2) must clear O_NONBLOCK / FD_CLOEXEC** — Added explicit "`pipe(2)` vs `pipe2(2)` flag semantics" subsection under §Kernel–Userspace Interface stating POSIX XSH behaviour verbatim.
- **[B2] ERESTARTSYS leaking to userspace** — Introduced `KERN_ERESTARTSYS` as a kernel-internal sentinel; syscall trampoline translates to either `%rip` rewind (restart) or `EINTR` per POSIX rules.
- **[B3] FIFO open rendezvous missing** — Added §FIFO open rendezvous subsection with full O_RDONLY / O_WRONLY / O_RDWR × O_NONBLOCK behaviour table.
- **[B4] termios struct layout mismatch with glibc** — Added a paragraph clarifying the layout matches Linux `termios2`/`ktermios`, not glibc `struct termios`; `const _: () = assert!(size_of::<Termios>() == 44);`.
- **[B5] Missing O_NOCTTY + open-path ctty acquisition** — Added §Job control rules: `open(tty, ..., !O_NOCTTY)` on the session leader of a ctty-less session acquires the ctty; O_NOCTTY suppresses.

### Academic

- **[B1] DAPRA novelty overstated** — Added §DAPRA prior art and positioning subsection citing Mogul & Ramakrishnan (TOCS 1997 livecock), Aron & Druschel (SOSP 1999 soft timers), EPOLLEXCLUSIVE, io_uring multishot, Pariag et al. (EuroSys 2007). Novelty framing downgraded to "a new point in this known design space, not unclaimed territory."

### Performance Engineer

- **[B1] 2 GB/s pipe throughput target undefended** — Marked **speculative pending measurement**. Added a Linux bare-metal calibration point (5–10 GB/s), explicit commitment to publish `pipe-bulk` benchmark numbers with the implementation PR, and a concrete below-500 MB/s investigation gate.
- **[B2] Same invalid `#[repr(align(64))]` issue** — See OS Engineer B1.
- **[B3] <1 µs RTT claim undefended** — Revised to **2–5 µs** with explicit cycle budget (context switches, wake, STAC/CLAC, mutex, waitqueue). `<2 µs` reclassified as a stretch goal for a later optimization pass.

### Out-of-scope

None — every blocking finding was addressed within the RFC.

Re-requesting review from **all five archetypes** (each had ≥ 1 blocking finding).

dburkart:
### Review: Security Researcher (cycle 2)

**Summary:** All three prior blocking findings are materially resolved: `Arc<WaitQueue>` in `PollEntry` closes the close-during-poll UAF, the 10 ms `DAPRA_MAX_DEFER_NS` cap plus peer-blocked-on-me promotion bound the cross-process DoS, and FIFO `wr_wait` + mutex ticket handoff eliminate the co-writer starvation vector. I find no new blocking issues introduced by cycle 1; a handful of documentation-consistency advisories remain.

**Blocking findings:**

(none)

**Advisory findings:**

- [A1] Stale Security-section paragraph on wait-entry lifetime — §Security Considerations still says "We implement wait entries as `Pin<&PollEntry>` stored in the `PollTable` (stack-allocated where possible), with explicit cancellation in `PollTable::drop`." This text predates the cycle-1 fix, which now holds `Arc<WaitQueue>` in `PollEntry`. The two descriptions are not contradictory but the §Security paragraph reads as if the raw-pointer model is still authoritative. Please rewrite it to reference the `Arc<WaitQueue>` lifetime rule defined in §Poll table so a reader doesn't infer that the Pin-on-stack model is the real invariant.

- [A2] "DAPRA wake starvation" security paragraph not updated — Still asserts "groups are per-process; its deadlines only defer its own wakes," which is exactly the claim cycle 1 had to walk back under B2. Please reword to cite the 10 ms cap and the peer-blocked-on-me promotion (and point at §DAPRA Invariants). Otherwise a future reader will re-derive the B2 attack from this paragraph alone.

- [A3] Peer-blocked-on-me is a cross-process timing oracle — The promotion rule lets process A infer "process B is now blocked on the complementary direction of fd F" because A's deferred wake fires immediately when B parks. This is observable by comparing A's wake latency against its `deferral_ns`. The information is already coarsely obtainable through normal pipe back-pressure patterns (EAGAIN, short writes), so this is not a new leak class, but it is worth one sentence in §Security acknowledging the side channel so it is an intentional design point rather than a surprise. Mitigation (e.g., jittering the promotion by a microsecond) is not warranted at this stage.

- [A4] `wake_poll_then_one` callback fan-out under the WaitQueue lock — Running every registered `PollEntry`'s poll callback inside the wait-queue lock bounds ISR/critical-section time by the number of watchers on a single pipe. A process with `RLIMIT_NOFILE`-many `dup`ed ends or many epoll groups targeting one fd can stretch a single `wake` critical section. Not a memory-safety issue and Linux has the same shape, but please either (a) state a bound on callback work (must be O(1), no allocation, no blocking locks) or (b) document the implicit limit via `RLIMIT_NOFILE`. This keeps the ISR-safety story airtight once DAPRA groups compose on shared fds.

- [A5] Ticket-handoff property is a contract on `IrqSafeMutex`, not on this RFC — The cycle-1 text says "both are already FIFO in the current vibix codebase, and this RFC commits to preserving that property." Please add an Open Question or roadmap item to add a compile-time or runtime assertion (or at minimum a unit test) that `IrqSafeMutex`'s ticket-handoff invariant is preserved across refactors, otherwise a future "optimize the mutex" PR can silently re-open the B3 starvation window.

- [A6] Peer-blocked-on-me is defined for pipes; unstated for TTY write path — The cross-process DoS in the original B2 was described against any writer blocked on a slow reader, explicitly including TTYs. The cycle-1 "complementary direction" language reads naturally for pipes (rd_wait / wr_wait) but is underspecified for TTYs where the peer is the line-discipline soft-IRQ worker or the driver output queue, not a userspace waiter. Please confirm whether TTY write-side back-pressure also triggers promotion, or state that TTY writers are never deferred by DAPRA at all (equivalent and simpler).

**Verdict:** LGTM

dburkart:
### Review: Academic (cycle 2)

**Summary:** The prior-art subsection now cites the canonical load-shedding / batching literature (Mogul & Ramakrishnan TOCS 1997, Aron & Druschel SOSP 1999, EPOLLEXCLUSIVE, io_uring multishot, Pariag et al. EuroSys 2007), and the novelty claim is honestly downgraded to "a new point in a known design space." The added formal two-pass invariant, wait-latching primitive contract, and DAPRA liveness bound are well-formed and do not contradict any published result I am aware of. My prior [B1] is addressed; I find no new blocking issues.

**Blocking findings:**
- (none)

**Advisory findings:**
- [A1] Liveness-bound edge case when `first_ready(g) < arrival_time(w)`. The bound `wake_time(w) ≤ min(arrival_time(w) + timeout(w), first_ready(g) + min_deferral(g), signal_time(w))` can be vacuous (or negative relative to `arrival_time(w)`) if the group already had a ready event before *w* joined. Operationally this is fine — the Probe pass returns immediately — but the formal statement should either (a) restrict `first_ready(g)` to the first readiness observed *after* `arrival_time(w)`, or (b) add the clause "or the Probe pass returns readiness." Clarifying this would close a minor ambiguity in the proof obligation.
- [A2] `min_deferral(g)` dynamism. The spec schedules the batch timer at `now + min(deferral_ns over all pollfds in group)` at first-ready time, but if a subsequent `sys_poll_deadline` call joins the group with a *smaller* `deferral_ns` after the timer is armed, the stated liveness bound `first_ready(g) + min_deferral(g)` can be tightened after the fact. Either the bound should read `min(deferral_ns at timer-arm time)` (i.e., the group's deferral is monotonically non-increasing and re-arms) or the timer should be re-armed when a tighter member joins. Worth spelling out; does not block.
- [A3] Peer-blocked-on-me promotion is conceptually analogous to priority inheritance (Sha, Rajkumar & Lehoczky, "Priority Inheritance Protocols: An Approach to Real-Time Synchronization," IEEE TC 1990). A one-line forward reference would strengthen the framing — the design is in fact using a well-studied pattern to resolve a DAPRA-induced inversion, which is a point in its favor, not against it.
- [A4] The DAPRA benefit conjecture ("≥ 2 fds ready within the deferral window") has a classical analog in IRQ-coalescing analyses (Salah et al., "Coalesce and Conquer," Computer Networks 2007) that empirically characterize the crossover point. Citing one such measurement would make the "not strictly beneficial" caveat sit within the known coalescing-throughput/latency literature rather than standing alone.

**Verdict:** LGTM

dburkart:
### Review: OS Engineer (cycle 2)

**Summary:** The cycle-1 edits correctly dispatch all seven of my prior blocking findings (CachePadded wrapper, formal two-pass invariant, wake_poll_then_one, wait-latching contract, readers-under-lock EPIPE coherence, bounded 128-entry deferred queue with RTS flow control). However, the cycle-1 rewrite introduces two new concurrency/interrupt-safety issues and leaves two internal inconsistencies that must be resolved before implementation.

**Blocking findings:**

- [B1] Deferred-call queue stores closures, not bytes — the ISR pseudocode at line 633 reads `deferred_call_queue.try_push(|| tty.driver.recv_byte(byte))`. Pushing a closure means either (a) heap-allocating a `Box<dyn FnOnce>` inside the UART/PS-2 ISR, which uses the global allocator and is categorically not IRQ-safe (the allocator may take a sleepable lock, and allocation failure under pressure is silent data loss on top of the drop-newest policy), or (b) boxing into some per-device slab that is also not specified. The bounded ring should carry raw `u8` bytes (the only payload the ldisc actually needs) with a per-device function pointer / vtable for the driver, exactly as Serenity does. The soft-IRQ worker then loops `for each byte { tty.ldisc.receive_byte(tty, byte) }` (which the RFC already shows at line 641-642). Respecify the queue element type as `u8` (or a small POD record, e.g. `{ driver_id: u8, byte: u8 }` when multiplexing) and delete the closure form — otherwise the "bounded, allocation-free, ISR-safe" property the whole B7 fix depends on is not actually delivered.

- [B2] Lock-order "mutually exclusive" clause is violated by the ldisc receive path — §Lock order (lines 485-489) declares the level-2 slot holds *exactly one* of `pipe.ring.lock` / `tty.termios` / `tty.ctrl` / `ntty.state` and states "no code path holds two of them at once." But `NTty::receive_byte` (lines 644-661) acquires `tty.termios.lock()` as `t` at line 645, then calls `handle_canonical_byte(b)` and `state.raw.push(b)`, which necessarily take `ntty.state` (the `IrqSafeMutex<NTtyState>` declared in the NTty struct at line 452). That is two level-2 locks held simultaneously on the hot ISR-deferred path, directly contradicting the stated invariant. Either (a) split the level-2 bucket into a strict sub-order `tty.termios → ntty.state` (and prove nothing in the TCSETS / ioctl / read paths acquires them in reverse), or (b) drop the `termios` lock after snapshotting the relevant flag bits and before touching `ntty.state`. Without one of these, SMP TCSETS-vs-receive will deadlock under contention.

**Advisory findings:**

- [A1] Errno table still says `ERESTARTSYS` — the prose at lines 697-700 explicitly states "the row for TTY background-write reads `EINTR (restartable if SA_RESTART)`," but the errno table at line 947 still reads `ERESTARTSYS | tty write | background write, SIGTTOU delivered`. This is the exact user-visible ABI row the fix was meant to correct; update the table to `EINTR` (with a footnote that it is restartable under `SA_RESTART`) so the RFC's own normative reference is consistent.

- [A2] Stale `wake_one` in TTY raw-mode branch — the B4 fix replaces `wake_one` with `wake_poll_then_one` on the pipe path, but `NTty::receive_byte` at line 660 still reads `tty.read_wait.wake_one()`. A raw-mode TTY with multiple epoll/DAPRA watchers loses the same edge-triggered wake the pipe path just fixed. Either apply `wake_poll_then_one` uniformly, or document per-path reasoning why the TTY raw ring is exempt (I don't see one — it is the same wait-queue shape as the pipe).

- [A3] `readers` / `writers` memory ordering on the decrement path — the reader-drop path (lines 592-601) decrements `readers` under `pipe.ring.lock`, but `readers` remains `AtomicUsize` for lock-free `pipe_poll` reads. The RFC does not state the ordering on the decrement (`Release`?) nor on the lock-free poll load (`Acquire`). On x86-64 TSO this is fine in practice, but per the §Pipe ring concurrency rule #4 precedent ("a future port to weakly-ordered ISAs will add an explicit fence"), spell out that `readers`/`writers` mutations use `Release` and lock-free observers use `Acquire`, same as the index rules.

- [A4] DAPRA peer-blocked-on-me promotion lock ordering unspecified — the §Security section promotion rule (lines 851-860) says "the moment B registers on the WaitQueue, the kernel checks if any member of the peer waiter set has a pending-deferred readiness" and cancels A's deferral timer. This is a new cross-queue operation that reaches from B's registration (level-3 WaitQueue lock) into A's DAPRA deferral-timer state (which must be protected by some level-2 or level-3 lock). The RFC does not place this promotion-check in the global lock order. If it acquires any level-2 lock from within a level-3 wait-queue registration, that's a lock-order inversion. Please add an explicit line to §Lock order describing where DAPRA deferral state sits and prove the promotion path does not take any lock above level-3.

dburkart:
### Review: Performance Engineer (cycle 2)

**Summary:** All three prior blocking findings are resolved. The throughput number is now clearly labeled speculative with a measurable gate, the RTT target has an explicit cycle budget, and the alignment bug is fixed via a `CachePadded<T>` newtype. No new blocking issues; a few small advisory cleanups.

**Blocking findings:**
- (none)

**Advisory findings:**
- [A1] RTT budget arithmetic is slightly loose — Summing the mid-points in §Performance ("Pipe small-message latency") yields roughly 1.0–1.8 µs (low: 400+200+100+50+200 ≈ 950 ns; high: 800+600+100+100+200 ≈ 1800 ns), yet the text then says "Sum: ≈2-4 µs before any I-cache/TLB effects." The committed 2–5 µs target is fine, but either the per-component numbers should absorb I-cache/TLB/scheduler-slowpath overhead explicitly, or the sum line should read "≈1–2 µs best-case, 2–4 µs with I-cache/TLB and scheduler overhead." Small edit; improves defensibility.
- [A2] Stale `#[repr(align(64))]` references in prose — After the `CachePadded<T>` fix the alignment is enforced on the wrapper type, but lines 153 ("Head and tail indices live on separate `#[repr(align(64))]` cache lines") and 1042 ("enforced by `#[repr(align(64))]`") still describe it as field-level. Update those two prose mentions to "via `CachePadded<T>`" so a future reader doesn't think the broken form is still in play.
- [A3] Cache-line discipline story is partially moot under the current design — §Performance advertises "only two cache lines bounce per round-trip" in the SPSC case, but the read/write fast path takes `IrqSafeMutex<()>` (the `lock` field in `PipeRing`). Under that mutex the head/tail CachePadded split buys very little for pipe I/O (the mutex cacheline is the hotter bouncing line). The split still helps the lock-free `pipe_poll` readers, which is the actual justification; worth calling that out so the trade-off is honest. Not a correctness issue.
- [A4] DAPRA "≥ 30% fewer schedule events" target is also speculative — noted in the RFC, fine to land, but apply the same "publish-on-implementation" commitment you applied to `pipe-bulk` so it doesn't drift into an unmeasured claim.
- [A5] Consider declaring the benchmark harness alongside the targets — `pipe-bulk` and `pipe-ring` are named but the harness (where it lives, how iteration count / warmup / QEMU-KVM invariants are pinned) isn't specified. Non-blocking for the RFC, but the investigation gate ("below 500 MB/s we investigate") is only enforceable if the measurement protocol is reproducible. A one-line pointer to where the bench code will land is enough.

**Verdict:** LGTM

dburkart:
### Review: User Space Staff Engineer (cycle 2)

**Summary:** Cycle-1 addressed the substance of every prior blocking finding — the flag-semantics subsection, `KERN_ERESTARTSYS` trampoline translation, FIFO rendezvous table, `termios2`/`ktermios` clarification + size assert, and the `!O_NOCTTY` session-leader ctty acquisition are all present and POSIX-faithful. One self-contradiction slipped into the errno table that will mislead implementers if not resolved; flagging it as blocking since it directly undoes the B2 fix at the exact spot the defense-1 note cited.

**Blocking findings:**
- [B1] Errno table still exposes `ERESTARTSYS` to userspace (contradicts the cycle-1 B2 fix) — §Kernel–Userspace Interface "Errno table for new paths" (line 947) still reads `| ERESTARTSYS | tty write | background write, SIGTTOU delivered |`. The cycle-1 defense note explicitly promised "The errno table in §Kernel–Userspace Interface is updated to reflect this: the row for TTY background-write reads `EINTR (restartable if SA_RESTART)`." The prose in §N_TTY (lines 686–701) is correct and the sentinel plumbing is right, but the reference table an implementer will copy from is wrong. Please change that row to `EINTR | tty write | background write, SIGTTOU delivered (restartable if SA_RESTART)` and, for symmetry, either add or confirm rows for `EINTR` on blocking pipe/FIFO open (FIFO rendezvous currently only documents `ENXIO`; a signal during the open also needs an errno row so libc authors know what `open(fifo, O_WRONLY)` can legitimately fail with).

**Advisory findings:**
- [A1] `pipe2` `O_DIRECT` documentation is self-contradictory — the syscall table row (line 885) says `pipe2` "accepts `O_NONBLOCK`, `O_CLOEXEC`, `O_DIRECT`" while the subsection below (line 896) says v1 rejects `O_DIRECT` with `EINVAL`. Pick one. Recommend keeping the EINVAL behaviour and changing the table row to `accepts O_NONBLOCK, O_CLOEXEC (O_DIRECT reserved, currently EINVAL)` so userspace authors writing feature-detection code do not assume packet mode works.
- [A2] `TCGETS2`/`TCSETS2` alias behaviour needs one more sentence — the termios paragraph says "`TCGETS2`/`TCSETS2` are accepted as aliases for `TCGETS`/`TCSETS` to ease porting of Linux userspace." Since vibix's canonical struct already carries `c_ispeed`/`c_ospeed` (i.e., the 44-byte ktermios shape), the aliases are trivially equivalent, but on Linux the `2` variants have a different contract around `BOTHER`/custom baud. Please state explicitly whether vibix supports `BOTHER`/custom-baud values in `c_cflag`, or whether baud is always one of the enumerated `Bnnn` constants. A porting shim that sets `BOTHER` + raw `c_ispeed` will otherwise silently misbehave.
- [A3] FIFO rendezvous `O_RDWR` deviation should be tagged as a formal non-POSIX extension — line 733 marks it "Linux-compatible extension," which is accurate but POSIX.1-2017 XSH `open` explicitly says `O_RDWR` on a FIFO is undefined. Recommend adding a one-line callout in the POSIX deviation list (alongside the `pipe2` deviation at line 103) so it is discoverable from a single place rather than only inside the rendezvous subsection.
- [A4] Job-control rule covers `open` path but not the "session leader with no ctty inherits across `execve`" question — the new §Job control paragraph cleanly specifies acquisition on `open(tty, !O_NOCTTY)`. It does not say what happens if a session leader inherits a TTY fd from a parent, has no ctty, and never re-`open`s it. Linux does *not* acquire a ctty in that case (acquisition is strictly `open`-time or `TIOCSCTTY`-time). Worth one explicit sentence: "Inheriting a TTY fd across `fork`/`execve` does not acquire a ctty; only `open(!O_NOCTTY)` and `TIOCSCTTY` do." This is the exact pitfall shell writers hit when a daemon is spawned from a login shell.
- [A5] FIFO rendezvous and `KERN_ERESTARTSYS`: please confirm that if the signal fires *after* `readers` has been incremented but *before* rendezvous completes, the reader count is decremented on the restart/EINTR path. The current prose ("The fd is installed in the caller's fd table only after the rendezvous completes") implies yes, but the cleanup order is not spelled out and a leaked refcount here causes a stuck writer on the other side — a classic FIFO bug.

dburkart:
## Defense cycle 2 — findings addressed

Updated RFC 0003 pushed as commit `8ee0ebc`. This is the **final** defense cycle (scoped cap: 2). Every `[Bx]` blocking finding from cycle-2 reviews is addressed below.

### OS Engineer (cycle 2)

- **[B1] Deferred-call queue pseudocode pushes a closure from the ISR** — Fixed. The queue is now a per-TTY `DeferredByteRing` — a fixed-capacity (128 bytes) MPSC ring of plain `u8` with `CachePadded<AtomicU8>` storage plus two atomic indices, embedded inside the TTY driver struct. Allocated once at driver construction. The ISR does only port I/O and atomic index bumps; no closure, no heap. RTS deassertion is plain MMIO. See updated `TTY input path` pseudocode under §Algorithms and Protocols.
- **[B2] Level-2 lock order says TTY locks are mutually exclusive, but `NTty::receive_byte` takes two** — Fixed. Replaced "mutually exclusive" with a **total order `2b → 2c → 2d`** inside the TTY subsystem (`tty.ctrl → tty.termios → ntty.state`). Every level-2 acquisition site is now documented with its exact ordering. For the per-byte ISIG signal-delivery path, `pgrp` is read via an `AtomicPid` snapshot updated at ctty-acquire / `TIOCSPGRP`, avoiding any `tty.ctrl` acquisition under `tty.termios`. Pipe's level-2 lock remains disjoint from all TTY locks (no handler touches both).

### User Space Staff Engineer (cycle 2)

- **[B1] Errno table still lists `ERESTARTSYS`** — Fixed. Replaced the `ERESTARTSYS` row with an `EINTR` row explicitly noting that the tty-write background-SIGTTOU case returns `EINTR` to userspace (restartable if `SA_RESTART`), and that `KERN_ERESTARTSYS` is kernel-internal only and translated at syscall-exit. Added an `EINVAL` row for `pipe2` with `O_DIRECT` in v1. Added a reinforcing paragraph immediately after the table stating ERESTARTSYS is never returned to userspace.

### Advisory-level cleanups also applied

- **OSEng [A2]** Stale `wake_one` on TTY raw-mode receive path → `wake_poll_then_one`.
- **Security [A1]** `Pin<&PollEntry>` paragraph in §Security rewritten around `Arc<WaitQueue>` and cancel-vs-fire serialization under the queue's own lock, plus the close-during-poll POLLNVAL-without-deref invariant.
- **Perf [A2]** Two stale `#[repr(align(64))]` references in prose replaced with `CachePadded<T>` newtype language.

### Out-of-scope

None — every blocking finding was addressed within the RFC.

### Cycle-2 LGTM reviewers (already clean)

Security Researcher, Academic, and Performance Engineer issued LGTM on cycle 1 and are not re-spawned. Re-requesting review from **OS Engineer** and **User Space Staff Engineer** only.

This is defense cycle 2 of 2 (scoped cap). Per `docs/agent-playbooks/rfc.md`, if any blocking findings remain after cycle 2 the RFC is withdrawn. If all clear, the RFC proceeds to §Approval.

dburkart:
### Review: User Space Staff Engineer (cycle 3 — final)

**Summary:** My cycle-2 blocker [B1] is fully resolved. The errno table (lines 989–999) now correctly reports `EINTR` as the user-visible error for background TTY writes, with `KERN_ERESTARTSYS` clearly demarcated as a kernel-internal sentinel. The reinforcing paragraph at lines 1001–1005 closes the loop by stating unambiguously that `ERESTARTSYS` is never returned to userspace and that the trampoline translates it to either a `%rip` rewind or `-EINTR`. The added `EINVAL | pipe2` row also tightens the `pipe2` contract (unknown flag bits, `O_DIRECT` for v1), matching POSIX/Linux ergonomics.

**Blocking findings:**
- None.

**Advisory findings:**
- [A1] Cross-reference hygiene — Consider a one-line forward reference from the `sys_pipe2` flag-semantics section (lines 940–951) to the errno table's `EINVAL | pipe2` row, so readers arriving at the flag description immediately see where the rejection errno is formally specified. Purely editorial; non-blocking.

**Verdict:** LGTM

dburkart:
### Review: OS Engineer (cycle 3 — final)

**Summary:** Both cycle-2 blockers are resolved. The `DeferredByteRing` is now a fixed 128-byte MPSC ring of plain `u8` embedded in the driver struct (zero allocation, ISR does only port I/O and atomic index bumps), and the TTY level-2 locks carry a documented total order `2b → 2c → 2d` with `tty.ctrl.pgrp` demoted to an `AtomicPid` snapshot for the ldisc ISIG path. No new blockers; a few small advisory notes on pseudocode prose consistency.

**Blocking findings:**

(none)

**Advisory findings:**

- [A1] Pseudocode at line 712 (`send_signal` resolving to `kill_pgrp(tty.ctrl.pgrp, sig)`) and line 721 (`if current.pgrp != tty.ctrl.pgrp`) still reads the field name `tty.ctrl.pgrp` literally. Per the normative rule at lines 511–516, the ldisc ISIG path and the TTY-write background-pgrp check run under `tty.termios` and MUST use the `AtomicPid` snapshot rather than touching `tty.ctrl`. The intent is clear from the lock-order section, but replacing those two occurrences with an explicit snapshot read (e.g., `tty.ctrl_pgrp_snapshot.load(Acquire)`) in the pseudocode would eliminate any chance of an implementer copying the literal form and re-opening the nested-lock hazard.

- [A2] `DeferredByteRing` is labeled MPSC but the only producer in the specified configurations (one UART ISR per port, one PS-2 ISR) is single-producer. MPSC is a superset so the label is not wrong, but a proper MPSC bounded byte ring requires a reservation-commit protocol (e.g., `fetch_add` on a "reserved head" with a second "committed head" that the consumer reads) — the current description ("two atomic indices, atomic index bumps") is under-specified if a future multi-producer case appears (e.g., a multi-IRQ-line UART with queued RX halves). Either (a) call it SPSC and pin it to one producer per device, or (b) spell out the reserve/commit indices before anyone tries to land MPSC semantics.

- [A3] Soft-IRQ worker falls back to "runs on the task that triggered the interrupt via a late-IRQ-return hook" until the soft-IRQ subsystem lands (lines 655–657). That fallback can re-enter an `IrqSafeMutex` that the preempted task was holding at the exact moment the IRQ fired — e.g., a reader in `NTty::read` holding `tty.termios` is interrupted, the late-IRQ-return hook runs `ldisc.receive_byte` which re-acquires `tty.termios`, deadlocking on `IrqSafeMutex`. Worth an explicit gate: either document that the late-IRQ-return hook defers ldisc work to after all kernel locks taken by the preempted task are released, or mark this as blocked on the soft-IRQ worker roadmap item and do not enable input-capable TTYs in the UP development config until that lands. Not re-blocking on this because the soft-IRQ worker is explicitly called out as the target design.

- [A4] The DAPRA "peer-blocked-on-me promotion" lock-ordering question I raised as cycle-2 [A4] still isn't explicitly resolved in §Lock order. The promotion path reaches from a waiter's `WaitQueue` registration (level-3) into the peer's DAPRA deferral-timer state, which is currently unplaced in the global order. Advisory only since it's a well-understood pattern, but a one-line placement in §Lock order ("DAPRA deferral state sits at a new level-2e, always acquired before any level-3 WaitQueue lock") would close the remaining ambiguity before implementation.

**Verdict:** LGTM

PR #381, opened 2026-04-15
signal: dispatch SIGSEGV/SIGBUS handlers from #PF via IRETQ frame rewrite

### LINKED ISSUE #346: signal: dispatch user SIGSEGV/SIGBUS handlers from #PF instead of defaulting to Terminate
## Motivation

Discovered while merging #344. The new `deliver_fault_signal_iret(SIGSEGV)` helper in `kernel/src/signal/mod.rs` collapses `Disposition::Handler(_)` into the same Terminate path as Default/Ignore. This means a user process cannot install a SIGSEGV/SIGBUS handler (common in JIT fault-recovery, stack-guard recovery, or as part of libc crash reporters). The #PF -> helper call happens from exception context on the ring-0 SYSCALL/IRQ stack, so we can't just call the existing `deliver_signal()` which builds a signal frame on the user stack: doing that from an interrupt frame requires switching to the task's kernel stack, rewriting the IRETQ frame to return to the handler trampoline instead of the faulting instruction, and setting up a sigreturn path. This is the blocker for real POSIX-style fault recovery.

## Work

- [ ] Add an IRETQ-frame rewrite path in `kernel/src/signal/mod.rs` (or a new `kernel/src/signal/fault.rs`) that, given a task's saved `InterruptStackFrame` + fault siginfo, builds a `rt_sigframe` on the user stack and redirects RIP to the handler trampoline.
- [ ] Wire `deliver_fault_signal_iret` (current call site: `kernel/src/arch/x86_64/idt.rs` page-fault handler, ring-3 branch near line 283) to branch on disposition: `Handler(_)` -> new rewrite path; `Default`/`Ignore` keep the current Terminate flow.
- [ ] Decide sigreturn mechanism: vdso-style trampoline page mapped per-process (Linux approach) vs. kernel-provided trampoline in a well-known VA. Pick one; document the choice in this issue before implementation.
- [ ] Integration test: ring-3 task installs a SIGSEGV handler, deliberately derefs a bad pointer, handler catches it, calls a sigreturn syscall, task exits cleanly with status 0. Gate on userspace test harness (likely after #225 lands).
- [ ] Extend to SIGBUS and SIGFPE once the SIGSEGV path is green.

## Context

- File: `kernel/src/signal/mod.rs`, function `deliver_fault_signal_iret` (added in merged PR #344 / commit 2ab675e).
- File: `kernel/src/arch/x86_64/idt.rs`, page-fault handler ring-3 branch at roughly line 277-284.
- Related issues: #337 (closed by #344), #225 (userspace test harness -- prerequisite for integration test).
- Related code: existing `deliver_signal` in `kernel/src/signal/mod.rs` (handles non-fault signals from syscall-return path); `check_and_deliver_signals`; `with_signal_state_for_task`.

dburkart:
Closes #346

## Summary

- **`deliver_fault_signal_iret`** now accepts `&mut InterruptStackFrame` and `fault_addr: u64`. For `Disposition::Handler(va)`, it pushes a `SigFrame` onto the user stack and rewrites the hardware-pushed RIP/RSP so `IRETQ` redirects to the signal handler. For `Default`/`Ignore`, it calls `task::exit()` as before (unchanged behavior).
- **`push_fault_signal_frame`** added to `signal/frame.rs`: extends `push_signal_frame` by writing `fault_addr` into `siginfo.si_addr` (bytes 16–23) and `SEGV_MAPERR=1` into `si_code` (bytes 4–7), so POSIX-compliant handlers can inspect the faulting address via `siginfo_t`.
- **IDT page_fault**: `frame` binding made `mut`; `&mut frame` and `addr_u64` passed to the helper; `return` added after the call so `IRETQ` fires for the handler path.

## Test plan

- [ ] Kernel compiles cleanly for `x86_64-unknown-none` (verified: `cargo check -p vibix` clean, no new errors or warnings)
- [ ] Existing signal module host unit tests pass (pre-existing aarch64 test-host limitation in this env; CI on x86_64 will confirm)
- [ ] Smoke test boot path unchanged (Default/Ignore fault handling is behavior-identical to before)
- [ ] Integration test for SIGSEGV handler dispatch deferred to after ring-3 test harness is wired

### DIFF
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -74,7 +74,7 @@ extern "x86-interrupt" fn general_protection(frame: InterruptStackFrame, code: u
     hang();
 }
 
-extern "x86-interrupt" fn page_fault(frame: InterruptStackFrame, code: PageFaultErrorCode) {
+extern "x86-interrupt" fn page_fault(mut frame: InterruptStackFrame, code: PageFaultErrorCode) {
     // Clear RFLAGS.AC immediately so the handler runs with SMAP live,
     // even if we entered mid-`stac` bracket from `uaccess::copy_*`. The
     // IRET at exit restores the saved RFLAGS (and therefore AC) from
@@ -280,7 +280,16 @@ extern "x86-interrupt" fn page_fault(frame: InterruptStackFrame, code: PageFault
             addr_u64,
             code,
         );
-        unsafe { crate::signal::deliver_fault_signal_iret(crate::signal::SIGSEGV) };
+        // SAFETY: we are in the #PF exception handler for the current task;
+        // `frame` is the live hardware-pushed InterruptStackFrame.  Passing
+        // `&mut frame` allows deliver_fault_signal_iret to rewrite RIP/RSP
+        // if a handler is installed (IRETQ redirects to the handler).  For
+        // Default/Ignore, deliver_fault_signal_iret calls task::exit() and
+        // never returns — IRETQ does not fire.
+        unsafe {
+            crate::signal::deliver_fault_signal_iret(crate::signal::SIGSEGV, &mut frame, addr_u64)
+        };
+        return; // handler path: IRETQ fires the signal handler
     }
 
     serial_println!(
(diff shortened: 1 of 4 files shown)


dburkart:
The xtask/src/main.rs change (llvm-objcopy selection) was added to fix a build failure encountered while working on this PR on an aarch64 host. It's logically independent but was included here since it was required to verify the build. Deferring to a follow-up issue for clean separation.

PR #424, opened 2026-04-15
fs: implement sys_getdents64 syscall

### LINKED ISSUE #411: vfs: implement getdents / readdir for directory iteration
## Motivation
Follow-up of #87 (VFS POSIX semantics tracking). Userspace `ls`, `find`, and libc `readdir(3)` need a kernel syscall that streams directory entries out of an opened directory fd.

## Work
- [ ] Add a `readdir` entry to the VFS inode op-set that yields `{ino, type, name}` records with an opaque cookie for resumption.
- [ ] Implement `sys_getdents64(fd, buf, count)` returning packed `linux_dirent64` records; return 0 at end of directory; advance the open-file position.
- [ ] Reject non-directory fds with `ENOTDIR`.
- [ ] Host unit + QEMU integration test that enumerates a directory with more entries than fit in one buffer, verifying resumption.

## Context
- Parent: #87.
- VFS trait lives under kernel/src/fs/.

dburkart:
Closes #411

## Summary

- Adds `FileBackend::getdents64` default returning `ENOTDIR`; overrides on `VfsBackend` to call `FileOps::getdents` with the shared `OpenFile.offset` as the resumption cookie.
- Wires `SYS_getdents64` (217) dispatch in `kernel/src/arch/x86_64/syscall.rs`: resolves the fd, validates the user range, bounces through a 4 KiB kernel buffer, and copies out via `uaccess::copy_to_user`.
- The tarfs `FileOps::getdents` already produces packed `linux_dirent64` records — this change just plumbs it end-to-end so userspace can enumerate directories.

## Test plan

- [x] Host unit tests: `cargo xtask build`
- [x] Integration test `syscall_getdents` (new) — 5 cases, all pass:
  - lists `/etc` and finds `hostname` (DT_REG) + `init` (DT_DIR); cookies monotonic
  - resumes correctly across multiple calls with a 32-byte buffer
  - `ENOTDIR` on a regular file, `EINVAL` on len=0, `EFAULT` on a kernel-half pointer
- [x] `cargo xtask smoke` — all 30 markers present
- [x] `cargo fmt --all --check` clean

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -193,6 +193,10 @@ name = "syscall_lseek"
 harness = false
 
 [[test]]
+name = "syscall_getdents"
+harness = false
+
+[[test]]
 name = "vm_integration"
 harness = false
 
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -358,6 +358,43 @@ pub unsafe extern "C" fn syscall_dispatch(
             }
         }
 
+        // getdents64(fd, buf, len) — read directory entries into a
+        // user buffer as packed `linux_dirent64` records. Advances the
+        // shared open-file offset (the per-open cookie) by however many
+        // entries were written. Returns bytes written, 0 at end-of-dir,
+        // or a negated errno (EBADF, EFAULT, ENOTDIR, EINVAL).
+        GETDENTS64 => {
+            let fd = a0 as u32;
+            let buf_va = a1 as usize;
+            let len = a2 as usize;
+            let backend = {
+                let tbl = crate::task::current_fd_table();
+                let x = match tbl.lock().get(fd) {
+                    Ok(b) => b,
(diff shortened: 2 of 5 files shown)


dburkart:
Thanks for the review.

- **EOF drain loop** — applied in `af70351`.
- **`USER_ACCESSIBLE` on the staging VMA** — deferred as not-a-bug here. These integration tests run entirely at CPL=0 (`syscall_dispatch` is called directly from kernel code), and `stage()` writes to the page via a raw `ptr::write_volatile` from kernel mode. With CR4.SMAP=1, adding `USER_ACCESSIBLE` would cause those supervisor writes to fault unless wrapped in STAC/CLAC. The syscall dispatcher itself uses `check_user_range` + `copy_to_user`, which handle SMAP correctly regardless of the PTE's user bit. Verified empirically: adding `USER_ACCESSIBLE` hangs the test on the first `stage()` call. Matches the established pattern in `kernel/tests/syscall_lseek.rs`.

PR #425, opened 2026-04-15
vfs: implement O_CREAT, O_EXCL, O_TRUNC, O_APPEND open flags

### LINKED ISSUE #412: vfs: implement O_CREAT, O_TRUNC, O_APPEND open flags
## Motivation
Follow-up of #87 (VFS POSIX semantics tracking). Standard userspace open semantics require these three flags; today `open()` cannot create, truncate, or append.

## Work
- [ ] `O_CREAT`: create the target file if it does not exist (mode via the third open arg). Combine with `O_EXCL` to fail with `EEXIST` if it does.
- [ ] `O_TRUNC`: truncate a writable regular file to length 0 on open.
- [ ] `O_APPEND`: writes atomically seek to end-of-file before appending, respected across `dup` / `fork`.
- [ ] Wire the new paths through the backing filesystem (initrd/ramdisk or subsequent rw fs).
- [ ] Host + QEMU integration tests for each flag including O_CREAT|O_EXCL collision.

## Context
- Parent: #87.
- VFS trait lives under kernel/src/fs/.

dburkart:
Closes #412.

## Summary

- `sys_openat` now honors `O_CREAT`, `O_EXCL`, and `O_TRUNC` instead of rejecting them with `EACCES`.
- `O_CREAT` splits off the parent directory, walks it, and calls `InodeOps::create`, then re-resolves to pick up the freshly materialized dentry/inode. `O_CREAT|O_EXCL` on an existing path returns `EEXIST`.
- `O_TRUNC` applies `setattr(SIZE=0)` on a writable regular file; `EISDIR` on a directory; silently ignored for read-only opens (matches Linux).
- `O_APPEND` is handled in `VfsBackend::write`: under the `OpenFile` offset mutex, the write offset is snapped to `inode.meta.size` so every write lands at EOF regardless of where the tracked offset points — dup'd fds and forked children that share the description serialize through the same mutex.
- Fixed an adjacent `ramfs::setattr` bug where the data vec was resized for `SetAttrMask::SIZE` but `inode.meta.size` was never updated, so `stat` / `fstat` continued to report the pre-truncate length after a truncate. `setattr(SIZE)` on a directory now returns `EISDIR`.

## Test plan

- [x] `cargo xtask build`
- [x] `cargo xtask test-unit` — 239 host unit tests pass.
- [x] `cargo test --test syscall_open_flags` (QEMU) — all 6 new tests pass: `create_new_file`, `create_without_excl_ok`, `excl_collision_eexist`, `open_missing_without_creat_enoent`, `trunc_zeroes_writable`, `append_writes_at_eof`.
- [x] `cargo xtask smoke` — 30/30 markers.
- [x] Pre-existing `blocking_sync::rwlock_writer_exclusion` flake reproduces on `main`; unrelated to this change.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -193,6 +193,10 @@ name = "syscall_lseek"
 harness = false
 
 [[test]]
+name = "syscall_open_flags"
+harness = false
+
+[[test]]
 name = "vm_integration"
 harness = false
 
--- a/kernel/src/arch/x86_64/syscalls/vfs.rs
+++ b/kernel/src/arch/x86_64/syscalls/vfs.rs
@@ -25,6 +25,7 @@ use super::super::syscall::copy_path_from_user_pub;
 use super::super::uaccess;
 use crate::fs::vfs::dentry::{DFlags, Dentry};
 use crate::fs::vfs::ops::{meta_into_stat, Stat};
+use crate::fs::vfs::ops::{SetAttr, SetAttrMask};
 use crate::fs::vfs::path_walk::{path_walk, LookupFlags, MountResolver, NameIdata, PATH_MAX};
 use crate::fs::vfs::super_block::SbActiveGuard;
 use crate::fs::vfs::Credential;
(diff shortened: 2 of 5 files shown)


dburkart:
Thanks CodeRabbit — the redundant `\0` on path literals is cosmetic (stage() appends the terminator regardless). Leaving as-is; not worth a follow-up commit or issue.

Issue #249, opened 2026-04-14
sync: add BlockingRwLock::waiter_count for test-harness synchronisation


dburkart:
## Motivation

Discovered while merging #245. CodeRabbit flagged that `rwlock_writer_not_starved` in `kernel/tests/blocking_sync.rs` uses a fixed 20-`hlt` grace period before dropping the driver's write guard to let spawned workers start running — this is flaky under extreme CI load. `Semaphore` and `WaitQueue` already expose `waiter_count()` for deterministic test synchronisation; `BlockingRwLock` doesn't. Adding the accessor lets the test poll until the expected number of waiters have parked instead of relying on a timing grace.

## Work

- [ ] Add `pub fn waiter_count(&self) -> usize` to `BlockingRwLock` in `kernel/src/sync/rwlock.rs`, delegating to `self.waiters.waiter_count()`.
- [ ] Update `rwlock_writer_not_starved` in `kernel/tests/blocking_sync.rs` to poll `RW_STARVE.waiter_count() >= RW_STARVE_READERS + 1` before dropping the driver's guard, removing the fixed 20-`hlt` grace loop.
- [ ] Document `waiter_count` as test-only in the doc comment, matching the existing `Semaphore::waiter_count` and `WaitQueue::waiter_count` wording.

## Context

- `kernel/src/sync/rwlock.rs` (`BlockingRwLock`)
- `kernel/src/sync/semaphore.rs` (pattern to mirror, lines 88-95 for `waiter_count`)
- `kernel/src/sync/waitqueue.rs` (base `WaitQueue::waiter_count`, lines ~115)
- `kernel/tests/blocking_sync.rs:954-959` (site of the fixed grace period)
- PR #245 (RFC 0002 item 1/15 — `BlockingRwLock` and `Semaphore` primitives)

PR #436, opened 2026-04-15
test: fix blocking_sync hangs by reaping worker tasks (#385)

### LINKED ISSUE #385: test: blocking_sync::rwlock_writer_exclusion times out under `cargo xtask test`
## Motivation

Discovered while working on #368 / PR #384. Running `cargo xtask test` on a clean checkout of `main` (commit 1e43c43) reproducibly times out in the `blocking_sync` integration test at `rwlock_writer_exclusion` after the previous test (`rwlock_concurrent_readers`) passes. The timeout fires at the 30s test-runner cap. This blocks the auto-engineer pipeline: every PR that touches kernel code currently fails the full `cargo xtask test` step on an unrelated test, forcing reviewers to special-case or ignore the failure. Per `docs/agent-playbooks/prioritization.md` §"Triage rules", CI flakes / broken-on-main status are P0.

## Work

- [ ] Reproduce by running `cargo xtask test` against `main` — currently deterministically hangs at `rwlock_writer_exclusion`.
- [ ] Inspect `kernel/tests/blocking_sync.rs` `rwlock_writer_exclusion` body and `kernel/src/sync/rwlock.rs` writer-acquire path for a missed wakeup, lost wait-queue entry, or starvation bug.
- [ ] Decide fix vs revert: if the test is asserting a real invariant, fix the rwlock; if the test is racy, tighten the test. Either is acceptable but the current state (broken-on-main) is not.
- [ ] Consider bisecting recent rwlock / waitqueue changes if no obvious root cause is found.

## Context

Failing output:

```
running 14 tests
...
test rwlock_concurrent_readers  fn() ... [ok]
test rwlock_writer_exclusion  fn() ... Error: "test blocking_sync-<hash> timed out after 30s"
```

- Affected files: `kernel/tests/blocking_sync.rs`, `kernel/src/sync/rwlock.rs`, `kernel/src/sync/waitqueue.rs`
- Head SHA used for repro: `1e43c43`
- Observed on PR branch `m368-file-backend-poll` with identical symptom.
- Related: #249 (sync: add `BlockingRwLock::waiter_count` for test-harness synchronisation) touches the same subsystem and may be relevant to the root cause or the test stabilisation work.

dburkart:
Closes #385

## Summary

Two changes that together make `cargo xtask test` deterministic again:

1. **`sync: add BlockingRwLock::waiter_count`** — mirrors the existing test-only accessor on `Semaphore` / `WaitQueue` / channel endpoints. `rwlock_writer_not_starved` now polls `RW_STARVE.waiter_count()` via `wait_for_parked` instead of a fixed 20-hlt grace period, closing the pre-park race the prior grace loop could lose under CI load (this is the work #249 describes — also closed by the first commit).

2. **`test: reap blocking_sync workers via task::exit`** — every worker in `kernel/tests/blocking_sync.rs` previously ended with `loop { hlt(); }`, which kept the task alive forever as a Ready scheduler entry. Across the 14 tests in that binary, ~25 zombie tasks accumulated by the time `rwlock_writer_exclusion` ran, each eating a 10 ms round-robin slice on every rotation. The scheduler fan-out was enough to push the later rwlock/semaphore tests past the 30 s test-runner cap deterministically on emulated CI hosts. Calling `task::exit()` — the same path `task_exit.rs` and `addrspace_drop.rs` already use — hands workers to the reaper as soon as they're done.

The blocking_sync binary now finishes in ~2 s with 0 zombies.

## Test plan

- [x] `cargo xtask test` — full suite green, 3/3 consecutive runs on this branch; previously 3/3 consecutive runs on `main` timed out in blocking_sync.
- [x] `cargo xtask smoke` — all 30 markers present.

### DIFF
--- a/kernel/src/sync/rwlock.rs
+++ b/kernel/src/sync/rwlock.rs
@@ -81,6 +81,15 @@ impl<T> BlockingRwLock<T> {
 }
 
 impl<T: ?Sized> BlockingRwLock<T> {
+    /// Number of tasks currently parked on this rwlock (readers or
+    /// writers waiting). Intended for test-harness synchronisation — a
+    /// driver can spin on `waiter_count() >= N` to confirm N workers
+    /// have actually enqueued before firing a release. Not appropriate
+    /// for production logic; the count can change the instant it's read.
+    pub fn waiter_count(&self) -> usize {
+        self.waiters.waiter_count()
+    }
+
     /// Try to acquire a shared (read) lock without parking.
     pub fn try_read(&self) -> Option<RwLockReadGuard<'_, T>> {
         let mut st = self.state.lock();
--- a/kernel/tests/blocking_sync.rs
+++ b/kernel/tests/blocking_sync.rs
@@ -138,9 +138,7 @@ fn ch_producer() -> ! {
         tx.send(i).expect("producer: send returned Err");
     }
     CH_PROD_DONE.fetch_add(1, Ordering::SeqCst);
-    loop {
-        x86_64::instructions::hlt();
-    }
+    task::exit();
 }
 
 fn ch_consumer() -> ! {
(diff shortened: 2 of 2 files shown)


dburkart:
@coderabbitai review

Issue #406, opened 2026-04-15
tty: wire 16550 serial UART rx path through Tty + LineDiscipline


dburkart:
## Motivation
Follow-up of #374. With the Tty + LineDiscipline primitives (sub-issue above) and the soft-IRQ primitive (sub-issue above) in place, the remaining work is driver-by-driver wiring. This issue covers the 16550 serial UART: on rx IRQ, bytes go into a per-driver `DeferredByteRing`; on the `SoftIrq::SerialRx` handler, the ring is drained and each byte is handed to the tty's ldisc via `receive_byte`. RTS deassertion at >= 75% full lands here too.

## Work
- [ ] Create a serial `TtyDriver` implementation wrapping the existing `uart_16550` path.
- [ ] Move the existing `serial: rx irq enabled` path in `kernel/src/drivers/serial.rs` (or wherever the serial rx IRQ handler lives) to push bytes into the driver's `DeferredByteRing`, raise `SoftIrq::SerialRx`, and return.
- [ ] Register a `SoftIrq::SerialRx` handler that drains the ring and invokes `tty.ldisc.receive_byte(&tty, b)` for each byte.
- [ ] Add MMIO RTS (ModemControl/FlowControl) deassertion when `ring.watermark_high()` first goes true; re-assert when it drops below.
- [ ] For now, bind a `PassthroughLdisc` by default so existing serial behavior does not change.
- [ ] Smoke-test: existing `serial: rx irq enabled` marker still present; all existing SMOKE_MARKERS still green.

## Context
- Parent (to be closed): #374
- Depends on (both are the sub-issues filed alongside this one):
  - Tty + LineDiscipline + DeferredByteRing primitives
  - soft-IRQ bottom-half primitive
- Merged RFC: `docs/RFC/0003-pipes-poll-tty.md`

Issue #405, opened 2026-04-15
tty: wire PS/2 keyboard rx path through Tty + LineDiscipline


dburkart:
## Motivation

Follow-up of #374. Counterpart to the serial-UART wiring follow-up. The existing PS/2 keyboard ISR currently pushes decoded characters directly to the framebuffer console; route it through a Tty + LineDiscipline instead so the canonical tty stack (N_TTY, line editing, echo) can apply uniformly to both input devices.

## Work

- [ ] Create a PS/2 `TtyDriver` implementation wrapping the existing keyboard decode path.
- [ ] On PS/2 rx IRQ: push decoded bytes into a `DeferredByteRing`, raise `SoftIrq::PS2Rx`, return.
- [ ] Register a `SoftIrq::PS2Rx` handler that drains the ring and invokes `tty.ldisc.receive_byte(&tty, b)`.
- [ ] Bind a `PassthroughLdisc` by default so behavior is preserved until N_TTY (#375) lands.
- [ ] Verify existing framebuffer/console smoke behavior still works end-to-end.

## Context

- Parent (to be closed): #374
- Depends on (both filed alongside this one):
  - Tty + LineDiscipline + DeferredByteRing primitives
  - soft-IRQ bottom-half primitive
- Existing PS/2 decode path lives under `kernel/src/drivers/` (grep for `ps2` / `keyboard`).

PR #438, opened 2026-04-15
task: soft-IRQ bottom-half worker primitive (#404)

### LINKED ISSUE #404: kernel: soft-IRQ bottom-half worker primitive (deferred ISR work)
## Motivation

Follow-up of #374. RFC 0003 requires ISR-entry handlers to be allocation-free and fast, with byte-level line-discipline work deferred to a soft-IRQ (bottom-half) context so the ISR can return quickly. No softirq infrastructure currently exists in the kernel — `grep -r softirq kernel/src` returns zero matches. This issue adds the generic primitive so the serial UART (#374) and PS/2 keyboard wiring can register drain callbacks against it. Unblocks the driver rewrites that are the remaining slices of the #374 decomposition.

## Work

- [ ] Implement `kernel/src/sched/softirq.rs` (or equivalent) with:
  - A fixed small set of soft-IRQ vectors (e.g. `SoftIrq::SerialRx`, `SoftIrq::PS2Rx`, extensible via enum).
  - `pub fn raise(vec: SoftIrq)` — safe from ISR (IRQ-off). Sets a pending bit atomically.
  - `pub fn register(vec: SoftIrq, handler: fn())` at init time.
  - A drain loop invoked from the scheduler tick tail or a dedicated per-CPU kernel task: check pending mask, clear each pending bit, run its handler. No closures across the ISR/soft-IRQ boundary — `fn()` only.
  - Re-raising while the handler runs re-schedules the handler (latch-into-pending).
- [ ] Wire the drain loop into the existing scheduler tick path or a dedicated kthread (whichever fits the current task model better).
- [ ] Host unit tests for the pending-bit semantics (IRQ-disabled atomics) where feasible; at minimum a kernel-target integration test that raises a softirq from a timer ISR and verifies the handler runs at least once.
- [ ] Do not modify any existing driver's ISR in this PR.

## Context

- Parent (to be closed): #374
- Merged RFC: `docs/RFC/0003-pipes-poll-tty.md` (§Soft-IRQ / deferred work)
- Blocks: the serial-UART-through-Tty and PS/2-through-Tty wiring follow-ups

dburkart:
Closes #404

## Summary
- New `kernel/src/task/softirq.rs`: fixed set of vectors (`SerialRx`, `PS2Rx`, two reserved slots), ISR-safe `raise`, init-time `register`, lock-free `drain`. Pending bits live in a single `AtomicU32`; handler slots are `AtomicPtr<()>` so neither the ISR path nor the drain takes a lock.
- `drain()` is called at the head of `preempt_tick` — runs with IRQs masked on the interrupted task's stack, before the scheduler lock is taken (so a handler that wakes a task can't deadlock on `SCHED`).
- Coalescing via `fetch_or` + `swap`: multiple raises between ticks run the handler exactly once. Re-raises inside a handler re-set the pending bit and are picked up on the next tick.
- No existing ISR is touched in this PR, per the scope cap in #404. Serial-UART (#406) and PS/2 (#405) follow-ups will wire the actual ISR→softirq edges.

## Test plan
- [x] New `tests/softirq.rs` with 6 cases: noop-drain, unregistered-drop, coalescing, multi-vector, handler-runs-on-tick, re-raise-latch — all pass
- [x] `tests/timer_tick.rs` and `tests/preempt.rs` still pass (drain hook doesn't perturb either path)
- [x] `cargo xtask smoke` — 30/30 markers
- [x] Host unit tests (`cargo test --lib`): 250/250 pass
- [x] `cargo fmt --all` clean; `cargo clippy --package xtask -- -D warnings` clean
- [ ] CI on PR

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -73,6 +73,10 @@ name = "preempt"
 harness = false
 
 [[test]]
+name = "softirq"
+harness = false
+
+[[test]]
 name = "pml4_switch"
 harness = false
 
--- a/kernel/src/task/mod.rs
+++ b/kernel/src/task/mod.rs
@@ -34,6 +34,7 @@
 
 pub mod priority;
 mod scheduler;
+pub mod softirq;
 mod switch;
 mod task;
 
@@ -905,6 +906,12 @@ fn reaper_loop() -> ! {
 /// a waitqueue); both paths restore a correct IF on return to task
 /// code.
 pub fn preempt_tick() {
+    // Drain any pending soft-IRQs before touching the scheduler lock.
+    // Handlers run IRQ-off on the interrupted task's stack; keeping
+    // them outside the SCHED critical section means a handler that
+    // wakes a task doesn't deadlock on our own lock.
+    softirq::drain();
+
     // `try_lock` + bail: block_current (or another preempt tick
     // mid-switch) may already hold SCHED. We'll get another tick in
     // 10 ms.
--- a/kernel/src/task/softirq.rs
+++ b/kernel/src/task/softirq.rs
@@ -1,0 +1,115 @@
+//! Soft-IRQ (bottom-half) primitive per RFC 0003.
+//!
+//! Hardware ISRs stay tiny: touch device state, EOI, return. Any work
+//! that can run later — parsing a scancode, pulling bytes through a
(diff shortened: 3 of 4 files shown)


dburkart:
@coderabbitai review

PR #444, opened 2026-04-15
tty: route 16550 serial UART rx through Tty + LineDiscipline

dburkart:
Closes #406

## Summary
- New `kernel/src/tty/serial.rs`: `SerialDriver` (wraps `crate::serial::write_bytes`), global `SerialDriver + PassthroughLdisc` Tty, `SERIAL_RX_RING: DeferredByteRing`, `push_byte_from_isr`, and a softirq drain handler that feeds each byte into `tty.ldisc.receive_byte`.
- `crate::serial::drain_rx_hardware` now additionally pushes each FIFO byte into the new ring and raises `SoftIrq::SerialRx`, leaving the legacy `RX_RING`/`try_read_byte` path untouched — shell/kernel consumers keep working.
- RTS flow control: on 75% watermark the ISR-side clears `MCR.RTS`; the softirq drain re-asserts once the ring drops below. Idempotent via an `AtomicBool` latch. No-op under QEMU `-serial stdio` (ignores MCR), meaningful on real hardware.
- `lib::init` calls `tty::serial::init()` after `tty::ps2::init()`.

Mirrors the PS/2 wiring pattern from #442.

## Test plan
- [x] `cargo xtask build` — clean.
- [x] `cargo xtask test` — 271 passed (+2 from new host-side `tty::serial` tests).
- [x] `cargo xtask smoke` — all 30 markers present (including `serial: rx irq enabled`).

PR #449, opened 2026-04-15
tty: N_TTY ISIG signal generation + orphaned-pgrp check

### LINKED ISSUE #431: tty: N_TTY ISIG signal generation + orphaned-pgrp check
## Motivation
Follow-up of #375 (decomposed). Signal-on-keypress is how Ctrl-C / Ctrl-\ / Ctrl-Z produce SIGINT/SIGQUIT/SIGTSTP to the foreground process group. The RFC mandates this runs before ICANON processing in `receive_byte`. The foreground pgrp must be read lock-free via the `AtomicPid` snapshot on `Tty.ctrl` to avoid acquiring `tty.ctrl` while holding `tty.termios` (RFC §Lock order inversion).

## Work
- [ ] In `NTty::receive_byte`: if ISIG is set in `termios.lflag` and the byte equals VINTR/VQUIT/VSUSP (from cc array), read foreground pgrp via `tty.ctrl.pgrp_snapshot.load()` (no lock) and call `signal::send_to_pgrp(pgrp, SIGINT|SIGQUIT|SIGTSTP)`. Do NOT commit the character to the line buffer in this branch (mark as consumed).
- [ ] Orphaned-pgrp check: mirror `drivers/tty/tty_jobctrl.c` — if the foreground pgrp has no live members in the same session, do not deliver the signal; deliver SIGHUP+SIGCONT to the pgrp instead (per Linux behavior when a background pgrp tries to read).
- [ ] Host unit tests using `process::test_helpers` (landed in #422) to set up session/pgrp state and verify signal dispatch; use a signal test shim if `send_to_pgrp` needs one.

## Context
- Parent (decomposed): #375
- Merged RFC: `docs/RFC/0003-pipes-poll-tty.md` (§ISIG, §Lock order)
- Depends on: #376 (TIOCSCTTY / TIOCSPGRP to populate `ctrl.pgrp_snapshot` at runtime). Host tests can set the snapshot directly via `Tty::new()` + `jc.set_pgrp(Some(pgid))`.
- Session/pgrp PCB landed in #422 (commit 8ab78fa).

dburkart:
Closes #431

## Summary
- New `NTty::receive_signal_or_byte` entry point: when `ISIG` is set and the byte matches `VINTR` / `VQUIT` / `VSUSP`, fan the corresponding `SIGINT` / `SIGQUIT` / `SIGTSTP` out to the foreground pgrp and consume the byte (don't commit to the line buffer). Otherwise delegates to the existing `receive_byte`.
- Orphaned-pgrp check: if the fg pgrp has no live members, deliver `SIGHUP` + `SIGCONT` instead of the originally-requested signal, matching Linux's background-pgrp read behaviour.
- Pgrp is read lock-free via `tty.ctrl.pgrp_snapshot` per RFC 0003 §Lock order — the hot path never acquires `tty.ctrl.lock()`.
- New `signal::send_to_pgrp(pgid, sig)` fan-out helper, backed by new `process::for_each_in_pgrp` and `process::pgrp_is_orphaned` accessors that keep the process `by_pid` table private.
- Signal delivery is abstracted behind a `SignalDispatch` trait. Production wiring uses `KernelSignalDispatch` (compiled only on `target_os = "none"`); host unit tests supply a capturing dispatcher so they don't need to build the gated `signal`/`process` modules.
- `receive_byte` is unchanged — the PS/2 (#442) and serial (#444) rx paths keep calling it while a future N_TTY `LineDiscipline` impl migrates to the signal-aware entry.

## Test plan
- [x] `cargo xtask test` — full suite green
- [x] `cargo xtask smoke` — 28/28 required markers
- [x] 7 new host unit tests in `kernel/src/tty/ntty.rs`:
  - `isig_ctrl_c_raises_sigint_and_consumes_byte`
  - `isig_ctrl_backslash_raises_sigquit`
  - `isig_ctrl_z_raises_sigtstp`
  - `isig_off_passes_through_to_receive_byte`
  - `isig_on_non_signal_byte_passes_through`
  - `isig_orphaned_pgrp_raises_sighup_sigcont_not_sigint`
  - `isig_pgrp_zero_drops_silently`
- [ ] End-to-end shell Ctrl-C SIGINT integration test — deferred; requires wiring an N_TTY `LineDiscipline::receive_byte` impl (tracked as a separate follow-up; the `receive_signal_or_byte` hook is the primitive that impl will call).

### DIFF
--- a/kernel/src/process/mod.rs
+++ b/kernel/src/process/mod.rs
@@ -405,6 +405,41 @@ pub fn sys_setpgid(pid: u32, pgid: u32) -> i64 {
     setpgid_for(current_pid(), pid, pgid)
 }
 
+/// Invoke `f(pid)` for every live entry whose `pgrp_id == pgid`.
+///
+/// Used by the N_TTY ISIG path to fan a signal out to the foreground
+/// pgrp (#431). Takes `TABLE.lock()` for the walk; the closure runs
+/// with the lock held, so callers must not re-enter the process table
+/// from `f` — typical use is `|pid| signal::raise_signal_on_pid(pid, sig)`
+/// which touches only the per-task signal state.
+pub fn for_each_in_pgrp(pgid: u32, mut f: impl FnMut(u32)) {
+    if pgid == 0 {
+        return;
+    }
+    let t = TABLE.lock();
+    for entry in t.by_pid.values() {
+        if entry.pgrp_id == pgid && matches!(entry.state, ProcessState::Alive) {
+            f(entry.pid);
+        }
+    }
+}
+
+/// Orphaned in the conservative sense the issue calls out: no live
+/// member of `pgid` remains in the process table. Linux's real
+/// `will_become_orphaned_pgrp` test also walks parents, but until
+/// session reparenting lands we approximate with "no members" — which
+/// matches the Linux behavior for a pgrp whose leader has exited and
+/// whose members all followed.
+pub fn pgrp_is_orphaned(pgid: u32) -> bool {
+    if pgid == 0 {
+        return true;
+    }
+    let t = TABLE.lock();
+    !t.by_pid
+        .values()
+        .any(|e| e.pgrp_id == pgid && matches!(e.state, ProcessState::Alive))
+}
+
 /// Session/pgrp test helpers, exposed so the `session_syscalls` kernel
 /// integration test can drive the table without the scheduler. Not
 /// gated on `cfg(test)` because the integration test is a separate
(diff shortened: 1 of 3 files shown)


dburkart:
Thanks — deadlock is real, pushed 2dc292f:

- Replaced `for_each_in_pgrp` with `collect_pgrp_members(pgid, &mut Vec<u32>)`, which populates the vec under `TABLE.lock()` and releases before returning.
- `send_to_pgrp` now walks the collected pids with the lock released, so `raise_signal_on_pid`'s internal re-acquisition no longer reenters a held lock.
- Doc comment on the helper updated to reflect the new pattern.

dburkart:
Pushed ec906c1 — rejecting `sig == 0` at entry with EINVAL. Kernel-internal callers only pass 1..=NSIG, and there's no "existence check" use case inside the kernel, so silently walking the pgrp and waking tasks for sig 0 is worse than a strict errno.

PR #453, opened 2026-04-16
tty: N_TTY ICANON line buffering (#429)

### LINKED ISSUE #429: tty: N_TTY ICANON line buffering (VERASE/VKILL/VEOF, \n/VEOF commit)
## Motivation
Follow-up of #375 (decomposed). ICANON is the line-buffered read mode POSIX shells rely on. Without it, `read(0, ..)` returns per-byte rather than per-line. This slice takes the byte already passed through input mapping and either edits the line-in-progress or commits it to the read buffer, waking readers.

## Work
- [ ] In `kernel/src/drivers/tty/ntty.rs` extend `NTty` with `line: LineBuffer` (owned by `NTtyState` lock, per RFC 0003 §Lock order rank 2d).
- [ ] `NTty::canon_input(termios, c)`: if ICANON is clear, push `c` directly into `state.raw`; else VERASE (\x7f) → pop last, VKILL (\x15) → clear line, VEOF (\x04) → commit as-is and mark EOF, \n or VEOL → append \n and commit. Plain chars append to line-in-progress.
- [ ] On commit: push all committed bytes into `state.raw`, call `wake_poll_then_one` on the read waitqueue (matches RFC wake semantics).
- [ ] Host unit tests: ICANON off (raw pass-through), ICANON on (line commit on \n), VERASE pops one, VKILL clears, VEOF empty line commit, long line wrap.
- [ ] No ISIG, no OPOST, no echo — the chained pipeline will be wired after those slices land.

## Context
- Parent (decomposed): #375
- Merged RFC: `docs/RFC/0003-pipes-poll-tty.md` (§ICANON)
- Depends on: #428 (N_TTY skeleton + input mapping)

dburkart:
Closes #429

## Summary

- Add ICANON canonical-mode line editing to N_TTY: `LineBuffer` (4 KiB line-in-progress) + `RawRing` (4 KiB committed read ring) behind `Mutex<NTtyState>`.
- `canon_input()` routes post-iflag bytes through line editing (VERASE/VKILL/VEOF/VEOL/newline commit) or directly to the raw ring (raw mode).
- EOF is recorded as a positional boundary on the raw ring (`eof_pos`), not a sticky flag, so the reader can return data up to the EOF point and continue after.
- `ReaderWake` trait decouples from Tty waitqueue — production wiring is a follow-up.
- `_POSIX_VDISABLE` (`c_cc[i] == 0`) handled via `matches_cc` helper.
- 13 new host unit tests covering all editing operations, raw mode, overflow, disabled-cc, ring wrap, positional EOF, wake semantics, and multi-line concatenation.

## Test plan

- [x] `cargo xtask build --release` — clean (1 pre-existing `is_guard_hit` warning).
- [x] `cargo xtask test` — 301 host unit tests + 5 QEMU integration tests pass.
- [x] `cargo xtask smoke` — all 28 required markers present.

### DIFF
--- a/kernel/src/tty/ntty.rs
+++ b/kernel/src/tty/ntty.rs
@@ -123,7 +123,6 @@ impl LineBuffer {
     fn as_slice(&self) -> &[u8] {
         &self.buf[..self.len]
     }
-
 }
 
 struct RawRing {
--- a/kernel/src/tty/ntty.rs
+++ b/kernel/src/tty/ntty.rs
@@ -220,11 +220,11 @@ impl NTty {
         b: u8,
     ) -> Option<u8> {
         if termios.c_lflag & ISIG != 0 {
-            let sig = if b == termios.c_cc[VINTR] {
+            let sig = if matches_cc(termios, VINTR, b) {
                 Some(SIGINT)
-            } else if b == termios.c_cc[VQUIT] {
+            } else if matches_cc(termios, VQUIT, b) {
                 Some(SIGQUIT)
-            } else if b == termios.c_cc[VSUSP] {
+            } else if matches_cc(termios, VSUSP, b) {
                 Some(SIGTSTP)
             } else {
                 None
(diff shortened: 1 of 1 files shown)


dburkart:
Re: partial raw-ring write on overflow (CodeRabbit finding on line 173) — this is an out-of-scope structural change. The line buffer and raw ring are the same size (4 KiB), so a partial write only occurs when the reader has stalled and the ring is nearly full. Linux N_TTY has the same drop-newest semantics in this edge case. Deferred to follow-up work.

Issue #118, opened 2026-04-13
GDB stub over serial (RSP protocol)


dburkart:
## Motivation

The kernel has good backtrace and symbol table support, but there is no way to set breakpoints, inspect live register state, or step through code after a fault. A minimal GDB Remote Serial Protocol (RSP) stub over UART would transform debugging — especially for the complex memory and scheduling work ahead.

## Scope

Implement a minimal RSP server activated by a triple-fault, deliberate `int3`, or a `gdbstub` shell command:

- **`?`** — report stop reason (signal number).
- **`g` / `G`** — read / write all general-purpose + segment registers.
- **`m` / `M`** — read / write memory (byte-granular).
- **`c`** — continue execution.
- **`s`** — single-step via EFLAGS.TF.
- **`z0` / `Z0`** — insert / remove software breakpoints (`int3` patching).

Transport: the existing UART write path; for RX, the serial RX infrastructure (see companion issue) or a simple polling read if that issue hasn't landed yet.

Expose the stub entry in the IDT `#BP` handler so `int3` in kernel code drops into the GDB loop. Document the QEMU invocation: `cargo xtask run` (with an added `-serial tcp::1234,server,nowait`) and `target remote :1234` in GDB.

## Out of scope

- Thread / task listing (`qfThreadInfo`, `T` packets) — deferred.
- Hardware breakpoints / watchpoints (DR registers) — follow-on.
- GDB Python scripts or `.gdbinit` automation.

## Done when

- `gdb vmlinux` (kernel ELF) + `target remote :1234` can hit an `int3`, inspect registers, and `continue`.
- A `gdbstub` shell command triggers the stop loop on demand.

dburkart:
Decomposed into sub-issues:

- #445 — gdbstub: RSP transport skeleton (packet framing + `?` stop reason) — foundation
- #446 — gdbstub: register read/write (`g`/`G`) + int3 (#BP) entry — depends on #445
- #447 — gdbstub: memory read/write (`m`/`M`) with safe-access wrapper — depends on #445 (parallel to #446)
- #448 — gdbstub: continue/step/breakpoints + `gdbstub` shell trigger — capstone, depends on #445 #446 #447

Closing in favor of the sub-issues. Auto-engineer will pick one of them on the next iteration.

PR #469, opened 2026-04-16
gdbstub: RSP transport skeleton + packet framer (#445)

### LINKED ISSUE #445: gdbstub: RSP transport skeleton (packet framing + `?` stop reason)
## Motivation
Follow-up of #118 (decomposed). First step of the GDB Remote Serial Protocol stub — the transport layer and packet framing, with just enough command surface to prove the loop works end-to-end (`target remote :1234` connects, issues `?`, gets a signal number back).

Establishes the foundation that sub-tasks for registers, memory, and control flow build on.

## Work
- [ ] New module `kernel/src/gdbstub/` (or similar). Packet framer: `$<data>#<cksum>`, two-hex-digit checksum, `+`/`-` ack handling.
- [ ] UART polling RX path suitable for use from debug context (the existing ISR-driven path won't work while we're stopped in the debugger — GDB is the only thing running).
- [ ] Parser dispatch for a single command: `?` → emit `S05` (SIGTRAP) as the canned stop reason.
- [ ] A `debug_entry()` entry point that drives the packet loop. Wire it to something trivial (a test-only public function) so an integration or QEMU test can exercise the loop. Actual fault-handler integration lands in the follow-up.
- [ ] Host unit test for packet framing (ack, checksum valid/invalid, escape bytes).

## Context
- Follow-up of #118
- Depends on: none (blocks #119-sub-2, #119-sub-3, #119-sub-4 — correct sub-issue numbers after filing)
- Labels: priority:P2, area:debug, area:driver

dburkart:
Closes #445

## Summary

First slice of the GDB Remote Serial Protocol stub — packet framing,
polling UART transport, and a minimal dispatch loop that handles `?`
(stop reason) and `D` (detach). Register/memory/control-flow commands
and the fault-handler wiring land in follow-up issues.

- `kernel/src/gdbstub/framer.rs`: pure byte-slice RSP codec
  (`\$<payload>#<xx>` framing, checksum, `}`-escape). Host-testable.
- `kernel/src/gdbstub/transport.rs`: \`Transport\` trait +
  \`VecTransport\` (in-memory) for tests.
- `kernel/src/gdbstub/uart.rs`: \`Com1PollingTransport\` that busy-polls
  LSR bits on raw 0x3F8 ports — safe from a halted fault-handler
  context where the interrupt-driven serial path would deadlock. Guarded
  behind `#[cfg(target_os = "none")]` so host builds don't need the UART.
- `kernel/src/gdbstub/mod.rs`: \`debug_entry::<T: Transport>\` — the
  packet loop. Unknown commands get an empty reply per RSP convention.

## Test plan

- [x] `cargo xtask build` — clean (pre-existing `is_guard_hit` dead-code
      warning unchanged).
- [x] Host unit tests written under `#[cfg(test)]` for framer (checksum,
      decode valid/bad/incomplete, escape roundtrip) and dispatch
      (`?`, bad checksum → NAK, unknown → empty reply, detach).
- [x] New QEMU integration test `kernel/tests/gdbstub_loop.rs` runs the
      same three scenarios inside a booted kernel to prove the code
      links and executes on `x86_64-unknown-none`.
- [ ] CI: host unit tests + QEMU integration (local test run skipped —
      host is aarch64 and \`pic8259\` requires \`target_arch = "x86_64"\`;
      CI on x86_64 is the source of truth).

## Not in scope

- Fault-handler (#BP/#UD) integration — tracked by the other #118
  sub-issues.
- Register read/write, memory read/write, continue/step.
- A \`gdbstub\` shell builtin to launch the stub manually.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -255,3 +255,7 @@ harness = false
 [[test]]
 name = "shell_help"
 harness = false
+
+[[test]]
+name = "gdbstub_loop"
+harness = false
--- a/kernel/src/gdbstub/framer.rs
+++ b/kernel/src/gdbstub/framer.rs
@@ -99,10 +99,7 @@ pub fn encode<'b>(payload: &[u8], out: &'b mut [u8]) -> &'b [u8] {
 /// `$` and `#`, matching the RSP spec and the encoder above. Only
 /// after the checksum succeeds is the payload unescaped into
 /// `scratch`.
-pub fn decode<'s>(
-    raw: &[u8],
-    scratch: &'s mut [u8],
-) -> Result<(Packet<'s>, usize), DecodeError> {
+pub fn decode<'s>(raw: &[u8], scratch: &'s mut [u8]) -> Result<(Packet<'s>, usize), DecodeError> {
     if raw.is_empty() {
         return Err(DecodeError::Incomplete);
     }
@@ -249,7 +246,10 @@ mod tests {
     #[test]
     fn decode_bad_checksum() {
         let mut scratch = [0u8; 16];
-        assert_eq!(decode(b"$OK#00", &mut scratch), Err(DecodeError::BadChecksum));
+        assert_eq!(
+            decode(b"$OK#00", &mut scratch),
+            Err(DecodeError::BadChecksum)
+        );
     }
 
     #[test]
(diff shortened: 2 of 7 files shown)


dburkart:
@coderabbitai review

PR #467, opened 2026-04-16
Add mkfifo/mknod FIFO support (#377)

dburkart:
Closes #377

## Summary
- Extends the VFS `InodeOps` trait with `mkfifo` and `fifo_pipe` methods (default bodies keep non-FIFO filesystems untouched) and grows ramfs with a `Fifo` body carrying a shared `Arc<Pipe>`.
- Teaches `Pipe` POSIX FIFO open rendezvous: new `open_read` / `open_write` / `open_rdwr` entry points drive blocking/NONBLOCK readers/writers against `open_wait_r` / `open_wait_w` queues with signal-safe rollback of the reader/writer refcount.
- Wires `SYS_MKNOD` (133) and `SYS_MKNODAT` (259) in the x86_64 dispatcher and teaches `openat()` a FIFO detour that installs a `PipeReadEnd` / `PipeWriteEnd` / dual-ended `FifoRdwrBackend` instead of a `VfsBackend`.
- Adds a `syscall_mkfifo` QEMU integration test covering the happy path, `EEXIST` on re-mknod, `ENXIO` on nonblocking O_WRONLY without a reader, nonblocking O_RDONLY without a writer, `O_RDWR` round-trip, `mknodat(AT_FDCWD)`, and `EPERM` for unsupported file types.

## Test plan
- [x] `cargo xtask build`
- [x] `cargo xtask smoke` (all 30 markers present)
- [x] `cargo test -p vibix --test syscall_mkfifo --target x86_64-unknown-none ...` — all 7 tests pass under QEMU
- [x] `cargo fmt --all -- --check` clean
- [ ] CI host unit tests (run on amd64 runner — local host is aarch64, pic8259 is x86-only)

PR #475, opened 2026-04-16
fs: make fcntl(F_SETFL) propagate to backend state

### LINKED ISSUE #435: fcntl(F_SETFL) does not propagate O_APPEND/O_NONBLOCK to live backends
## Motivation
Discovered while merging #427. `fcntl(F_SETFL)` currently only mutates `FileDescription.flags` (in `kernel/src/fs/mod.rs::set_status_flags`), but the live I/O backends consult other state: `VfsBackend::write` reads `OpenFile.flags` for `O_APPEND` handling, and `PipeReadEnd`/`PipeWriteEnd` keep their own `nonblocking` atomics. So `F_GETFL` reflects the new flag but writes still ignore `O_APPEND` and pipes still block — `F_SETFL` is effectively a no-op for semantics.

## Work
- [ ] Decide source of truth for mutable status flags (`FileDescription.flags` vs `OpenFile.flags` + pipe atomics).
- [ ] Option A: make `VfsBackend::write` consult `FileDescription.flags` for `O_APPEND`; remove the `OpenFile.flags` append path.
- [ ] Option B: make `set_status_flags` also update `OpenFile.flags` and pipe `nonblocking` atomics, crossing the backend boundary.
- [ ] Add regression tests: open a file with `O_WRONLY`, `F_SETFL O_APPEND`, write, verify offset went to EOF. Pipe: `O_NONBLOCK` via `F_SETFL` causes read on empty pipe to return `EAGAIN`.

## Context
- `kernel/src/arch/x86_64/syscall.rs:419-421` (F_SETFL dispatch).
- `kernel/src/fs/mod.rs` `set_status_flags` CAS loop.
- `kernel/src/fs/vfs/backend.rs` (`VfsBackend`).
- `kernel/src/ipc/pipe.rs` (`Pipe*End` nonblocking).
- CodeRabbit review on PR #427 flagged this as Critical; commit b72f9c4 deferred it.

dburkart:
Closes #435

## Summary

- `fcntl(F_SETFL)` flipped the mutable status bits on `FileDescription.flags` but every live I/O path consulted a separate cached copy, so the edits were silently ignored on the very next read/write. `VfsBackend::write` reads `OpenFile.flags` for the `O_APPEND` EOF snap, and the pipe ends each carry their own `nonblocking: AtomicBool`.
- Fix with a new `FileBackend::set_flags` hook that `set_status_flags` calls after the CAS on the description succeeds. Hand the post-update flag word to the backend; each impl mirrors the subset it cares about. Default is a no-op so `SerialBackend` stays untouched.
- `OpenFile.flags` is promoted from `u32` to `AtomicU32` so concurrent `F_SETFL` through dup'd fds is lock-free.

## Test plan
- [x] Host unit tests: `VfsBackend::set_flags` toggling `O_APPEND` drives the next write to EOF; access mode + `O_CLOEXEC` preserved; end-to-end via `FileDescTable::set_status_flags`; both pipe ends flip/clear `nonblocking` and ignore non-`O_NONBLOCK` bits.
- [x] Integration tests: `fcntl_setfl_o_append_writes_at_eof` (seed `/tmp` file → F_SETFL(O_APPEND) → next write lands at EOF) and `fcntl_setfl_o_nonblock_on_pipe_read_returns_eagain` (blocking pipe → F_SETFL(O_NONBLOCK) on rfd → empty read returns EAGAIN).
- [x] `cargo xtask build`
- [x] `cargo xtask test-integration` — all green including the two new fcntl tests
- [x] `cargo xtask smoke`

### DIFF
--- a/kernel/src/arch/x86_64/syscalls/vfs.rs
+++ b/kernel/src/arch/x86_64/syscalls/vfs.rs
@@ -393,6 +393,10 @@ impl FileBackend for FifoRdwrBackend {
     fn poll(&self, pt: &mut crate::poll::PollTable) -> crate::poll::PollMask {
         self.read_end.poll(pt) | self.write_end.poll(pt)
     }
+    fn set_flags(&self, new_flags: u32) {
+        self.read_end.set_flags(new_flags);
+        self.write_end.set_flags(new_flags);
+    }
 }
 
 /// `stat(path, *statbuf)` (follow=true) / `lstat(path, *statbuf)`
--- a/kernel/src/fs/mod.rs
+++ b/kernel/src/fs/mod.rs
@@ -91,6 +91,21 @@ pub trait FileBackend: Send + Sync {
     fn as_vfs(&self) -> Option<&vfs::VfsBackend> {
         None
     }
+
+    /// Propagate mutable open-file status flags into backend-local state.
+    ///
+    /// Called from [`FileDescTable::set_status_flags`] after the
+    /// `FileDescription.flags` CAS succeeds. `new_flags` is the full
+    /// post-update flag word (the description's canonical value) so that
+    /// backends holding a duplicate atomic (e.g. pipes' `nonblocking`) can
+    /// synchronise their own bits without re-deriving them.
+    ///
+    /// Only `O_APPEND`, `O_NONBLOCK`, and `O_ASYNC` will ever change here —
+    /// POSIX pins the other bits. The default is a no-op, which is correct
+    /// for backends that either read the description's flags directly on
+    /// every I/O (regular files snap `O_APPEND` from `OpenFile.flags`) or
+    /// that simply don't care about status flags (serial stdio).
+    fn set_flags(&self, _new_flags: u32) {}
 }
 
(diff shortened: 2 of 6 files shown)


dburkart:
Rebased onto main (#467 mkfifo/mknod landed while this was waiting for review) and addressed CodeRabbit's stale-mirror race finding.

**CAS-retry mirror loop** (kernel/src/fs/mod.rs): after mirroring to the backend, re-load `desc.flags` and re-mirror if a newer F_SETFL has landed. Per-iteration the mirror reflects the flag word observed after the most recent CAS, so the final backend state always matches the winning atomic regardless of scheduler order.

**FifoRdwrBackend::set_flags** (kernel/src/arch/x86_64/syscalls/vfs.rs): fans out to both pipe ends so O_NONBLOCK via fcntl propagates when a FIFO was opened O_RDWR — closing the same propagation gap for the new FIFO backend PR #467 introduced.

Re-ran cargo xtask build + smoke (all 31 markers) locally after rebase.

Issue #485, opened 2026-04-16
kernel: add jump_to_ring3 register / fault diagnostics for #478 smoke flake (#478 split 2/2)


dburkart:
## Motivation

Split 2 of #478 (P0 smoke flake "init: hello from pid 1" missing after IRETQ). To distinguish a silent #GP/#PF at ring-3 entry from a scheduling race or bad stack setup, we need kernel-side diagnostics that log the exact RIP/RSP/CS/SS/RFLAGS values pushed for IRETQ and surface any fault taken on the first userspace instruction. Discovered while triaging #478.

## Work

- [ ] In `kernel/src/arch/x86_64/syscall.rs` `jump_to_ring3`, log RIP, RSP, CS, SS, RFLAGS immediately before IRETQ at debug level so the smoke harness can match them.
- [ ] Ensure the early fault handlers (#GP, #PF, #UD entries in `kernel/src/arch/x86_64/idt.rs`) log enough state to tell if the fault happened at the first ring-3 RIP when one occurs pre-init-output.
- [ ] Register a new unique smoke marker for the log line emitted just before IRETQ, and another inside the #GP/#PF/#UD handler exit path if it fires on a CPL=3 RIP before any userspace marker has been emitted, so the smoke harness can distinguish the failure modes.
- [ ] Add the marker(s) to `xtask/src/main.rs` `SMOKE_MARKERS`.

## Context

- Parent: #478 (P0 smoke flake, area:userspace, track:userspace)
- Files: `kernel/src/arch/x86_64/syscall.rs`, `kernel/src/arch/x86_64/idt.rs`, `kernel/src/init_process.rs`, `xtask/src/main.rs`
- Related failed runs: CI run 24518303695, CI run 24514940883
- Complements split 1/2 (userspace-side markers).

Issue #478, opened 2026-04-16
smoke intermittently fails: userspace init never emits "init: hello from pid 1" after IRETQ


dburkart:
## Motivation
Discovered while merging #477 (which fixed the xtask smoke poll loop). With the new `"init: iretq to ring-3"` diagnostic marker now in `SMOKE_MARKERS`, failed smoke runs show the kernel reaching IRETQ but no userspace output ever appears — the first `write(1, HELLO_MSG)` in `userspace/init/src/main.rs` never produces serial output. This happens ~10-20% of the time on CI (saw it in PR #477's first run, reproducible in earlier failure run 24514940883 on main). The xtask fix from #476 now exits fast on this failure mode (10 min vs. old hang), but the underlying kernel-or-userspace entry bug is still present and surfaces as smoke flakiness. CI flakiness on the PID-1 critical path is a P0.

## Work
- [ ] Add more diagnostic markers inside userspace init (e.g. immediately before the first `write` syscall, and on syscall return) so we can tell whether we re-enter kernel at all.
- [ ] Investigate the SYSRET/IRETQ path in `kernel/src/arch/x86_64/syscall.rs` `jump_to_ring3` — check RFLAGS, CS/SS selectors, RIP/RSP values actually pushed, and whether any #GP/#PF is silently taken on the first userspace instruction.
- [ ] Check for a possible race between the shell task and init task scheduling (ordering/preemption during the first ring-3 transition).
- [ ] Verify userspace ELF load segments are correctly mapped on the failing path (add loader tracking / log p_vaddr, p_memsz, permission bits at load time).
- [ ] Consider whether the RLIMIT-sized guard VMA or the auxv `write-initial-stack` helper has a latent corner case that occasionally corrupts the ring-3 stack.

## Context
- Failed PR #477 run, first attempt: CI run 24518303695 (smoke captured serial showed kernel hit `init: iretq to ring-3` then no userspace output).
- Earlier main failure with the same signature: run 24514940883 (on #465 merge) — reached `init: entering ring-3` then silence.
- Main success runs: e.g. 24517503945 — all 31 markers present including `init: hello from pid 1`.
- Files:
  - `kernel/src/init_process.rs` (`init_ring3_entry`, around lines 199-248)
  - `userspace/init/src/main.rs` (`_start`, around line 61)
  - `kernel/src/arch/x86_64/syscall.rs` (`jump_to_ring3`)
[...]

PR #486, opened 2026-04-16
userspace/init: add pre/post-write markers for #478 smoke flake

### LINKED ISSUE #484: userspace init: add fine-grained diagnostic markers around first write syscall (#478 split 1/2)
## Motivation

## Work

## Context

dburkart:
Closes #484

## Summary
- Adds two new smoke markers to userspace init — `"init: pre-write marker"` (fd=2) emitted on ring-3 entry before the first `write(1, HELLO_MSG)`, and `"init: post-write marker"` (fd=1) emitted immediately after HELLO_MSG returns.
- Registers both markers in `SMOKE_MARKERS` (xtask) so the smoke harness fails a CI run that loses any of the four bracket markers.
- Purpose: split 1/2 of the #478 P0 flake ("init: hello from pid 1" missing after IRETQ). Together with the existing kernel-side `"init: iretq to ring-3"` and the userspace `"init: hello from pid 1"`, these localise where the ring-3 entry path breaks when it breaks:
  - `iretq to ring-3` fires, `pre-write` doesn't → first `SYSCALL` instruction or entry trampoline silently faulted.
  - `pre-write` fires, `hello from pid 1` doesn't → fd=1 specific bug.
  - `hello from pid 1` fires, `post-write` doesn't → write syscall return path is the culprit.

Both markers reuse the existing `write()` helper so they exercise the identical syscall path as HELLO_MSG; a separate minimal-syscall path would add a second regression surface without a diagnostic payoff.

## Test plan
- [x] `cargo xtask build` — succeeds locally.
- [x] `cargo test -p xtask` — 26/26 passing, SMOKE_MARKERS satisfaction tests still green with the two new entries.
- [ ] CI green on `fmt + build`, `host unit tests`, `integration tests (QEMU)`, `smoke test`.
- [ ] Follow-up (#485): kernel-side `jump_to_ring3` register / fault diagnostics for the other half of the diagnostic picture.

### DIFF
--- a/userspace/init/src/main.rs
+++ b/userspace/init/src/main.rs
@@ -54,12 +54,26 @@ use core::panic::PanicInfo;
 /// Smoke-test marker — asserted by `cargo xtask smoke`.
 const HELLO_MSG: &[u8] = b"init: hello from pid 1\n";
 
+/// Emitted immediately on ring-3 entry, before the first `write(1, ...)`.
+/// Localises the #478 flake: if `init: iretq to ring-3` fires but this
+/// marker doesn't, the first `SYSCALL` instruction or entry trampoline
+/// never ran. fd=2 is used so it's separable from HELLO_MSG on fd=1 and
+/// rules out an fd-1-specific bug.
+const PRE_WRITE_MSG: &[u8] = b"init: pre-write marker\n";
+
+/// Emitted right after the first `write(1, HELLO_MSG)` returns.
+/// If PRE_WRITE fires and HELLO_MSG fires but this doesn't, the write
+/// syscall return path is the culprit.
+const POST_WRITE_MSG: &[u8] = b"init: post-write marker\n";
+
 /// Emitted after the parent collects the child's exit status.
 const DONE_MSG: &[u8] = b"init: fork+exec+wait ok\n";
 
 #[no_mangle]
 pub extern "C" fn _start() -> ! {
+    write(2, PRE_WRITE_MSG);
     write(1, HELLO_MSG);
+    write(1, POST_WRITE_MSG);
 
     // fork() — child PID returned to parent; 0 returned to child.
     let fork_ret: i64;
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -98,7 +98,13 @@ const SMOKE_MARKERS: &[&str] = &[
     // `ring3-first-fault:` diagnostic line emitted by the IDT fault
     // handlers in that case).
     "ring3-iretq: rip=",
+    // Fine-grained userspace markers bracketing the first `write`
+    // syscall. See userspace/init/src/main.rs — together with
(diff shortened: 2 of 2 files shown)


dburkart:
**Auto-engineer parking this PR for human review.**

**Park reason:** CI smoke test reliably fails. Reproduced locally under QEMU TCG.

**What the markers revealed (this PR doing its job):**

The `#478` flake is now fully observable end-to-end. With this PR's userspace markers stacked on top of #487's kernel-side diagnostics, the failure localises to a concrete **page fault on the userspace stack**:

```
init: iretq to ring-3
init: pre-write marker           ← fd=2 write succeeds
ring3-first-fault: #PF rip=0x400047 rsp=0x7fffff4c cs=0x23 ss=0x1b rflags=0x202
                   cr2=0x7fffff4c code=PROTECTION_VIOLATION | CAUSED_BY_WRITE | USER_MODE
ring3-iretq: rip=0x400000 rsp=0x7fffff58 ...  ← fork+exec path runs
hello: hello from execed child
init: fork+exec+wait ok
```

So `write(2, PRE_WRITE_MSG)` succeeds, then the next user-mode instruction (around `rip=0x400047`, likely the push/call for the second `write(1, HELLO_MSG)`) takes a `#PF` on a write at `cr2=0x7fffff4c` — 12 bytes below the initial rsp of `0x7fffff58`. Userspace stack is `virt=0x7ffff000 top=0x80000000`, so the address is in range but the page is not writable. Most likely a stack page mapped RO, or mapped without user/ RW permissions, or not backed after the first few pushes.

This is the real `#478` bug: **userspace stack mapping is inconsistent between the initial rsp location and addresses a few bytes below it.** That is out of scope for this PR (which is purely diagnostic markers).

**Attempts across this cycle:**
- SHA `d55fed5` (initial push): smoke failed with missing markers — confirmed flake.
- Job rerun on `d55fed5`: smoke failed identically.
- SHA `7523138` (rebased onto main carrying #487 kernel diagnostics, conflict resolved in `xtask/src/main.rs`): smoke failed — this run is what surfaced the `ring3-first-fault: #PF` diagnostic above.

**What a human reviewer should consider:**
1. The PR changes are 100% diagnostic (two marker constants + a second `write()` call to fd=2). They add no new code path beyond the existing `write()` helper. The fact that they reliably trip `#PF` may indicate the flake was always deterministic under this access pattern, just not previously observable.
2. If the diagnostic value is wanted, one option is to **merge this PR regardless of smoke** (it's the bug the PR is documenting) and let the follow-up `#478` fix make smoke pass again.
3. Alternative: shrink the PR to only the new SMOKE_MARKERS entries (no userspace change) so it doesn't provoke `#PF`; file a separate issue for the stack-mapping bug now that it's localised.

**Commits:** `7523138` (current HEAD on branch `m484-init-pre-post-write-markers`).

Assigning `needs-human-review`; loop continues to next iteration.

PR #487, opened 2026-04-16
arch/x86_64: ring-3 entry + first-fault diagnostics for #478

### LINKED ISSUE #485: kernel: add jump_to_ring3 register / fault diagnostics for #478 smoke flake (#478 split 2/2)

dburkart:
Closes #485

## Summary

Split 2/2 of #478 — kernel-side diagnostics so the next reproduction of the P0 smoke flake leaves enough trace in the serial capture to tell IRETQ failure modes apart.

- New `ring3-iretq: rip=... rsp=... cs=... ss=... rflags=...` line emitted from `jump_to_ring3` at the last instruction before the `iretq` opcode. Registered as a hard `SMOKE_MARKER` so a regression that skips ring-3 entry fails smoke directly, instead of hiding behind a missing userspace marker.
- New `ring3-first-fault: <VECTOR> rip=... rsp=... cs=... ss=... rflags=... <extra>` line emitted from a one-shot `AtomicBool` latch inside the `#UD`, `#GP`, and `#PF` IDT handlers. Fires only when `CPL=3`, and only on the first ring-3 fault of the session, so legitimate SIGSEGV delivery later in the run does not flood the log. Formatting reads only scalars out of the hardware `InterruptStackFrame` and `CR2` — no user memory is touched, so this is safe to call from inside a page-fault handler.

Complements split 1/2 (userspace-side pre/post-write markers) which is in flight as PR #486.

## Decoding a failing run

- `ring3-iretq:` present + `ring3-first-fault:` absent + init marker absent → IRETQ executed, first ring-3 instruction executed without faulting, then something later went missing (scheduling / stdout path).
- `ring3-iretq:` present + `ring3-first-fault:` present → ring-3 entry faulted immediately; the vector + CR2 tell us which mapping is wrong.
- `ring3-iretq:` absent → kernel never called `jump_to_ring3` — the regression is on the init-process / scheduler side.

## Test plan

- [x] \`cargo xtask build\` clean (only pre-existing \`is_guard_hit\` dead-code warning).
- [ ] Full \`cargo xtask test\` + \`cargo xtask smoke\` via CI (local host is aarch64).
- [ ] Next occurrence of the #478 flake exercises the new diagnostic and its capture tells us which branch fired.

### DIFF
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -2,6 +2,8 @@
 //! during early boot. Each handler logs to serial then halts — we don't
 //! have a panic-unwind story yet, so there's nothing to return to.
 
+use core::sync::atomic::{AtomicBool, Ordering};
+
 use spin::Lazy;
 use x86_64::structures::idt::{InterruptDescriptorTable, InterruptStackFrame, PageFaultErrorCode};
 use x86_64::structures::paging::PageTableFlags;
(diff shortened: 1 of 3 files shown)


dburkart:
Nit deferred to a follow-up issue — the `0x202` RFLAGS literal appears twice (serial_println + asm push) and could be pulled into a shared `const IRETQ_RFLAGS`. Not blocking merge; filing as a separate cleanup.

PR #489, opened 2026-04-16
arch/x86_64 + gdbstub: capture full GPR set on #BP via asm trampoline

### LINKED ISSUE #482: gdbstub: capture general-purpose registers under #BP via asm trampoline
## Motivation

Discovered while merging #479. The newly-added int3 to gdbstub diversion only captures `rip`/`rsp`/`rflags`/`cs`/`ss` from the `InterruptStackFrame` — `extern "x86-interrupt"` clobbers GPRs before Rust runs, so `regs.rax..r15` round-trip as zero. A connected `gdb` will therefore show bogus zero values for every GPR when it breaks into a kernel `int3`, which is misleading enough to make the debugger useless for anything but control flow.

## Work

- [ ] Add a naked asm trampoline (e.g. in `kernel/src/arch/x86_64/gdb_trampoline.S` or inline `global_asm!`) that the IDT breakpoint vector points at directly instead of the current `extern "x86-interrupt" fn breakpoint`.
- [ ] The trampoline must push GPRs (rax, rbx, rcx, rdx, rsi, rdi, rbp, r8..r15) and pass a pointer to the saved block into a Rust entry that fills `GdbRegs` fully (including `rsp` from the interrupt frame, not the saved `rsp` which is the kernel stack at trampoline entry).
- [ ] On resume, the trampoline restores any GPRs the debugger mutated via `G`, then `iretq`.
- [ ] Extend `gdbstub::dispatch` (or a successor handler) so `G` writes back the mutated GPRs in addition to `rip`/`eflags`. With full capture, the "honest G" logic that currently rejects non-`rip`/`eflags` writes with `E01` should be loosened to accept all supported registers.
- [ ] Unit test: a `G` that sets `rax` round-trips through the dispatch path and is visible on a subsequent `g`.

## Context

- Caused by `kernel/src/arch/x86_64/idt.rs` breakpoint handler added in 6dd5fed (PR #479).
- Follow-up of #446 — the parent issue was scoped to packet-level register support; GPR capture was called out as out of scope in the PR body's "Out of scope (follow-up)" section.
- Related: `kernel/src/gdbstub/regs.rs` `GdbRegs` struct already has slots for all 16 GPRs.
- Note on labels: the filing agent initially suggested `area:arch`, but that label does not exist in the repo taxonomy and the prioritization playbook forbids inventing new labels in-flight. Filed under `area:debug` to match the sibling gdbstub issues (#447, #448). `priority:P2` per the suggestion — gdb is a debug tool, not a kernel boot blocker; the framework lands in #446 and this is quality-of-life on top.

dburkart:
Closes #482.

## Summary

- Adds a naked asm trampoline at `kernel/src/arch/x86_64/gdb_trampoline.rs` that pushes all 15 caller/callee-saved GPRs before Rust runs, passes pointers to the saved block + hardware interrupt frame into a `sysv64` Rust entry, and on resume pops any mutations and `iretq`s.
- Rewires the IDT `#BP` vector via `set_handler_addr` instead of `set_handler_fn`, so the compiler-generated prologue can no longer clobber rax..r15 before the stub reads them. Historical int3 semantics are preserved — the Rust entry early-returns unless the stub is armed and the trap came from ring 0.
- Loosens the gdbstub `G` dispatch to accept all 17 u64 + eflags + segment fields now that GPRs round-trip through the trampoline, replacing the `E01`-on-any-GPR-change guard.
- Replaces the host unit test `big_g_rejects_unsupported_fields` with `big_g_writes_rax`, exercising the new round-trip.

CS/SS/RSP writeback to the hardware frame is deliberately omitted — mutating those without a coordinated task switch is unsound. The stub still accepts the writes into its in-memory snapshot (a subsequent `g` reads them back) and replies `OK`.

## Test plan

- [ ] `cargo xtask build` — clean (only a pre-existing `is_guard_hit` dead-code warning in `task/task.rs`).
- [ ] `cargo xtask test` — `gdbstub_loop::big_g_writes_rax` passes under the new dispatch; all existing gdbstub tests (`framer_checksum`, `question_then_detach`, `unknown_returns_empty`, `g_reply_is_hex_blob`, `big_g_writes_regs`) still pass.
- [ ] `cargo xtask smoke` — kernel still boots through init; no regression in the ring-3 path (int3 is not on any smoke codepath).

### DIFF
--- a/kernel/src/arch/x86_64/gdb_trampoline.rs
+++ b/kernel/src/arch/x86_64/gdb_trampoline.rs
@@ -155,11 +155,11 @@ global_asm!(
     "mov rdi, rsp",
     // rsi = &HwFrame (just above the 15 GPR pushes = 120 bytes)
     "lea rsi, [rsp + 15*8]",
-    // Align stack to 16 bytes across the call: 15 u64 pushes + 5 hw
-    // pushes = 20 u64 = 160 bytes = already 16-byte aligned on entry.
-    // But System V ABI requires rsp % 16 == 8 *before* the call
-    // instruction (call pushes 8). So we need an extra push to get
-    // there.
+    // Align stack to 16 bytes across the call. System V requires
+    // rsp % 16 == 0 at the `call` instruction (call then pushes the
+    // 8-byte return address, so inside the callee rsp % 16 == 8).
+    // After the 15 GPR pushes (120 bytes) on top of the 5 hw pushes
+    // (40 bytes), rsp % 16 == 8 — one 8-byte sub gets us to 0.
     "sub rsp, 8",
     "call gdb_breakpoint_entry",
     "add rsp, 8",
--- a/kernel/src/arch/x86_64/gdb_trampoline.rs
+++ b/kernel/src/arch/x86_64/gdb_trampoline.rs
@@ -1,0 +1,190 @@
+//! Naked asm trampoline for the `#BP` (int3) vector. Captures the full
+//! GPR set on entry so `gdbstub` can show accurate register values —
+//! the old `extern "x86-interrupt"` handler let the compiler-generated
+//! prologue clobber rax..r15 before Rust ran, leaving them as zero.
+//!
+//! Layout on the kernel stack at trampoline entry (low→high address):
+//!   [rsp+0]   RIP  (hw-pushed)
+//!   [rsp+8]   CS
+//!   [rsp+16]  RFLAGS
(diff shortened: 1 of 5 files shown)


dburkart:
Thanks for the review! Applied the ABI alignment comment fix in 2a01332. Deferring the non-canonical-rip serial log suggestion to a follow-up issue — it's a diagnostic improvement on an already-safe path (we reply OK and silently drop the bogus write; a warning would help UX but isn't load-bearing for correctness).

PR #495, opened 2026-04-16
block: add set-associative read cache with write-invalidate

### LINKED ISSUE #82: virtio-blk: block layer caching
## Motivation

Issue #43 deliberately skips caching to keep the initial virtio-blk + ramdisk implementation simple. Once reads (and eventually writes) are working, a block cache will be needed for reasonable performance.

## Scope

- Implement a simple in-memory block cache (e.g., fixed-size LRU or direct-mapped) sitting between the VFS and the raw virtio-blk driver.
- Cache `VIRTIO_BLK_T_IN` results; invalidate on writes.
- Expose cache hit/miss stats (useful for debugging).

## Out of scope here

- Write-back / dirty page flushing (coordinate with write support once that exists).
- Multi-level or adaptive replacement policies.

## Notes

Explicitly deferred from #43. Depends on write support (#81) before write-back caching makes sense. A read-only cache can be added independently once the read path is stable.

dburkart:
Closes #82

## Summary
- New `block::cache` module — 32-line 4-way set-associative cache (4 KiB lines, 128 KiB `.bss` total) sitting between the public `block::{read,write}` facade and `virtio_blk::{read,write}`.
- Reads are routed through `cache::read_through`: hits copy from the resident page, misses pull one page from the device and fill. LRU eviction within each set.
- Writes are routed through `cache::write_invalidate`: every cache line overlapping the written LBA range is dropped, then the write goes straight to the driver. Write-back / dirty flushing is out of scope per #82's stated bounds.
- Monotonic hit/miss/invalidation/fill counters exposed via `block::cache_stats()` — lock-free atomics, safe to sample from debug paths.
- Cache core is generic over a `Backend` trait; `VirtioBackend` (kernel-only) delegates to `virtio_blk`. Host tests use a `MockDisk` that counts read/write calls, so the cache can be exercised without QEMU.

## Test plan
- [x] `cargo xtask build` passes on aarch64 cross-compile.
- [ ] CI: `cargo xtask test-unit` runs new host tests (`miss_then_hit`, `write_invalidates_overlapping_line`, `write_outside_cached_range_no_invalidation`, `eviction_is_lru_within_set`, `multi_page_unaligned_read`, `bad_align_rejected`, `write_spanning_two_pages_invalidates_both`).
- [ ] CI: `cargo xtask test` integration suite — unchanged surface (`block::read` / `block::write` signatures are identical) so existing VFS-through-block tests exercise the cached path for free.
- [ ] CI: `cargo xtask smoke` — main.rs probe (write then read of a fixed LBA) now round-trips through the cache. Identical data must come back.

### DIFF
--- a/kernel/src/block/cache.rs
+++ b/kernel/src/block/cache.rs
@@ -122,10 +122,8 @@ impl<const LINES: usize> Cache<LINES> {
         while cur < end {
             let page_lba = cur & !(LINE_SECTORS - 1);
             let intra_sector = (cur - page_lba) as usize;
-            let sectors_in_page = core::cmp::min(
-                LINE_SECTORS as usize - intra_sector,
-                (end - cur) as usize,
-            );
+            let sectors_in_page =
+                core::cmp::min(LINE_SECTORS as usize - intra_sector, (end - cur) as usize);
             let bytes = sectors_in_page * SECTOR_SIZE;
             let src_off = intra_sector * SECTOR_SIZE;
 
--- a/kernel/src/block/cache.rs
+++ b/kernel/src/block/cache.rs
@@ -116,7 +116,11 @@ impl<const LINES: usize> Cache<LINES> {
             return Err(BlkError::BadAlign);
         }
         let sectors = (buf.len() / SECTOR_SIZE) as u64;
-        let end = lba + sectors;
+        // Reject ranges that would wrap past u64::MAX. Without this a
+        // caller passing `lba` near the top of the address space could
+        // walk a wrapped `end`, return Ok(()) with an unfilled buffer,
+        // or (in invalidate) iterate the wrong pages unboundedly.
+        let end = lba.checked_add(sectors).ok_or(BlkError::BadAlign)?;
         let mut cur = lba;
         let mut out_off = 0;
         while cur < end {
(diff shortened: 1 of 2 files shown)


dburkart:
Auto-engineer is parking this PR for human triage. Exhausted the 3-fix-round budget without a green CI.

## Status

- Host unit tests: ✅
- fmt + release build: ✅
- QEMU integration tests: ✅
- **Smoke test: ❌** (reproducible, not a flake)

## Failure signature

```
Error: "smoke: missing markers [\"init: hello from pid 1\"]"
```

Kernel boots, probes virtio-blk OK (`block: write+readback ok at lba 2047`), enters ring 3 via IRETQ, then silence for 10 minutes. Userspace `write(1, "init: hello from pid 1\n")` (syscall 1 at `userspace/init/src/main.rs:55`) never produces output. Other concurrent PRs' smoke runs pass; `main` smoke is green at run 24529732210 — so the regression is caused by this PR, not infra.

## Commits

- 398b446 — initial cache implementation + wiring
- 8032884 — `cargo fmt` fix for CI
- 7ce1710 — CodeRabbit Critical: `checked_add` overflow guard on LBA+sectors

## Hypothesis (unverified)

Adding `static CACHE: Mutex<Cache<TOTAL_LINES>>` contributes ~131 KiB to `.bss` (32 lines × 4112 bytes). This may shift the kernel image layout such that user PML4 construction in `AddressSpace::new_empty()` (`kernel/src/mem/addrspace.rs:121`) copies a different view of the kernel high-half — breaking the path the trampoline at `kernel/src/arch/x86_64/syscall.rs:157` needs to re-enter the kernel on user syscall.

This is a hunch; I could not bisect locally (dev host is aarch64 and the repo's memory instructs skipping local smoke). A fix-round 3 shrinking the cache would be speculative.

## Suggested next steps for human reviewer

1. Bisect: does reducing `TOTAL_LINES` (e.g., 32 → 8, trimming `.bss` by ~100 KiB) restore userspace entry? If yes, confirms layout sensitivity and points at a deeper bug in boot/paging.
2. If no: check the kernel ELF LOAD segments for boundary alignment differences between `main` and this branch.
3. Alternative: move `CACHE` behind a heap-allocated `OnceCell` so the cache backing comes from the allocator rather than `.bss`.

PR #500, opened 2026-04-16
signal: preserve user syscall arg regs across SA_RESTART+handler sigreturn

### LINKED ISSUE #499: signal: preserve user syscall arg regs across SA_RESTART+handler sigreturn path
## Motivation

Discovered while merging #497. That PR made bare `KERN_ERESTARTSYS` restarts (no user signal handler) replay with the original syscall registers by saving `rax, rdi, rsi, rdx, r10, r8, r9` on the kernel stack and restoring them before `SYSRETQ` when `SYSCALL_RESTART_PENDING` is set. The `SA_RESTART + Disposition::Handler` path, however, reroutes the return to the user handler first; the restart happens only after `sigreturn`. The sigreturn path today restores `user_rip / user_rflags / user_rsp` from the saved `SigFrame`'s `uc_mcontext` but does NOT restore the syscall arg registers — so a SA_RESTART-style restart routed through a user handler still re-executes the `SYSCALL` instruction with clobbered `rax/rdi/.../r9`. This is a known gap explicitly deferred in the commit message for #497.

## Work

- [ ] Extend the `SigFrame`'s `uc_mcontext` (`kernel/src/signal/frame.rs`) to save all 7 user syscall regs alongside the existing `rip/rflags/rsp` when a SA_RESTART+handler delivery occurs.
- [ ] Teach `sys_sigreturn` / the sigreturn path in `kernel/src/signal/mod.rs` to restore those 7 regs into the saved `SyscallReturnContext` so the post-sigreturn `SYSRETQ` re-executes `SYSCALL` with the original state.
- [ ] Also consider whether the trampoline should set `SYSCALL_RESTART_PENDING` on the handler path (or a sigreturn-side equivalent) so the existing restore arm in the asm does the work.
- [ ] Add a QEMU integration test that installs a SA_RESTART handler on `SIGTTOU`, triggers the tty tostop gate, verifies the handler runs, and verifies the restarted `write` replays with the original fd / buffer / length.

## Context

- Parent PR: #497 (merged as commit `5fcc354`).
- Closed issue: #451.
- Key files:
  - `kernel/src/signal/mod.rs` (`check_and_deliver_signals`, `signal_to_deliver`, `sys_sigreturn`)
  - `kernel/src/signal/frame.rs` (`SigFrame` / `push_signal_frame` / `restore_signal_frame`)
  - `kernel/src/arch/x86_64/syscall.rs` (`syscall_entry` asm, the `SYSCALL_RESTART_PENDING` atomic at ~line 107, the restore path around the `cmp qword ptr [rip + {restart_pending}], 0`).

dburkart:
Closes #499.

## Summary

- #497 landed the no-handler `SYSCALL_RESTART_PENDING` arm that reloads `rax/rdi/rsi/rdx/r10/r8/r9` from the kernel stack before `SYSRETQ`, but explicitly deferred the SA_RESTART+handler case — where the restart happens only *after* `sigreturn` returns. Without a fix, the restarted `SYSCALL` re-executes at the rewound rip with clobbered arg regs.
- The SigFrame's `mcontext_t gregs` array already has slots for those GPRs (Linux indices 0/1/2/8/9/12/13). This PR populates them on `push_signal_frame` and restores them on `sys_sigreturn` via new `SIGRET_USER_*` shuttle statics, then arms the existing asm restore path so the user GPRs reach `SYSRETQ`.
- A new `UC_VIBIX_SYSCALL_REGS` bit in `uc_flags` distinguishes syscall-return frames from fault frames and from non-restart handler deliveries — only the SA_RESTART+handler restart path sets it, so the other paths stay byte-identical (avoids a spurious zero-fill of user GPRs on a `#PF`-delivered sigreturn, and avoids clobbering a syscall's return value in `rax` on a non-restart handler sigreturn).

## Test plan

- [x] `cargo xtask test-integration` — full QEMU integration suite green, including the new 4-test `signal_sarestart_handler_regs` binary that pins the push/restore contract (Some→Some roundtrip, None→None on fault and non-restart-handler frames, rip/rflags/rsp/saved_mask survive alongside).
- [x] `cargo xtask smoke` — all 33 markers present.
- [x] Host unit tests in `kernel/src/signal/frame.rs` extended with a greg-index distinctness assertion.
- End-to-end SA_RESTART-handler-replay of a real syscall (e.g. `SIGTTOU` handler on top of a tostop-gated `write`) requires ring-3 job-control and is deferred to the shell-level smoke suite once that lands.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -267,3 +267,7 @@ harness = false
 [[test]]
 name = "gdbstub_loop"
 harness = false
+
+[[test]]
+name = "signal_sarestart_handler_regs"
+harness = false
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -100,6 +100,35 @@ pub static FORK_USER_RSP: AtomicU64 = AtomicU64::new(0);
 #[no_mangle]
 pub static SIGRETURN_PENDING: AtomicU64 = AtomicU64::new(0);
 
+/// Set to 1 by the SIGRETURN dispatch arm when the SigFrame carried saved
+/// user syscall arg regs (i.e. was pushed on the syscall-return path).
+/// Tells `check_and_deliver_signals` to both repopulate ctx's arg-reg
+/// slots from the `SIGRET_USER_*` statics *and* arm `SYSCALL_RESTART_PENDING`.
+/// On a fault-delivered signal's sigreturn, this stays 0 and the arg-reg
+/// slots in ctx are left alone — avoiding a spurious zero-fill of user
+/// GPRs. Cleared after consumption.
+#[no_mangle]
+pub static SIGRET_HAS_SYSCALL_ARGS: AtomicU64 = AtomicU64::new(0);
+
+/// User syscall arg regs (rax/rdi/rsi/rdx/r10/r8/r9) shuttled from
+/// `sys_sigreturn` to `check_and_deliver_signals` so the sigreturn path
+/// can repopulate the kernel-stack `SyscallReturnContext` before SYSRETQ.
+/// Only meaningful when `SIGRET_HAS_SYSCALL_ARGS == 1`.
+#[no_mangle]
+pub static SIGRET_USER_RAX: AtomicU64 = AtomicU64::new(0);
+#[no_mangle]
+pub static SIGRET_USER_RDI: AtomicU64 = AtomicU64::new(0);
+#[no_mangle]
+pub static SIGRET_USER_RSI: AtomicU64 = AtomicU64::new(0);
+#[no_mangle]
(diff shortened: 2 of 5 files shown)


dburkart:
Addressed in 928a6d4. The nested-delivery window between sigreturn and SYSRETQ is now handled in `check_and_deliver_signals`: snapshot `SYSCALL_RESTART_PENDING` after processing the sigreturn branch, and if it's armed while a fresh signal is about to be delivered, treat it as a restart-handler (propagate args onto the new frame via `UC_VIBIX_SYSCALL_REGS`) and consume the flag. The nested handler's eventual sigreturn re-arms from its own frame. Added frame-layer test `nested_delivery_propagates_syscall_args` in `signal_sarestart_handler_regs.rs` covering the two-hop round-trip. Full ring-3 coverage remains in the shell-level job-control smoke suite.

Issue #507, opened 2026-04-17
CI: 100× smoke soak as acceptance gate for fork flakiness (epic #501)


dburkart:
## Motivation

Parent epic: #501. Issue #303 explicitly deferred a 100× smoke soak as the acceptance gate for the original flakiness work, and #460 raised the CI HARD_CAP from 180s to 300s partly because single-run smoke was too noisy to interpret. Without a soak test, any fix we land in sub-issues #3/#4 could still harbor a 1% flake and we'd never know.

This issue adds a CI job that runs `cargo xtask smoke` 100 times per invocation; the job must be 100% clean for 3 consecutive nightly runs before the epic #501 can close.

## Work

- [ ] Add a `cargo xtask smoke-soak` subcommand that runs the existing smoke target N=100 times (configurable via `--count`), collects per-run pass/fail, and exits nonzero if any run fails. Aggregate per-marker failure counts in the summary.
- [ ] Add a new GitHub Actions workflow `.github/workflows/smoke-soak.yml` that runs `cargo xtask smoke-soak --count 100` on a nightly schedule (or on-demand via `workflow_dispatch`). Upload the full serial log from each failed run as an artifact.
- [ ] Add a manual trigger path — on push to any branch with `soak` in the name — so developers can stress-test a change without waiting for nightly.
- [ ] Document in `docs/agent-playbooks/` (or similar) that the soak is the acceptance gate for flakiness-related fixes.
- [ ] Once epic #501's structural fixes land, run the soak 3 consecutive nights; only then close epic #501.

## Testing

- The xtask subcommand should return exit code 0 when all 100 runs pass, non-zero with a summary when any fail.
- The workflow should succeed on a clean main (post-fix); fail with useful artifacts on a flake.

## Out of scope

- Changes to the smoke markers themselves (sub-issue #420 covers that).
- Changes to what the smoke test exercises.

## Context

- Epic: #501.
- Original deferral: #303.
- CI HARD_CAP history: #460 / #464 / #465.
- Existing smoke: `xtask/src/main.rs` (smoke harness + marker list at :49-79, :821-889).
- Existing workflows: `.github/workflows/` (current CI config).

Issue #506, opened 2026-04-17
reproducer: init-driven fork+exec+wait loop with heartbeat panic (epic #501)


dburkart:
## Motivation

Parent epic: #501. The current smoke test runs init once: `fork → execve → wait4 → print markers → exit`. This gives exactly one fork per boot, which is both slow to reproduce flakes (minutes per run) and provides no amplification — a 1%-rate bug takes ~100 boots to catch once.

A deterministic reproducer — init spawning N fork+exec+wait cycles in a tight loop, with a heartbeat assertion that panics if any cycle stalls beyond a threshold — makes the hang reproduce in seconds locally and lets sub-issue #1's investigation be empirical, not speculative. It also becomes a regression gate for sub-issues #3/#4 to prove the fix.

## Work

- [ ] Add a `fork_loop` userspace binary under `userspace/fork_loop/` that:
  - In a loop of N=500 (configurable via compile-time const): `fork`; child `execve`s a minimal `userspace_tick.elf` that writes `tick: i=N\n` and `exit(0)`; parent `wait4`s the child.
  - Between cycles, increment a counter and `write(1, ...)` every 50 cycles (heartbeat).
  - Each cycle also records timestamp via `rdtsc`-style read (if userspace has access); if a cycle takes >1s of wall time (via a simple PIT-tick-count read syscall, or just the kernel's existing `time::jiffies` exposed as a new syscall), panic-log and exit.
- [ ] Build `fork_loop` as an alternative init binary; add an xtask command `cargo xtask repro-fork` that builds the ISO with `fork_loop` as init and runs under QEMU.
- [ ] Add kernel-side panic: after N jiffies with no progress from init, panic with "fork loop stuck" so QEMU exits and the hang is observable as a nonzero exit code.
- [ ] Document in `docs/agent-playbooks/` or a new `docs/repros/fork-loop.md` how to run it.

## Testing

- On a clean HEAD (with the current known flake), `cargo xtask repro-fork` should hang within 500 cycles about half the time — matches the observed boot rate.
- After sub-issue #3 + #4 land, `cargo xtask repro-fork` must complete 500 cycles clean 20 consecutive runs.

## Out of scope

- CI integration (sub-issue #6 handles that).
- The actual root-cause (sub-issue #1).

## Context

- Epic: #501.
- Existing init: `userspace/init/src/main.rs`.
- Existing hello (for execve target reuse): `userspace/hello/src/main.rs`.
- xtask entry: `xtask/src/main.rs`.

Issue #505, opened 2026-04-17
structural: single choke point for syscall_stack_top / TSS update on context switch (epic #501)


dburkart:
## Motivation

Parent epic: #501. PR #206 added a per-task `syscall_stack_top` and the requirement that every context switch update `SYSCALL_KERNEL_RSP` and `TSS.rsp[0]` to the incoming task's stack. The update is currently duplicated across **three** distinct switch sites in `kernel/src/task/mod.rs` (lines 778, 973, 1094 — the preempt, block, and yield paths). If any future switch path forgets the update, a task entering SYSCALL will land on the previous task's kernel stack and corrupt it — a subtle and catastrophic class of bug.

Consolidating to a single helper (called from each switch path) eliminates the drift risk.

## Work

- [ ] Add a helper, e.g. `fn set_active_syscall_stack(task: &Task)` in `kernel/src/task/mod.rs`, that sets both `SYSCALL_KERNEL_RSP.store(top, Relaxed)` and `TSS.rsp[0] = top`. Use `task.syscall_stack_top` as the source of truth.
- [ ] Replace the three inlined copies in task/mod.rs (:778, :973, :1094) with calls to the helper.
- [ ] Also audit `init_process.rs:229`-area setup + `kernel/src/arch/x86_64/syscall.rs:143-147` (`setup_ring3_stacks`) for similar inlined updates; if they exist, consolidate.
- [ ] Add a `debug_assert!(current.syscall_stack_top != 0)` inside the helper when the task is a ring-3 task (identifiable by whether `fd_table` is user-visible or similar) — catches a bug where a task that should have a syscall stack has 0.
- [ ] Add a `cfg!(debug_assertions)` post-switch check: confirm `rdmsr(IA32_KERNEL_GS_BASE)` and TSS.rsp[0] agree with the helper's last write. If cheap, leave it on in release.

## Testing

- `cargo xtask test` + `cargo xtask smoke` both pass.
- Unit test: simulate a switch sequence (kernel-only task → ring-3 task → ring-3 task → kernel-only) and assert the helper was called with the right stack top each time.
- Integration: run a small in-kernel test that spawns two ring-3-ish tasks under the scheduler, forces context switches between them, and confirms the TSS.rsp[0] after each switch matches the incoming task's `syscall_stack_top`.

## Out of scope

- Replacing `FORK_USER_*` globals (sub-issue #3).
- Changing the syscall entry ABI.

## Context

- Epic: #501.
- Sites to consolidate:
[...]

Issue #504, opened 2026-04-17
structural: replace FORK_USER_* globals with per-task saved state (epic #501)


dburkart:
## Motivation

Parent epic: #501. PR #206 introduced three `AtomicU64` globals — `FORK_USER_RIP`, `FORK_USER_RFLAGS`, `FORK_USER_RSP` (`kernel/src/arch/x86_64/syscall.rs:81-95`) — written by the SYSCALL entry trampoline on *every* syscall and read asynchronously by the `fork()` handler (:504-506), signal machinery (`kernel/src/signal/mod.rs:707-713`), and `sigreturn` (syscall.rs:611-613). If any syscall handler yields before the reader loads these globals, another task's SYSCALL entry clobbers them, and the reader gets stale data.

For fork, this means the child gets primed with the wrong user RIP/RSP/RFLAGS and `sysret`s into a garbage address — silently faulting. This is the H2 race hypothesized in #303, never fixed.

Replacing the globals with per-task saved state removes the race by construction, regardless of whether it is the root cause of the current hang (sub-issues #1/#2 will confirm or reject). This is also called out by #314 as the exact class of `spin::Mutex`/static state that needs auditing.

## Work

- [ ] Design choice (document in a comment before coding): choose between
  - **Option A: save on kernel stack.** SYSCALL trampoline pushes user RIP/RFLAGS/RSP onto the per-task kernel stack as part of its entry frame. Handlers read by offset. Pros: matches Linux. Cons: fork handler has to find them via rsp-relative arithmetic.
  - **Option B: `Task` field.** Each `Task` has `user_regs: UserSavedRegs { rip, rflags, rsp }`. SYSCALL trampoline stores via `gs:[offset]` or a per-CPU pointer. Pros: Rust-friendly access. Cons: needs stable per-task pointer in the trampoline.
- [ ] Implement the chosen option. Update all call sites:
  - SYSCALL entry trampoline (`kernel/src/arch/x86_64/syscall.rs` asm block, currently using `FORK_USER_*` as `sym`).
  - `fork()` handler (syscall.rs:504-506).
  - `sigreturn` (syscall.rs:602-616).
  - `check_and_deliver_signals` (`kernel/src/signal/mod.rs:707-713`).
- [ ] Delete the three `#[no_mangle] pub static FORK_USER_*: AtomicU64` declarations.
- [ ] Add a compile-time `#[deny(unused)]` or similar to catch any forgotten reader.
[...]

PR #514, opened 2026-04-17
diag(fork): instrument fork syscall path + capture trace log (#502)

### LINKED ISSUE #502: diagnostic: root-cause init fork→exec→wait hang (epic #501)
## Motivation

Parent epic: #501. The shell hangs ~50% of boots on HEAD. Captured serial log (see epic) shows init's `write(1, "init: hello from pid 1\n")` completes, then **nothing** prints for the rest of the run. The next userspace instruction after that write is `syscall(57) = fork()`. No further kernel or userspace output appears, and PS/2 / serial input are also dead. Strong signal that fork's syscall handler is spinning with IF=0 and never yielding.

This issue is the **root-cause investigation** phase. The goal is to produce a serial log that pinpoints exactly *where* in the fork path the hang occurs — so the structural fix in sub-issue #3 (epic) can target the actual race, not a hypothesis.

## Work

- [ ] Add `serial_println!` probe points (guarded by a compile-time `FORK_TRACE` feature flag or always-on for now) at the following locations in `kernel/src/arch/x86_64/syscall.rs` and `kernel/src/task/mod.rs` / `kernel/src/task/task.rs`:
  - SYSCALL entry trampoline: after `FORK_USER_*` store, log `[syscall:enter nr=N user_rip=... user_rsp=...]`.
  - `syscall_dispatch` nr=57 entry: log `[fork:enter parent_pid=...]`.
  - `fork_current_task` — log at each major step: stack-slot alloc, `map_range`, `fork_address_space` entry/exit, `Task::new_forked` entry/exit, `process::register`, return.
  - Child's first schedule: add a log in `fork_child_sysret` (push a marker byte to COM1 directly from asm, since we can't call Rust before sysretq cleanly — or log from `Task::new_forked` priming that the child stack is primed with `sysret_target=<rip>`).
  - `execve` (nr=59) entry, `clear_for_exec`, `load_user_elf_with_vmas`, `jump_to_ring3`.
  - `wait4` (nr=61) entry, each `reap_child` attempt, each `wait_while` entry/exit.
- [ ] Build + boot under QEMU locally. If the hang is 50% reproducible, boot in a tight loop (e.g. `cargo xtask run` in a shell `for` loop) until a hang is captured.
- [ ] Attach the captured serial log + the probe-point build to this issue as a comment.
- [ ] Identify the last probe that fires before the hang. That is the root-cause site.
- [ ] Report findings in a comment on this issue + on the epic #501.

## Out of scope

- Actually fixing the hang (that's sub-issue #3 / #4 of the epic).
- Replacing `FORK_USER_*` globals (sub-issue #3).
- Soak test / reproducer harness (sub-issue #5).

## Testing

- Probe build must not break `cargo xtask test` or `cargo xtask smoke` (non-fork tests must still pass).
- Reproducer: boot the probe build ≥10 times locally; capture at least one hang with a clear "last probe" signal.

## Context

- Epic: #501.
- Prior investigation (H2, deferred): #303.
- Files to instrument:
  - `kernel/src/arch/x86_64/syscall.rs:81-95, 504-506, asm trampoline`
  - `kernel/src/task/task.rs:299-401` (`Task::new_forked`)
  - `kernel/src/task/switch.rs:93-113` (`fork_child_sysret`)
  - `kernel/src/task/mod.rs` (context-switch paths at 778/973/1094)
  - `kernel/src/process/mod.rs` (`wait4` condvar path)

dburkart:
Closes #502. Wave 1 of epic #501.

## Summary

Adds bracketed `serial_println!` probes at every major step of the fork syscall
handler so the last surviving probe in a captured serial log pinpoints the stuck
lock / wedged code path. Pure diagnostic work — no structural fixes (those are
tracked in #504 and #505).

New `fork_trace!` macro (in `kernel/src/arch/x86_64/fork_trace.rs`) expands to
a `serial_println!("ftrace: ...")` under `cfg(debug_assertions)` and to a nop
under `--release`. Probe points:

- `syscall_dispatch` `FORK`/`EXECVE`/`WAIT4` arm entry/exit + `CHILD_WAIT`
  park/wake transitions. FORK also records ring-0 RFLAGS on entry (with IF
  broken out — expected to be 0 because `SFMASK` clears it on SYSCALL entry).
- `fork_current_task`: every lock boundary and each call into a sub-step
  (`fork_address_space`, `Task::new_forked`, `process::register`,
  `SCHED.push_ready`).
- `fork_address_space`: child PML4 allocation, VMA snapshot, rollback path.
- `Task::new_forked`: stack slot alloc, `map_range`, stack priming (with the
  captured `fork_child_sysret` address), FPU clone.
- `process::register`: entry, `TABLE.lock()` acquisition, exit.
- `fork_child_sysret`: emits a single `'C'` byte straight to COM1 via
  `in al, dx; out dx, al` before touching any Rust state. Always-on (3
  instructions, one-shot per `fork()`) — tripwire for any future regression
  that breaks child first-schedule delivery.

## Key findings (captured in `docs/investigations/502-fork-trace.md`)

1. `IF=0` confirmed on fork dispatch entry — SFMASK is doing its job.
2. The `'C'` byte from `fork_child_sysret` reaches the serial log right
   before the child's `execve` ftrace line, confirming context_switch
   delivers the child to its first ring-0 instruction correctly.
3. Full walk through `fork_current_task → fork_address_space →
   Task::new_forked → process::register` completes cleanly, with every
   lock bracketed.
4. **The 50% fork hang did not reproduce across 50 instrumented local QEMU
   boots.** Lengthening the critical section with extra `serial_println!`
   calls masks the race. This strongly indicates H2 from #303
   (`FORK_USER_{RIP,RFLAGS,RSP}` globals clobbered between parent's fork
   entry and child's first schedule), not H1 (spinlock held with IF=0
   inside `fork_current_task`). The structural fix is #504 / #505, held
   for wave 3 of the epic.

Captured serial log: `docs/investigations/502-fork-trace-happy-path.log`
(106 lines, includes full `ftrace:` sequence end-to-end).

## File overlap discipline

Kernel-side only. `userspace/init/src/main.rs` (touched by #506) and
`.github/workflows/` (touched by #507) are not modified.

## Test plan

- [x] `cargo xtask build` — clean (one pre-existing dead-code warning)
- [x] `cargo xtask test` — passes, including `fork_cow_write_divergence`
      which emits the new `fork_address_space` ftrace lines inline
- [x] `cargo xtask smoke` — all 33 markers present (soft markers
      `hello: hello from execed child` and `init: fork+exec+wait ok`
      present)
- [x] 30x local QEMU boots of instrumented `vibix.iso` — 30/30
      `init: fork+exec+wait ok`
- [x] 20x local QEMU boots — 19/20 passed; the 1 failure was an
      unrelated pre-fork init hang (no `init: hello from pid 1`),
      matching #478's failure mode, not #502's

### DIFF
--- a/docs/investigations/502-fork-trace-happy-path.log
+++ b/docs/investigations/502-fork-trace-happy-path.log
@@ -1,0 +1,106 @@
+limine: Loading executable `boot():/boot/vibix`...
+limine: Loading module `boot():/boot/userspace_init.elf`...
+limine: Loading module `boot():/boot/userspace_init.elf`...
+limine: Loading module `boot():/boot/userspace_hello.elf`...
+limine: Loading module `boot():/boot/rootfs.tar`...
+limine: Loading module `boot():/boot/ld-musl-x86_64.so.1`...
+vibix booting…
+memory map: 28 entries, 244 MiB usable
+hhdm offset: 0xffff800000000000
+rsdp: 0xf52b0
+cpu: SSE SSE2 SSE4.1 SSE4.2 AVX AVX2 XSAVE PAT SMEP SMAP FSGSBASE RDTSCP POPCNT LZCNT RDRAND RDSEED
+uaccess: smep=on smap=on
+fpu: FXSAVE context switch online
+syscall: SYSCALL/SYSRET enabled
+PIC remapped to 0x20/0x28 and masked
+GDT + IDT loaded
+paging: mapper online
+heap: 1024 KiB @ 0xffffc00000000000 (reserved 16384 KiB, grows in 64 KiB chunks)
+paging: IST guard installed @ 0xffffffff804f0000
+paging: switched to kernel PML4
+paging: reclaimed 4132 KiB bootloader memory (17 L3/L2/L1 intermediate table frames, PML4=0xff43000)
+acpi: MADT parsed (cpus=1, ioapics=1, overrides=5, lapic=0xfee00000)
+apic: BSP online (lapic_id=0)
+ioapic: initialized (1 controller(s))
+ioapic: IRQ0 -> gsi 2 -> vec 0x20 on lapic 0 (ioapic 0)
+ioapic: IRQ1 -> gsi 1 -> vec 0x21 on lapic 0 (ioapic 0)
+ioapic: IRQ4 -> gsi 4 -> vec 0x24 on lapic 0 (ioapic 0)
+serial: rx irq enabled
+hpet: initialized (period=10000000fs, timers=3, base=0xfed00000)
+hpet: periodic timer armed (100 Hz, 1000000 ticks/period)
+timer: 100 Hz
+rtc: 2026-04-17 14:35:32
+pci: 7 device(s) on bus 0
+pci:   00:00.0 8086:29c0 class 06:00 pi=00 (host-bridge)
+pci:   00:01.0 1234:1111 class 03:00 pi=00 (vga)
+pci:   00:02.0 8086:10d3 class 02:00 pi=00 (network)
+pci:   00:03.0 1af4:1001 class 01:00 pi=00 (scsi)
+pci:   00:1f.0 8086:2918 class 06:01 pi=00 (isa-bridge)
+pci:   00:1f.2 8086:2922 class 01:06 pi=01 (sata)
+pci:   00:1f.3 8086:2930 class 0c:05 pi=00 (serial-bus)
+block: virtio-blk ready (io_base=0xc000, queue_size=256)
+block: lba0[0..16]=[56, 49, 42, 49, 58, 2a, 4c, 4b, 30, 00, 00, 00, 00, 00, 00, 00]
+block: write+readback ok at lba 2047
+framebuffer: 1280x800 @ 32 bpp, pitch=5120 bytes
+vibix online.
+interrupts enabled
+timer: TSC 2419 MHz
+userspace module ELF entry=0x400000 load_segments=2
+tasks: scheduler online
+userspace: loader mapping image (8728 bytes)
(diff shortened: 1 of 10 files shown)


dburkart:
Thanks for the review! Applied the two fixes in `7ca926c`:

- `syscall.rs`: propagate the concrete `ForkError` variant into the fork trace line.
- `docs/investigations/502-fork-trace.md`: markdown-lint nits (fenced-block language + `walk-through`).

Deferring the "bounded spin loop in `emit_raw_byte()`" nitpick as out-of-scope for this PR:

1. This is diagnostic-only code on the fork-child trampoline path — a ring-0 trampoline with an ad-hoc primed stack and no saved/restored scratch registers around a Rust call. Its whole reason to exist is to bypass the normal `serial_println!` path (which is `without_interrupts` + `COM1.lock()`).
2. If the UART is actually stuck (THR never goes empty), the correct behaviour for a diagnostic probe is to spin and block the boot so the failure is visible — silently dropping the canary byte would mask the exact regression this probe exists to detect.
3. An infinite spin here has never been observed on QEMU-TCG or QEMU-KVM in the 50-run reproducer sweep documented in `docs/investigations/502-fork-trace.md`.

No follow-up issue filed for this one — it's an explicitly optional nit and the trade-off points the other way in this specific context.

dburkart:
Closing as superseded. The same agent pivoted to a v2 branch (`m502-fork-trace-v2`) as PR #515, which merged at 16:44Z and closed #502. #514 and #515 cover the same scope; no content lost.

Branch `m502-fork-hang-trace` can be deleted when convenient.

Issue #501, opened 2026-04-17
Quality sprint — root-cause fork/exec/wait flakiness from PR #206


dburkart:
## Motivation

The user bisected a ~50% reproducible shell hang on boot of the vibix ISO to commit `ab4a61eb21f13cd16db67d02df430fb66023683d` (PR #206, *"feat: process table + fork/exec/wait lifecycle (#123)"*). The bug presents as: shell prompt (`vibix> `) prints, then every keystroke is dropped **and** the expected init markers `hello: hello from execed child` and `init: fork+exec+wait ok` never appear. The captured serial log (below) shows init kernel-prep, PML4 switch, ring-3 entry, and the userspace marker `init: hello from pid 1` all succeed — then total print-death and typing-death the instant init executes `syscall(57) = fork`.

Prior flakiness investigations (#303, #460, #420) circled this for weeks without closing it. #303 fixed the H1 COM1-lock race but explicitly deferred the H2 hypothesis on `FORK_USER_{RIP,RFLAGS,RSP}` statics — which PR #206 introduced and which remain live today. #420 is tagged *"re-harden fork+exec smoke markers once flakiness root-caused"* — its acceptance gate is exactly this epic's completion. This epic drives a focused sprint to root-cause the hang, replace the racy primitives PR #206 introduced, and soak-test the fix until the shell is deterministic.

**Leading hypothesis:** SYSCALL enters ring-0 with IF=0 (masked by `MSR_FMASK`). The fork syscall handler path — `fork_current_task` → `map_range` on a new kernel stack → `fork_address_space` (CoW clone) → `Task::new_forked` → `process::register` — takes at least one `spin::Mutex` whose holder requires interrupts to release, and the CPU spins forever with no preemption, no PS/2 IRQ, no timer. This matches the observed symptom exactly: total print-death + typing-death + fork as the last operation attempted.

**Secondary hypothesis (H2 from #303):** `FORK_USER_{RIP,RFLAGS,RSP}` `AtomicU64` globals get clobbered by a second syscall between the parent's fork-entry and the child's first schedule, so the child `sysret`s to a garbage RIP, faults into a #PF handler that itself hangs. Fits the same symptom shape.

## Work

This epic is a meta-issue. Concrete work is split into 8 sub-issues (filed immediately after this epic) and 7 existing backlog issues pulled into scope.

**New sub-issues** (children of this epic):
[...]

PR #520, opened 2026-04-17
kernel: replace FORK_USER_* globals with per-task SyscallReturnContext (#504)

### LINKED ISSUE #504: structural: replace FORK_USER_* globals with per-task saved state (epic #501)

dburkart:
## Summary

Closes #504 (epic #501). Removes the three `AtomicU64` statics
`FORK_USER_RIP / _RFLAGS / _RSP` plus the `SIGRETURN_PENDING` flag and
wires the FORK / SIGRETURN syscall arms to read and write the caller's
own `SyscallReturnContext` on its kernel stack instead.

Concretely:

- `syscall_dispatch` gains a leading `ctx: *mut SyscallReturnContext`
  argument (passed as `rdi` from the asm trampoline). The asm CALL
  remap is updated from 7→8 SysV args so `arg6 = a4` and `arg7 = a5`
  now land in two stack slots, preserving 16-byte alignment at the
  CALL site.
- `FORK` reads `(*ctx).user_{rip,rflags,rsp}` for `fork_current_task`.
- `SIGRETURN` reads `(*ctx).user_rsp`, decodes the `SigFrame`, and
  writes the restored `rip/rflags/rsp` straight back into `(*ctx)`.
  `check_and_deliver_signals` no longer needs the `SIGRETURN_PENDING`
  fast path and loses its old global-reload branch.
- A small pure helper `fork_abi::prime_fork_child_stack` now owns the
  stack-frame layout for `Task::new_forked`. It's dual-gated so host
  unit tests cover the slot math.
- A new `SYSCALL_SCRATCH_RSP` single-qword static replaces the former
  `FORK_USER_RSP` as the memory spill slot for user RSP during the
  two-instruction kernel-stack swap. It lives entirely inside the
  non-preemptible SYSCALL entry window; the doc comment spells out why
  this is structurally distinct from the removed cross-syscall
  globals.

Option A (per-task kernel stack) from the epic, because the state is
already there — no new per-`Task` field is needed.

## Test plan

- [x] `cargo test --package vibix --lib` — 377 passed including the
      new `fork_abi::tests::prime_fork_child_stack_*` tests (layout +
      two-children isolation, the regression anchor for the cross-CPU
      race).
- [x] `cargo xtask build`
- [x] `cargo xtask build --features fork-trace`
- [x] `cargo xtask smoke` — all 33 markers present, `init: fork+exec+wait ok`.
- [x] `cargo xtask test` — 234 integration-test passes, 0 failures.

Follow-ups tracked separately:

- Argument-register preservation across `SA_RESTART` + user handler
  (the restart-via-sigreturn path is unchanged by this PR).


dburkart:
Thanks @coderabbitai for the review.

- **fork_abi.rs doc comment (pops vs movs)**: Fixed in 241622e.
- **signal/mod.rs SA_RESTART+handler arg-reg preservation**: this is a pre-existing limitation, not introduced by this PR. The old code had the same gap (`sys_sigreturn` only restored rip/rflags/rsp before this PR too). The fix requires round-tripping rax/rdi/rsi/rdx/r10/r8/r9 through the `SigFrame` + `sys_sigreturn`. I noted this in the PR description as a follow-up and am keeping it out of scope here to keep the #504 global-removal surgical. I'll open a separate issue to track it.

PR #535, opened 2026-04-18
RFC 0004: Ext2 Filesystem Driver

dburkart:
## Summary

RFC 0004 proposes shipping a real, on-disk, writable, permission-aware
filesystem for vibix — an ext2 driver plus the missing POSIX write syscalls,
per-task credential enforcement, a block-device buffer cache, `mount(2)`/
`umount2(2)`, a boot path that mounts an ext2 image as root, and a
pjdfstest-based conformance suite.

will be reviewed by a 5-archetype panel (Security Researcher, OS Engineer,
User Space Staff Engineer, Academic, and Filesystem Engineer for on-disk
layout / fsck / crash-consistency review), revised until all blocking
findings are addressed, then merged. Implementation issues will be filed
from the Roadmap after merge, partitioned into seven dependency-ordered
workstreams (A: write syscalls · B: credentials · C: buffer cache · D: ext2
read · E: ext2 write · F: mount + root-fs · G: pjdfstest).

## RFC sections

- Abstract
- Motivation
- Background (prior art: Card 1994, Linux fs/ext2, SerenityOS Ext2FS, xv6,
  redoxfs, POSIX.1-2017, pjdfstest, Bach, Ganger, FSCQ)
- Design (on-disk types, feature-flag gate, buffer cache, credentials,
  mount, indirect-block walker, directory ops, allocator, orphan list,
  rename atomicity, boot path, syscall list)
- Security Considerations
- Performance Considerations
- Alternatives Considered
- Open Questions
- Implementation Roadmap (7 workstreams, dependency-ordered)

## Test plan

- [ ] All blocking findings are addressed within the scoped 2-cycle cap.
- [ ] RFC frontmatter updated to \`status: Accepted\` before merge.

### DIFF
--- a/docs/RFC/0004-ext2-filesystem-driver.md
+++ b/docs/RFC/0004-ext2-filesystem-driver.md
@@ -1,7 +1,7 @@
 ---
 rfc: 0004
 title: Ext2 Filesystem Driver
-status: In Review
+status: Accepted
 created: 2026-04-18
 ---
 
--- a/docs/RFC/0004-ext2-filesystem-driver.md
+++ b/docs/RFC/0004-ext2-filesystem-driver.md
@@ -191,10 +191,44 @@ target if ever needed.
   `bg_inode_table` are absolute block numbers.
 - `Ext2Inode` — 128 bytes in rev 0, `s_inode_size` bytes in rev 1.
   `i_block[15]`: 12 direct + 1 single-indirect + 1 double + 1 triple.
-  `i_blocks` is in **512-byte units**, not fs-block units — the most common
-  hobby-kernel bug.
+  `i_blocks` is in **512-byte units**, not fs-block units. Every block
+  allocated against the inode — data block, *and every indirect/
+  double-indirect/triple-indirect block* — adds `(block_size / 512)` to
+  `i_blocks`; every freed block subtracts the same. Truncate and file
+  extension paths must account for the freed/allocated indirect blocks, not
+  just the data blocks, or `e2fsck -fy` will rewrite `i_blocks` on every
+  touched image.
 - `Ext2DirEntry2` — variable-length `{ u32 inode, u16 rec_len, u8 name_len,
   u8 file_type, name[name_len] }`, 4-byte aligned, cannot span a block.
+- **On-disk inode writes are always read-modify-write of the raw slot** —
+  the full 128 bytes (or `s_inode_size` in rev 1) are read into memory,
+  parsed fields update in place, and the whole slot is written back.
+  Unknown/reserved fields (`i_generation`, `i_file_acl`, `i_faddr`, the
+  osd2 reserved bytes, `l_i_uid_high`/`l_i_gid_high`) are preserved
+  verbatim — dropping them would break NFS fh generation and `e2fsck -D`.
+- **`l_i_uid_high`/`l_i_gid_high` (osd2.linux2) carry bits 16..31 of
+  uid/gid.** The driver reads/writes both halves; `chown` of a value that
+  does not fit in u32 returns `EOVERFLOW` (u32 limit is well above POSIX
+  UID_MAX = 65535 in practice, but the check protects against future
+  enlargement).
+- **Nanosecond timestamps truncate.** `utimensat`'s `timespec` nanoseconds
+  are dropped at the disk layer; `stat` returns `tv_nsec = 0`. ext4-style
+  `*_extra` fields are out of scope.
+
+**Bitmap-to-absolute-block math (all allocator paths).** Bit `i` of the
+block bitmap for group `G` corresponds to absolute block number
+`s_first_data_block + G * s_blocks_per_group + i`, *not* to block `i`.
+On 1 KiB-block filesystems `s_first_data_block == 1` (the superblock
+occupies block 1); on ≥2 KiB-block filesystems `s_first_data_block == 0`.
+Mount asserts the relation. Forgetting the offset either leaks one block
+per group or double-allocates the superblock/BGDT. Bitmap bits past
+`s_blocks_count` (in the partial last group) are treated as **set**
+(unavailable); `mkfs.ext2` zero-fills them and the driver honors the
+count, not the bitmap width.
+
+Similarly, inode bitmap bit `i` of group `G` corresponds to inode number
+`G * s_inodes_per_group + i + 1` (inodes are 1-indexed; ino 0 is never
+valid). Bits past `s_inodes_count` are treated as set.
 
 Constants: `EXT2_ROOT_INO = 2`, `EXT2_GOOD_OLD_FIRST_INO = 11`,
 `EXT2_GOOD_OLD_INODE_SIZE = 128`, `EXT2_GOOD_OLD_REV = 0`,
@@ -233,7 +267,7 @@ pub struct Ext2Fs {
     sb: BlockingRwLock<Ext2SuperBlock>,
     bgdt: BlockingRwLock<Vec<Ext2GroupDesc>>,
     inode_cache: Mutex<BTreeMap<u32, Weak<Inode>>>,
-    orphan_list: Mutex<Vec<u32>>,
+    orphan_list: Mutex<BTreeMap<u32, Arc<Inode>>>, // strong ref — see below
     mount_flags: MountFlags,
 }
 impl FileSystem for Ext2Fs { ... }
@@ -244,10 +278,29 @@ pub struct Ext2Inode {
     ino: u32,
     meta: BlockingRwLock<Ext2InodeMeta>,  // parsed i_mode/uid/gid/size/etc.
     block_map: BlockingRwLock<Option<BlockMap>>, // lazy indirect-walk cache
+    unlinked: AtomicBool, // set when i_links_count reaches 0 while open
 }
 impl InodeOps for Ext2Inode { ... }
 impl FileOps for Ext2Inode { ... }
 ```
+
+**Orphan-list residency invariant.** `orphan_list` holds **strong**
+`Arc<Inode>` references for every unlinked-but-open inode on this mount —
+not inode numbers, not `Weak` refs. This is the only reference path that
+keeps the `Inode` resident after the last `OpenFile` closes between the
+`unlink` and a racing final-close on another CPU. Before dropping an
+`Inode` into `gc_queue`, `Inode::drop` in `kernel/src/fs/vfs/inode.rs:106`
+checks `unlinked.load()`; if set, eviction/`evict_inode` is vetoed (the
+orphan-list-final-close path owns the free). This simultaneously closes
+two hazards the previous draft left open: (a) data blocks being freed by
+`evict_inode` before the orphan-list pass walks, and (b) a re-`iget` of
+the same ino between last-close-on-CPU-A and orphan-walk-on-CPU-B
+observing a partially-freed inode.
+
+`inode_cache` remains `Weak<Inode>` — it is a lookup shortcut, not an
+ownership root. When the last `OpenFile` closes on an unlinked inode,
+the `orphan_list.remove(ino)` drop decrements strong count to zero and
+triggers the real free path.
 
 #### Buffer cache — `kernel/src/block/buffer_cache.rs`
 
@@ -279,11 +332,39 @@ impl BlockCache {
 }
 ```
 
-Replacement: **CLOCK-Pro** (Jiang, Chen, Zhang, USENIX ATC 2005). One
-reference bit per buffer, advanced on access, cleared on scan. Small caches
-(≤1,024 buffers) are most vulnerable to scan pollution, which plain LRU
-handles badly; CLOCK-Pro is scan-resistant at O(1) per access with a single
-byte of metadata per entry.
+Replacement: **CLOCK-Pro** (Jiang, Chen, Zhang, USENIX ATC 2005), chosen
+for its scan resistance at O(1) per access (Jiang et al. 2005). ARC
+(Megiddo & Modha, FAST 2003) is stronger but its 2-list metadata cost is
+unjustifiable at the cache sizes we target; LRU is cheap per-access but
+unbounded-degrades under `find /`-shape scans. At 4 MiB ≈ 1,024 buffers a
+single recursive-walk would otherwise thrash LRU. Metadata per entry:
+one reference byte + one state byte + the `Arc` strong-count.
+
+**Eviction invariants** (Workstream C must uphold all four, any one
+violation is a correctness bug — not a perf bug):
+
+1. **Never evict a pinned buffer.** The CLOCK-Pro hand skips any
+   `BufferHead` whose `Arc::strong_count > 1` (i.e., a caller holds a
+   live handle from `bread`). If an entire sweep finds no evictable
+   buffer, `bread` returns `ENOMEM` rather than blocking, and the caller
+   is expected to release its handles. This prevents the draft's
+   previous hazard of two distinct `Arc<BufferHead>`s coexisting for the
+   same `(dev, blk)` — which would silently desynchronize the dirty bit.
+2. **Never evict DIRTY + LOCKED_IO.** A buffer mid-`sync_dirty_buffer`
+   has both bits set; eviction must skip it. A buffer that is DIRTY but
+   not LOCKED_IO is written back *synchronously* by the writeback daemon
+   (not the evictor — see #3) and only then evicted.
+3. **`bread` never performs synchronous writeback during eviction.** If
+   the cache is full and every entry is either pinned or dirty,
+   `bread` returns `ENOMEM`; it must not call `sync_dirty_buffer` from
+   inside a path that a VFS caller is holding. The background writeback
+   daemon (or an explicit `sync_fs`) is the only legitimate origin of
+   dirty flushing. This closes the OS-engineer B5 hazard: no VFS lock
+   ever transitively holds the virtio-blk spin lock.
+4. **Single-cache-entry invariant.** An evicted buffer is removed from
+   `entries` before its `Arc` refcount can drop to 0 from the map side;
+   `bread` that finds no entry allocates, inserts, and pins *under*
+   `entries.write()` so no race creates a duplicate.
 
 Write-back policy: all metadata writes go through `mark_dirty`+sync (via
 `sync_dirty_buffer`) on atomicity-critical paths (bitmap allocate, dirent
@@ -292,9 +373,16 @@ list). Data writes are `mark_dirty`-only (delayed). `fsync(fd)` flushes
 all dirty buffers whose owner inode matches the fd's inode; `sync_fs`
 flushes all. A background writeback daemon flushes every 30 seconds.
 
-Each Ext2Fs owns its own `BlockCache` keyed by its own block size; this
-avoids block-size-mismatch hazards when multiple filesystems mount on the
-same block device in the future.
+**Writeback daemon lifecycle.** The daemon runs as a dedicated kernel
+thread per `BlockCache` (so one per Ext2Fs mount); it acquires an
+`SbActiveGuard` per sweep to keep the superblock resident and skips any
+mount whose `SuperBlock.draining == true`. `SuperOps::unmount` sets
+`draining`, synchronously flushes the cache, joins the daemon, and only
+then releases the block device.
+
+Each Ext2Fs owns its own `BlockCache` keyed by `(DeviceId, u64)`. Block
+size is carried on the `BlockCache` struct itself — not in the key —
+since each mount owns its cache at its own block size, avoiding mismatches.
 
 #### Credentials — `kernel/src/fs/vfs/mod.rs` (extended)
 
@@ -339,9 +427,20 @@ read `task.credentials.read().clone()` once at entry and pass the resulting
 4. Read BGDT: block 2 on 1 KiB filesystems, block 1 otherwise; `N = ceil(
    s_blocks_count / s_blocks_per_group)` descriptors.
 5. If `s_state != EXT2_VALID_FS` → log warning, force `MS_RDONLY`.
-6. In-memory: clear `s_state` → `EXT2_ERROR_FS`, mark superblock buffer
-   dirty (will write back on next `sync_fs` or `umount`).
-7. `iget(EXT2_ROOT_INO = 2)` → root inode; attach via
+6. If the mount is read-write (not forced-RO, not `MS_RDONLY` from caller,
+   not RO from an unknown `RO_COMPAT` bit): clear `s_state` →
+   `EXT2_ERROR_FS`, `sync_dirty_buffer` the superblock **before
+   `mount(2)` returns** (one extra 1 KiB write at mount time — cheap and
+   closes the OS-A4 hazard where a crash between mount and first flush
+   would leave on-disk `s_state == VALID_FS`, defeating the force-`e2fsck`
+   signal). **RO mounts skip step 6 entirely** — a read-only mount must
+   not touch the disk at all, to honor the contract that RO is
+   non-destructive even when the on-disk `s_state` was already stale.
+7. Walk the on-disk orphan chain from `s_last_orphan` before allowing
+   userspace access (see §unlink-while-open + orphan list for the
+   recovery protocol and chain-validation rules). RO mounts log orphan
+   entries but do not drain them.
+8. `iget(EXT2_ROOT_INO = 2)` → root inode; attach via
    `SuperBlock::set_root()`; register in the global mount table via the
    existing `mount_table.rs` API.
 
@@ -358,11 +457,35 @@ iblock in [12+P+P^2, 12+P+P^2+P^3)      -> i_block[14]  three-level
 
 Allocating-variant (`create=true`): walks the same path; on a zero pointer
 it allocates a new block via the bitmap allocator (§Allocator below) and
-writes the slot atomically (sync-dirty) before returning.
+writes the slot atomically (sync-dirty) before returning. Every allocated
+indirect block (single, double, triple) increments the inode's `i_blocks`
+by `(block_size / 512)` and counts against the parent group's free-block
+count — same accounting as a data block.
+
+**Every block pointer read from any slot — `i_block[0..12]`, a single-
+indirect slot, a double-indirect slot, a triple-indirect slot — is
+validated before it is handed to `bread`:**
+
+- Pointer `p = 0` → hole; read returns a zero block, write allocates.
+- `p < s_first_data_block` → `EIO`, force-RO the mount.
+- `p >= s_blocks_count` → `EIO`, force-RO the mount.
+- `p` aliases a metadata region — the superblock (block 0 or 1 depending
+  on `s_first_data_block`), the BGDT blocks, any `bg_block_bitmap`,
+  `bg_inode_bitmap`, or `bg_inode_table` run for any group — → `EIO`,
+  force-RO the mount. A cached "metadata-forbidden" bitmap is computed
+  once at mount and consulted on every pointer read; the cost is one
+  byte per block (or a range list) and closes the most dangerous
+  confused-deputy attack in the driver (crafted image aims a user-data
+  write at the BGDT or an inode table, rewriting `i_mode` to gain
+  privilege).
 
 Per-inode `BlockMap` caches resolved indirect-block numbers to avoid
-re-reading indirect blocks on sequential access; invalidated on any
-`setattr`/`truncate`/write-that-extends.
+re-reading indirect blocks on sequential access. **Invalidation uses a
+`u64` epoch stamp:** writers that change the map (setattr/truncate/
+extend) bump the epoch under the `block_map` write lock; readers record
+the epoch before dereferencing a cached entry and re-walk if the epoch
+differs post-read. This is strictly stronger than lock-drop-on-write and
+avoids the SMP stale-resolution hazard from OS-A5.
 
 #### Directory operations
 
@@ -370,19 +493,52 @@ Directories are regular files whose content is a packed stream of
 `Ext2DirEntry2` records padded to 4-byte alignment. Records never cross a
 block boundary.
 
-**Lookup:** iterate records, match by `name_len` + `name[]`. O(entries).
+**Per-record validation (on every read, before consumption).** The RFC's
+§Security bullet on `rec_len` is refined here and is normative for the
+iterator implementation:
 
-**Insert:** scan each block for an entry whose slack
-`rec_len - align4(8 + name_len) ≥ align4(8 + new_name_len)`. If found,
-split the entry (shorten its `rec_len`, append new record). If not,
-append a new full-block record by extending the directory file.
+- `rec_len >= 8` (record header size) and `rec_len` is 4-byte aligned.
+- `rec_len + cursor <= block_end` (no overrun into the next block).
+- `name_len + 8 <= rec_len` (name fits inside the record).
+- **For a live entry (`inode != 0`):** `name_len > 0`. A zero-length
+  name on a live record is an image corruption — return `EIO` and
+  force-RO.
+- **For any entry:** `file_type` (when `INCOMPAT_FILETYPE` is enabled)
+  must be one of the known values (`UNKNOWN/REG/DIR/CHRDEV/BLKDEV/FIFO/
+  SOCK/SYMLINK`); an out-of-range `file_type` on a live entry is `EIO`.
+- **`inode == 0` is the universal tombstone.** It can appear at any
+  slot, not just the first. All iterators (lookup, getdents64,
+  insertion slot-scan, rename walk) **unconditionally skip** records
+  with `inode == 0` — they never surface to userspace and never hand a
+  name back as live. The insertion path treats any zero-inode slot
+  whose `rec_len` accommodates the new entry as reusable.
+- **`inode != 0` but `inode < EXT2_GOOD_OLD_FIRST_INO` (and not
+  `EXT2_ROOT_INO = 2`)** — `EIO`. Reserved ino range 1..10 excluding 2
+  is never directly referenced by a live dirent; an image that does is
+  malicious.
+- **`inode > s_inodes_count`** — `EIO`.
 
-**Delete:** find the entry's predecessor in the same block; grow the
-predecessor's `rec_len` to swallow the deleted record. Special case:
-deleted entry is first in block → zero its inode field; scanners treat
-`inode == 0` in a first slot as a valid blank. This exact convention is
-what `e2fsck` expects (getting it wrong is the single most common
-hobby-kernel ext2 bug).
+**Lookup:** iterate records, skipping `inode == 0`; match by `name_len` +
+`name[]`. O(entries).
+
+**getdents64:** same iteration; `d_ino` is the dirent's `inode`, never 0
+(skipped); `d_type` mirrors `file_type` when `INCOMPAT_FILETYPE` is live,
+else `DT_UNKNOWN`.
+
+**Insert:** scan each block for a slot where either (a) the slot is a
+live entry (`inode != 0`) whose slack `rec_len - align4(8 + name_len) >=
+align4(8 + new_name_len)` lets the entry be split; or (b) the slot is a
+tombstone (`inode == 0`) whose `rec_len >= align4(8 + new_name_len)`, in
+which case the tombstone is overwritten in place. If neither fits in any
+existing block, append a new full-block record by extending the
+directory file.
+
+**Delete:** preferred path — find the entry's predecessor in the same
+block and grow the predecessor's `rec_len` to swallow the deleted
+record. When the deleted entry is first in the block (no predecessor),
+fall back to zeroing its `inode` field and leaving `rec_len` unchanged;
+the slot is now a tombstone reusable by a future insert. `e2fsck`
+tolerates both forms.
 
 #### Allocator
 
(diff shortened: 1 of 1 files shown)


dburkart:

**Summary:** The RFC is security-aware and enumerates the big-picture DAC surface, but several concrete kernel-corruption and privilege-escalation surfaces are glossed over — notably orphan-chain validation, fast-symlink detection, and the B-before-A sequencing claim. The "feature-flag gate + bounds-check" framing is necessary but not sufficient for a mountable untrusted on-disk format.

**Blocking findings:**
- [B1] Orphan chain not validated — §Design "unlink-while-open + orphan list" recovers by walking `s_last_orphan` with zero stated invariants on the chain. A malicious image can supply a cycle (infinite loop / livelock at mount), an ino `< EXT2_GOOD_OLD_FIRST_INO` (frees reserved inodes 1..10 — including root ino 2 — rendering the FS unbootable or, worse, re-usable by a later `iget`), or `ino > s_inodes_count` (out-of-bounds bitmap write, kernel memory corruption via BGDT-derived bitmap block number). Required mitigations that belong in §Security: (a) bound-check every ino against `[EXT2_GOOD_OLD_FIRST_INO, s_inodes_count]` before dereference, (b) cap chain length to `s_inodes_count` and detect revisits via a visited set, (c) refuse to free root/reserved inodes.
- [B2] Fast-symlink path confusion leaks `i_block[]` — Workstream D step 7 reads fast symlinks when `i_blocks == 0`, but the RFC never states the `i_mode` must be `S_IFLNK` or that `i_size ≤ 60`. A crafted image can set `i_blocks = 0` on a non-symlink inode (or set `i_size` up to `u32::MAX`), and a subsequent `readlink` will copy up to 60 bytes of the inline `i_block[15]` (or, if `i_size` isn't clamped, wildly out-of-bounds inode-table memory) to userland — disclosing on-disk metadata or adjacent inode-table contents. Fix: gate fast-symlink on `S_ISLNK(i_mode) && i_size <= 60 && i_blocks == 0` and clamp the copy to `min(i_size, 60)`.
- [B3] Indirect-block pointer not bounds-checked — §Design "Indirect-block walker" describes the math but never says each pointer read from a direct/indirect slot is validated against `[s_first_data_block, s_blocks_count)`. A malformed image with `i_block[12] = 0xFFFFFFFF` (or pointing at the superblock/BGDT/bitmap) will drive `bread(blk)` into arbitrary sectors — at minimum an info-leak of arbitrary on-disk bytes through `read(2)`, at worst a write-path aliasing attack where user data overwrites the BGDT or another inode table (full kernel-state corruption, trivial privesc via inode-mode rewrite). Add explicit "every block pointer is validated before `bread`, and must not alias metadata regions" to §Security Considerations bullet 3.
- [B4] Workstream sequencing does not actually close the root-cred window — §Security bullet 1 claims "wave ordering enforces B landing before A's *new* syscalls become reachable from ring-3." But §Overview places A and B in the same wave 1 with no declared merge ordering, and §Implementation Roadmap Workstream A has no gate on B. If Workstream A lands first (or partially), `mkdir`/`unlink`/`chmod`/`chown` are reachable from ring-3 with `Credential::kernel()` — a direct, unauthenticated privesc (any user becomes root-equivalent for every DAC check in the window). Make this a hard CI gate: A's syscall-wiring PRs must not merge until B's `Task::credentials` plumbing is in main, OR A's new syscalls must return `ENOSYS` until a runtime flag flips after B lands.
- [B5] `rec_len` validation omits name vs. record-tail overlap and the 4-byte-aligned `name_len` story — §Security bullet 3 lists `rec_len ≥ 8`, aligned, `rec_len + cursor ≤ block_end`, `name_len ≤ rec_len - 8`. Missing: `name_len > 0` for non-sentinel entries (a zero-length name can falsely match a `lookup("")` and also confuses `getdents64` reclen math), and validation that `file_type` (with `INCOMPAT_FILETYPE`) is in the known set — a bogus value is passed to userspace via `getdents64 d_type`, which is an API contract break but also signals a corrupt dirent that should probably `EIO` rather than be silently surfaced. More importantly, the `inode == 0` convention for a deleted-first-slot entry is explicitly allowed by the design — but the RFC does not say "skip such entries during `lookup`/`getdents64`." A crafted image can set `inode == 0` with a non-empty name that `lookup` then hands to `iget(0)`, which must be rejected (ino 0 is reserved and the inode table is 1-indexed).

**Advisory findings:**
- [A1] Errno differentiation across disk-error classes — §Security bullet 5 asserts errnos are pure enum discriminants, but the concern isn't kernel-address leakage, it's oracle leakage: distinguishing `EIO` (unreadable sector) from `EINVAL` (off-disk block number) from `EUCLEAN`/`ENOENT` lets an attacker probe whether a specific block number is valid on an arbitrary ext2 image. For MVP this is acceptable, but folding "off-disk pointer" and "read failure" both to `EIO` closes the channel cheaply.
- [A2] Rootfs-image trust boundary — §Security "Setuid exec" correctly keeps `MS_NOSUID` on non-root mounts but the rootfs `/bin/su`-style justification assumes the build-time image is trusted. Workstream F builds the image via `mkfs.ext2` in-build; if an attacker can influence the build (malicious dep, compromised CI), a suid-root binary in the image is a one-shot full kernel takeover. Suggest: document that the rootfs image must be reproducible and hashed, and that CI verifies the image hash before boot. Belongs as an explicit §Security bullet, not a build-system assumption.
- [A3] `setuid(2)` POSIX §2.4 ambiguity around supplementary groups — the RFC says `setuid` follows saved-set-uid rules but does not spell out whether `setuid(0) → setuid(nonzero)` clears supplementary groups (Linux does not; BSDs vary). `setgroups` is listed but the interlock with uid transitions is unspecified. An unprivileged user who was granted supplementary group `0` (wheel) via a previous privileged `setgroups` could retain group-0 DAC power after dropping uid. Pin the semantics in the RFC or the Workstream B task.
- [A4] SMP mid-epic risk — §Performance notes the per-Ext2Fs `BlockCache` and `inode_cache` are single-locked, "fine for UP." If SMP lands during this epic (soak/review/etc.), the `BlockingRwLock<Ext2SuperBlock>` + bitmap allocator path has a classic TOCTOU between "scan bitmap for free bit" and "set bit + sync." Two racing allocators can double-allocate the same inode/block, enabling a later attacker-controlled overlap of inodes / files. Recommend: an explicit RFC clause that this driver is UP-only until a follow-up RFC adds per-group locks + atomic bitmap CAS, and a compile-time `#[cfg(not(smp))]` assertion on `Ext2Fs` so SMP landing forces a re-review rather than silently breaking allocators.
- [A5] TOCTOU between `permission()` and the dirent op — §Security bullet 4 dismisses this only for rename's destination slot. The broader case (walk a path, `permission()` succeeds on a directory, another thread/process `chmod`s or `rename`s the directory before the op) is not addressed. For UP + single-locked inode this is narrow, but under SMP (see A4) or even cooperative preemption mid-syscall, the standard mitigation is "re-check permission after re-resolving on the still-pinned dentry" — worth stating explicitly so it isn't forgotten in Workstream E.
- [A6] `NGROUPS_MAX = 32` is low — Linux uses 65536 for a reason: supplementary-group-based ACLs are common. Not a security hole per se, but a too-small cap can mask a DAC denial that the user expected to succeed, leading to workarounds that widen access (e.g., chmod 777). Consider `NGROUPS_MAX = 65536` or at least 1024.
- [A7] Integer-overflow surface enumeration incomplete — §Security bullet 4 covers file-offset math but not: `rec_len` arithmetic during dirent-split insert (adding aligned `new_name_len` can overflow a `u16` if `rec_len` is near `block_size`), bitmap bit-index math (`byte * 8 + bit` where the chosen byte comes from a malformed `s_inodes_per_group`), and `s_blocks_count * block_size` for the on-disk size check. Spell these out.
- [A8] Buffer-cache eviction under DIRTY+LOCKED_IO — not strictly security but adjacent: the `BufferHead::state` atomic with `VALID | DIRTY | LOCKED_IO` needs a documented invariant that evictions never drop a DIRTY buffer that is mid-`sync_dirty_buffer`, or else crash recovery via orphan-list/valid-FS-flag can run against partial state. Cite the invariant in Workstream C tasks.

dburkart:

**Summary:** The RFC's kernel-internal design is coherent, but the userspace-facing surface has gaps that will block real C/Rust programs from running on the resulting fs. Several POSIX-mandated behaviors are unspecified, and one "deferral" (mmap MAP_SHARED → ENODEV) is load-bearing for dynamic linking.

**Blocking findings:**

- [B1] `mmap(MAP_SHARED)` on ext2 returns `ENODEV` — deal-breaker for userspace. glibc's `ld.so` maps the PLT/GOT pages of every shared library with `MAP_PRIVATE` *and* `MAP_SHARED` depending on segment; even `MAP_PRIVATE` for file-backed mappings is the universal path for ELF loading on Linux. The RFC says `read`/`write` copy-in-copy-out is fine, but no real userspace beyond a shell builtin will run without file-backed mmap. Either (a) ship `MAP_PRIVATE` file-backed mmap now (read-only, page-in-on-fault via the buffer cache — doable without a full page cache) and return `ENODEV` only for `MAP_SHARED`, or (b) explicitly restrict the epic's "done" criterion to static binaries and document that dynamic loading is deferred. Silently returning `ENODEV` for both flavors means `execve` of a dynamically linked binary from ext2 fails in confusing ways.

- [B2] `utimensat` semantics unspecified — POSIX-required flags missing. The RFC wires `utimensat` (280) but does not document `UTIME_NOW`, `UTIME_OMIT`, `AT_SYMLINK_NOFOLLOW`, or the permission rule. The permission rule is subtle and commonly gotten wrong: `(UTIME_NOW, UTIME_NOW)` (or `times==NULL`) requires write permission OR ownership OR `CAP_FOWNER`; explicit times require ownership (or `CAP_FOWNER`). Without this in the RFC, implementers will either over-restrict (breaking `touch file`) or under-restrict (allowing any writer to backdate mtime, a known anti-forensics vector). Please spell out both flag semantics and the per-case permission check.

- [B3] `chmod`/`chown` do not document POSIX-mandated SUID/SGID clearing. POSIX `_POSIX_CHOWN_RESTRICTED` (which Linux implements unconditionally) requires `chown`/`fchown`/`lchown`/`fchownat` to clear `S_ISUID` and (if group-exec is set) `S_ISGID` on success by a non-privileged caller. Similarly, POSIX `chmod` clears `S_ISGID` when the caller is not a member of the file's group. The RFC's `setattr`-for-chmod bullet does not mention this. Skipping this is a privilege-escalation footgun: a user chowns their own setuid binary to another user and it retains the SUID bit. This must be specified before implementation, because "I'll add the bit-clear later" is how CVEs happen.

- [B4] Per-syscall errno tables missing. The RFC says "Errno returns follow POSIX.1-2017 per the list in §Background" but §Background cites POSIX only generically. Userspace binds to *exact* errno values. At minimum the RFC must enumerate: `rename` — `EXDEV` (cross-device), `ENOTEMPTY` or `EEXIST` (non-empty dir target — POSIX allows either, pick one and commit), `EBUSY` (mountpoint), `EINVAL` (source is ancestor of dest); `unlink` — `EISDIR`, `EBUSY`, `EPERM` (sticky); `link` — `EXDEV`, `EMLINK`, `EPERM` (dir link); `mount` — `EBUSY`, `ENODEV`, `EPERM`, `EINVAL`; `setuid` family — `EPERM` (unprivileged transition outside {ruid, euid, suid}); `truncate` — `EFBIG`, `EISDIR`, `EINVAL`. Without this, pjdfstest will pass/fail on the specific errno constants and the implementation will ping-pong. Pick the values in the RFC, not in PR review.

**Advisory findings:**

- [A1] Syscall number cross-check — `renameat` (264) is missing from the table. RFC lists `rename` (82) and `renameat2` (316), but Workstream A also says "wire `rename`/`renameat` + `renameat2`". Add 264 to the numbered table. Otherwise the listed numbers I spot-checked (truncate=76, ftruncate=77, rename=82, mkdir=83, rmdir=84, link=86, unlink=87, symlink=88, readlink=89, chmod=90, fchmod=91, chown=92, fchown=93, lchown=94, getuid=102, getgid=104, setuid=105, setgid=106, geteuid=107, getegid=108, setreuid=113, setregid=114, getgroups=115, setgroups=116, setresuid=117, setresgid=119, mount=165, umount2=166, fchownat=260, mkdirat=258, unlinkat=263, linkat=265, symlinkat=266, readlinkat=267, fchmodat=268, faccessat=269, utimensat=280, renameat2=316, faccessat2=439) are all correct for Linux x86_64.

- [A2] `access(2)` vs `faccessat2(AT_EACCESS)` — real UID vs effective UID distinction not documented. `access(2)` (syscall 21, not in table — may be assumed implicit via `faccessat(AT_FDCWD, ...)`) uses the *real* UID, `faccessat2(AT_EACCESS)` uses the *effective* UID. This is a common subtle bug (setuid programs probing permissions on behalf of the invoking user rely on `access` using ruid). Please spell this out in the setattr/permission section.

- [A3] `rename` self-rename — wording "no-op" is incomplete. POSIX: "If the old argument and the new argument resolve to *either the same existing directory entry or different directory entries for the same existing file*, rename() shall return successfully and perform no other action." The "different directory entries for same inode" case (two hard links, or path-with-symlinks-that-resolve-same) is the important one — please confirm the check is by resolved-inode, not by string-equality of paths.

- [A4] `RENAME_EXCHANGE` deferred — acceptable but call out the workaround gap. Real-world use case: atomic config-file replacement with rollback (both old and new exist on disk during edit; atomic swap replaces). Without `RENAME_EXCHANGE`, userspace falls back to `rename(new, old)` which loses the original atomically. This is a userspace-visible limitation; document it in the RFC so editor/systemd-like daemons know to expect non-atomic config swap.

- [A5] `setreuid`/`setresuid` unprivileged semantics terse. The RFC gives the privileged case exhaustively and says "similarly for the rest" for unprivileged — please spell out: unprivileged `setreuid(r, e)` requires `r ∈ {-1, old ruid, old euid}` and `e ∈ {-1, old ruid, old euid, old suid}`; `suid` updates only if `ruid` changed or `euid` changed to a value ≠ old ruid. These rules are wrong in at least one hobby kernel I've audited; spell them out verbatim.

- [A6] `statfs`/`statvfs` out-of-scope will bite. `df`, glibc's `pathconf(_PC_NAME_MAX)` / `fpathconf`, and any program that checks "is there enough space before starting a large write" need `statfs`. Even a stub that fills in `f_bsize`, `f_blocks`, `f_bfree`, `f_namemax=255` is ~50 LOC and unblocks a lot of userspace. Consider adding a minimal stub in this epic rather than deferring.

- [A7] `readlink`/`readlinkat` NUL-termination contract — POSIX says readlink does *not* NUL-terminate. Many hobby kernels get this wrong. Please specify in the RFC.

- [A8] `rename` cross-dir crash-window "temporary hardlink" — acceptable since Linux ext2 is identical, but add a pointer in `docs/` that `e2fsck` is required post-crash, and that `/sbin/init` should not assume crash-consistency on the rootfs. Document it in the user-facing release notes, not just the RFC.

- [A9] `access` (bare syscall 21) not in table. Either add it or say "`access` is provided via `faccessat(AT_FDCWD, path, mode, 0)`" so glibc knows to translate.

- [A10] `mount(2)` superuser-only — fine for MVP but `Alternatives Considered` mentions user-namespace mount. Please add an `Open Question` explicitly reserving design space for `CLONE_NEWNS`-style unprivileged mount so we don't paint into a corner at the syscall level (e.g., if `mount(2)` starts stashing caller-must-be-euid-0 deep in the dispatch, refactoring later is painful).

- [A11] Unlink-while-open + `s_last_orphan` recovery — the RFC's on-mount recovery story is not explicit. `e2fsck` handles the chain on a dirty mount, but vibix's mount path says it force-ROs on `s_state != EXT2_VALID_FS`. That means a crash with orphaned inodes leaves the fs read-only until a host runs `e2fsck`. For a dev kernel that's fine; for a rootfs that wants to self-heal, consider adding a minimal in-kernel orphan-chain drain at mount time (before flipping to RW), so a clean reboot after `kill -9` on a process with deleted-but-open files doesn't wedge into RO. At minimum, document the recovery protocol.

- [A12] `NGROUPS_MAX = 32` — Linux is 65536 since 2.6.4. 32 is very tight; `id` output on any distro-adjacent user will truncate. Consider `NGROUPS_MAX = 65536` or at least 256; the cost is a `Vec<u32>` per `Credential`.

dburkart:
### Review: OS Engineer

**Summary:** The data-flow and on-disk choices are sound and match the Linux/SerenityOS precedent cited, but the concurrency contract is under-specified in several places where the existing VFS already has hard-won invariants. As written the proposal will deadlock or lose orphan inodes; none of the problems are fundamental to ext2, all are fixable in the RFC.

**Blocking findings:**

- [B1] Blocking-lock over spin-lock inversion in `sync_dirty_buffer` — §Design/Buffer cache and §Algorithms/unlink promise `sync_dirty_buffer` on bitmap-alloc, dirent insert/delete, rename dst and orphan-list writes. The only block device today (`kernel/src/block/virtio_blk.rs:220`) is a `spin::Mutex<Option<BlkDevice>>` around a *polled* transport — `submit_and_wait` spins on the used-ring with preemption effectively pinned. Path-walk, `create`, `unlink`, `rename` all hold `Inode.dir_rwsem` (a `BlockingRwLock`) and frequently `sb.rename_mutex`; calling `sync_dirty_buffer` under those locks means a blocking-sleepable lock holder takes a non-sleepable spin lock and then busy-polls device I/O. On any future SMP or interrupt-driven completion that becomes a scheduler priority inversion and a deadlock risk if a holder of the virtio spin lock ever contends for a VFS lock (none today, but the RFC doesn't name the invariant). Fix: specify the lock class explicitly — the buffer cache and the block-device surface must be expressible as "no spin lock is held across a block I/O wait" — and state that VFS locks strictly dominate any block-layer lock.

- [B2] `inode_cache: Mutex<BTreeMap<u32, Weak<Inode>>>` cannot keep orphans resident — §Design/In-memory types uses `Weak<Inode>` throughout, and §Algorithms/unlink says "in-memory openers > 0 → add to `orphan_list: Mutex<Vec<u32>>`." But `Vec<u32>` is inode numbers, not `Arc<Inode>`, and there's no other strong reference path when the last `OpenFile` closes between `unlink` and final cleanup (e.g., `fd` closed by exit-group on one CPU while another CPU is mid-`rename` holding target-ino in the orphan list). `Inode::drop` in `kernel/src/fs/vfs/inode.rs:106` also enqueues into `gc_queue`, which will run `evict_inode` and free data blocks *before* the orphan-list pass — the POSIX "unlinked-but-open" invariant is silently violated. Fix: orphan list must hold `Arc<Inode>`, or `InodeState.unlinked` must veto `gc_queue` eviction.

- [B3] Rename lock order contradicts the existing VFS contract — `kernel/src/fs/vfs/ramfs.rs:340` and `super_block.rs:58` document the canonical order: (1) `sb.rename_mutex`, then (2) two `Inode.dir_rwsem` write locks taken in **ino order**. RFC 0004 §Design/Key data structures proposes per-Ext2Fs locks (superblock rwlock, BGDT rwlock, block cache rwlock, inode_cache mutex, per-inode meta rwlock, per-inode block_map rwlock) and never specifies how rename interleaves them with `sb.rename_mutex` and `dir_rwsem`. Under SMP two crossed-rename callers that each allocate a dirent block under the block-cache lock while holding opposite `dir_rwsem`s will deadlock. Fix: RFC must publish the total lock order (at minimum: `sb.rename_mutex` > `dir_rwsem[min(ino)]` > `dir_rwsem[max(ino)]` > `Ext2Inode.meta` > `Ext2Inode.block_map` > `Ext2Fs.bgdt` > `BlockCache.entries` > per-`BufferHead.data`) and assert it matches the `rename_mutex`/`dir_rwsem` invariant ramfs already upholds.

- [B4] `BufferHead` eviction race vs. live `Arc<BufferHead>` — §Design/Buffer cache says `release(bh)` may evict and Workstream C bullet 4 says "sync-on-eviction when cache is full and no clean candidate." A caller holding `Arc<BufferHead>` returned from `bread` has its own strong ref, but the `entries: RwLock<HashMap<..., Arc<BufferHead>>>` map also holds one; the CLOCK-Pro hand may pick a buffer whose refcount > 1 and either (a) block waiting for it (no API to signal completion) or (b) evict it from the map while callers still hold `Arc`s, creating two cache entries for the same `(dev, blk)` on the next `bread` and losing write-coalescing/dirty state. Fix: specify that eviction skips any `BufferHead` with `Arc::strong_count > 1` *or* that `bread` returns a pinned guard that holds the map entry. Today's draft lets the same `(dev, blk)` exist as two distinct `Arc<BufferHead>`s in the cache simultaneously — catastrophic for the dirty-bit invariant.

- [B5] Sync-on-eviction inside `read()` pulls block-device lock under VFS lock — §Perf and Workstream C explicitly permit "sync-on-eviction when full and no clean candidate." A normal `read()` syscall holds the inode's I/O path; if it calls `bread` and the cache needs to evict a dirty buffer, the eviction path calls `sync_dirty_buffer` → virtio-blk `spin::Mutex` while the caller transitively holds VFS locks above. This is the same hazard as B1 but reached from the read path, not just the metadata path the RFC flags. Fix: specify that `bread` never synchronously writes; dirty eviction defers to the writeback daemon or returns `ENOSPC` to the caller, or the read path must drop VFS locks across the flush.

- [B6] Allocation/free symmetry on partial-failure paths unspecified — §Algorithms/Allocator and Workstream E list five allocating paths (inode bitmap, block bitmap, indirect block, dirent slot, orphan list) but never say what happens when `create` allocates an inode via the inode bitmap, `sync_dirty_buffer`s it, then fails to insert the dirent (e.g., parent directory extend hits `ENOSPC`). On-disk the inode bit is set, in memory no `Arc<Inode>` references it — the inode leaks until next `fsck`. Same hazard for `rename` cross-dir between steps 1 and 2 if step 2 fails for reasons other than a crash (I/O error, not just power loss). Fix: specify rollback ordering (write dirent first, allocate inode second, or journal-less "undo log" in the buffer cache) or explicitly declare the leak semantics and require the valid-FS flag to be cleared already so `e2fsck` is forced.

**Advisory findings:**

- [A1] `BlockDevice` trait does not yet exist — `Arc<dyn BlockDevice>` appears throughout §Design but there is no such trait in `kernel/src/block/` today; `virtio_blk::{read, write}` are free functions keyed on a global `DEVICE`. Workstream C should explicitly list "introduce `pub trait BlockDevice { fn read_at; fn write_at; fn block_size; fn capacity; }`" and migrate virtio-blk behind it; otherwise Workstream D has an undeclared dependency.

- [A2] `Credential` field-order is a soft ABI break — §Design/Credentials extends the struct from `{uid, gid, groups}` to `{uid, euid, suid, gid, egid, sgid, groups}` with `pub` fields. Today only `Credential::kernel()` is used outside the module (grep across `kernel/src` + `kernel/tests` confirms no field-literal construction), so this is a soft break, but the RFC should (a) note the field-literal pattern is now discouraged and (b) add a constructor `Credential::from_task_ids(...)` so future call sites don't hard-code field order.

- [A3] Writeback daemon lifecycle vs. `umount` — §Design/Buffer cache says "background writeback daemon flushes every 30 seconds" but doesn't specify which task it runs on, its priority, or how it coordinates with Phase B of unmount (`SuperOps::unmount` in `kernel/src/fs/vfs/ops.rs:184`). If the daemon is iterating dirty buffers for an `Ext2Fs` whose `SuperBlock.draining == true`, what happens? Specify: daemon takes `SbActiveGuard` per pass and skips draining superblocks, and `SuperOps::unmount` must synchronously flush then tear down the daemon's handle to this mount.

- [A4] In-memory `s_state = ERROR_FS` clear is racy under the current "lazy writeback" wording — §Algorithms/Mount step 6 says "clear `s_state` → `ERROR_FS`, mark dirty (will write back on next `sync_fs` or `umount`)." A crash between mount and the first background flush leaves on-disk `s_state` still `VALID_FS`, so a subsequent boot will *not* force `e2fsck`, defeating the whole point of the flag. Fix: step 6 must `sync_dirty_buffer` the superblock before `mount(2)` returns. Cheap — it's a single 1 KiB write at mount time.

- [A5] Per-inode `block_map` cache invalidation is under-specified for SMP — §Algorithms/Indirect-block walker says the cache is "invalidated on any `setattr`/`truncate`/write-that-extends." Under `BlockingRwLock`, a concurrent reader can hold a reference to a cached resolution while a writer clears the map and re-walks; if the reader then reads the stale resolved block number, it can read a block that was just freed and reallocated to another inode. Fix: invalidation must either (a) take the `block_map` write lock, drop it only after the caller re-reads, or (b) use an epoch/version stamp the reader checks post-read. The RFC should state which.

- [A6] `s_last_orphan` chain and in-memory `orphan_list` are two sources of truth — §Algorithms/unlink updates both. If a buffer-cache eviction flushes the superblock but not the inode's `i_dtime`, the on-disk chain head points at an inode whose `i_dtime=0` — `e2fsck` will consider it live. Fix: either batch both writes via sync-dirty, or document that the orphan-list recovery path in `e2fsck` tolerates this (it does, but the RFC should cite it).

- [A7] Buffer-cache keyed `(DeviceId, u64)` — §Design/Buffer cache says `entries: RwLock<HashMap<(DeviceId, u64), Arc<BufferHead>>>` and §Workstream C says keyed on `(device, block_no, block_size)`. Inconsistent — pick one. Since each Ext2Fs owns its own `BlockCache` at its own block size, the block size can be implicit in the `BlockCache` instance (the struct already carries `block_size`).

- [A8] `NGROUPS_MAX = 32` is very small — Linux defaults to 65,536. 32 is fine for MVP but will surprise POSIX tests that set >32 supplementary groups; call this out in §Open Questions.

dburkart:
### Review: Filesystem Engineer

**Summary:** The on-disk structure walk-throughs and feature-flag posture are broadly correct, but the RFC leaves several crash-consistency and on-disk-bookkeeping contracts underspecified in ways that will cause `e2fsck -fy` to complain (or, worse, silently corrupt images) unless pinned down before implementation.

**Blocking findings:**

- [B1] `s_first_data_block` unhandled in bitmap math — The RFC correctly notes the 1 KiB vs ≥2 KiB block-size split for where the superblock and BGDT live, but nowhere does it say that block-bitmap bit *0* of block-group 0 corresponds to absolute block `s_first_data_block` (1 on 1 KiB, 0 on ≥2 KiB), not to block 0. Every `bg_block_bitmap` scan must offset by `s_first_data_block + G * s_blocks_per_group`. Forgetting this is an off-by-one that either leaks one block per group or double-allocates the superblock/BGDT. Specify the bitmap-to-absolute-block formula explicitly and add a mount-time assertion that `s_first_data_block` matches the block-size invariant.

- [B2] `i_dtime` double-use (orphan next-pointer vs deletion-time) is contradictory as written — §unlink-while-open says step 4 on final close sets `i_dtime` to the deletion time, but §orphan list says `i_dtime` is repurposed as the chain-next pointer while the inode is on `s_last_orphan`. These cannot both be live simultaneously. The correct sequence (Linux `ext2/ialloc.c`): while on the orphan list, `i_dtime` holds the next-ino link and the inode is still referenced via `s_last_orphan`; only at the moment the inode is unlinked *from* the orphan list and freed to the inode bitmap does `i_dtime` get rewritten to the deletion wall-clock and the inode bitmap bit cleared. If the driver writes a wall-clock `i_dtime` before removing the inode from the chain, `e2fsck` follows a corrupt chain into a random ino and either frees a live inode or loops. Spell out the ordering.

- [B3] Orphan-list ordering rule missing — adjacent to B2: the RFC gives no ordering between "bitmap-free the inode" and "unlink from orphan chain" and "write `i_dtime`". A crash mid-sequence must leave the chain walkable. Required ordering is: (1) unlink ino from the `s_last_orphan` chain on disk and sync; (2) free data blocks + free inode bitmap bit; (3) set `i_dtime` (wall-clock) and `i_links_count = 0` in the now-free inode slot. Reversing (1) and (2) can produce an orphan-chain that references an already-free inode slot, which `e2fsck` flags as "Inode N, i_blocks is X, should be 0" or worse.

- [B4] `i_blocks` update on write/truncate is unspecified — the RFC correctly calls out the 512-byte-unit gotcha for reading, but §Indirect-block walker does not say who increments `i_blocks` on allocation, by how much, or whether indirect-block blocks themselves are counted (they are: every allocated indirect/double-indirect/triple-indirect block contributes `(block_size / 512)` to `i_blocks`, same as a data block). Missing this makes `e2fsck` print "Inode N, i_blocks is X, should be Y" on every image the driver touches. Specify the increment/decrement rule precisely in both the allocating walker and the truncate path.

- [B5] Dirent-iteration contract for `inode == 0` entries is ambiguous — §Directory operations says deleted-first-of-block entries zero the inode field and "scanners treat `inode == 0` in a first slot as a valid blank." But the lookup/readdir text doesn't specify *skipping* `inode == 0` entries unconditionally (not just in first slot); and `getdents64` must not return them. Additionally, the insert path must treat any `inode == 0` slot as reusable, not only first-of-block. Pin down: (a) all iterators skip `inode == 0`; (b) insert reuses any zero-inode slot whose slack fits; (c) the "zero the inode" tombstone convention is used *only* when the record is the first in the block — middle/last deletions merge into the predecessor.

- [B6] Cross-directory rename write ordering inverts the required safe order — §rename step 1 says "write new dirent in dst_dir" then step 2 "delete old dirent in src_dir". This is the right order, but it is missing a prior step that must come first: bump `i_links_count` on the target inode *before* writing the new dirent (because after step 1 and before step 2 the inode has two dentries, so its link count must reflect 2, else `e2fsck` will see one orphan-looking entry and decrement or remove it). Conversely on the dst-overwrite case, the victim's link-count decrement must follow the new-dirent write. Specify: `i_links_count++` → sync inode → write new dirent → sync dst block → remove old dirent → sync src block → `i_links_count--` → sync inode. Current two-step sketch will leave images that `e2fsck -fy` "fixes" by mutating link counts.

- [B7] Write-ordering for create is not sequenced — on create the soft-updates rule is: (a) allocate+write inode bitmap bit, (b) write the inode with valid `i_links_count ≥ 1`, (c) write the dirent pointing at it — each with `sync_dirty_buffer` before the next. The RFC says "allocator + dirent use `sync_dirty_buffer`" but never states the dirent-after-inode ordering. If the dirent lands before the inode is populated, a crash leaves a dangling dirent referring to a free or garbage inode — `e2fsck` will remove the dirent (best case) or corrupt it (worst). Make the ordering normative in §Algorithms.

**Advisory findings:**

- [A1] `l_i_uid_high` / `l_i_gid_high` in `i_osd2` — Linux ext2 stores the upper 16 bits of uid/gid in `osd2.linux2.l_i_uid_high`/`l_i_gid_high`. The RFC is silent. Document either (a) full 32-bit uid/gid support reading+writing both halves, or (b) an explicit restriction to uid/gid < 65536 with `EOVERFLOW` on `chown` above that. Silently truncating ownership is an on-disk corruption that survives reboots.

- [A2] Nanosecond timestamps truncated — `utimensat` accepts `timespec` nanoseconds; ext2 stores only `__le32 i_{a,c,m}time` seconds. Specify that nanos are truncated on write and reads return `tv_nsec = 0`. (ext4's `*_extra` fields are out of scope, which is fine; just say it.)

- [A3] Preserve unknown/reserved inode fields — `i_generation`, `i_file_acl`, `i_faddr`, reserved bytes in `i_osd2`. The driver must read-modify-write the full 128 (or `s_inode_size`) bytes of the slot, not reconstruct from a parsed struct that drops unknown fields. Otherwise `e2fsck -D` and NFS (via stale `i_generation`) will flag the image. Add to the on-disk types section: "inode writes are always RMW of the raw bytes; parsed fields update in place."

- [A4] Sparse superblock backups — RFC lists `RO_COMPAT_SPARSE_SUPER` as supported but doesn't say whether the driver updates all backup superblocks on every write. Linux policy is: only primary on mount/umount/routine; backups only on resize/tune2fs. Document that explicitly; otherwise a reader may infer per-write backup propagation (expensive, incorrect).

- [A5] Fast-symlink boundary — RFC says "< 60 chars stored inline." Linux stores up to and including 60 bytes inline (the full `i_block[15]` array, 15·4 bytes, no NUL required). The check is `i_size <= 60 && i_blocks == 0`. Fix the comparator; require both conditions (not just `i_blocks == 0`) to distinguish a zero-length regular file from a zero-byte symlink body.

- [A6] `mkfs.ext2` default feature set on modern e2fsprogs — confirm tested against `mkfs.ext2 -t ext2` on the host the xtask will run on: Ubuntu 22.04+ / Fedora enable `ext_attr` (COMPAT — ignored, fine) and `resize_inode` (COMPAT — fine), but some vendored e2fsprogs also set `dir_index` (COMPAT for the bit, but HTree dirs exist in practice when files are added). Verify the xtask `mkfs.ext2 -t ext2 -b 4096 -O ^dir_index,^has_journal,^ext_attr` (or equivalent explicit feature mask) so the rootfs image never contains features the driver force-ROs.

- [A7] Valid-FS flag flow for force-RO mounts — §Mount step 6 says clear `s_state` to `EXT2_ERROR_FS` on mount. This should be skipped when mounting RO (including force-RO from unknown RO_COMPAT bits or stale `s_state`); otherwise a mount that was supposed to be non-destructive writes the superblock anyway and you've touched a disk you promised not to.

- [A8] BGDT bounds on partial last group — ensure allocator and bitmap readers honor that the last block-group may hold fewer than `s_blocks_per_group` blocks (and fewer than `s_inodes_per_group` inodes). Bitmap bits past `s_blocks_count`/`s_inodes_count` must be treated as set (not free). RFC doesn't mention this edge case; `mkfs.ext2` zero-fills those bits and expects the driver to respect the count.

dburkart:
### Review: Academic

**Summary:** The RFC engages seriously with the canonical literature (Card 1994, McKusick 1984, Bach 1986, Ganger 2000, Chen 2015, Amani 2016) and the design itself is sound. However, several citations are misattributed or over-claim what the cited work actually says, and two published results that bear directly on the design's crash-consistency and cache-replacement claims are missing. None of these findings rise to contradicting a safety or liveness property; all are advisory.

**Blocking findings:**
- (none)

**Advisory findings:**
- [A1] **Card 1994 §3.2 "quadratic hash" misattribution** — §Background and §Design both cite "Card §3.2 policy" for the inode allocator's "parent-group-first, quadratic-hashes to a different group on failure" behavior. The Card/Tweedie/Ts'o 1994 paper ("Design and Implementation of the Second Extended Filesystem," Linux Expo 1994) does not contain a §3.2 on quadratic hashing and does not describe a quadratic-probe spill policy; it describes only the "try the parent's group first" locality heuristic. The quadratic-hash spill is a Linux `fs/ext2/ialloc.c` implementation detail introduced later (see Ts'o, `find_group_dir`/`find_group_other`). Recommend: cite the Linux source (`fs/ext2/ialloc.c`) for the quadratic-hash spill and cite Card 1994 only for the parent-group-locality rule. The RFC's actual implemented policy ("first-fit in parent's group, linear spill") is in fact simpler than both, which is fine — but the citation chain is currently wrong.
- [A2] **CLOCK-Pro "small caches most vulnerable to scan pollution" is not in the paper** — Jiang, Chen, Zhang, "CLOCK-Pro: An Effective Improvement of the CLOCK Replacement" (USENIX ATC 2005) argues scan resistance and workload-adaptive behavior **across cache sizes**; it is not specifically an argument about caches ≤1,024 buffers. The paper's evaluation uses a wide range of sizes, and its core claim is that CLOCK-Pro inherits LIRS's scan resistance at CLOCK's O(1) cost regardless of cache size. The RFC's framing ("Small caches (≤1,024 buffers) are most vulnerable to scan pollution, which plain LRU handles badly") is a reasonable intuition but is not what the cited paper proves. Recommend: soften to "CLOCK-Pro is scan-resistant at O(1) per access (Jiang et al. 2005); at 4 MiB / 1,024 buffers a single `find /` would otherwise thrash LRU." Consider also citing Megiddo & Modha (ARC, FAST 2003) for the counter-argument — ARC's 2-list metadata cost is the actual reason to prefer CLOCK-Pro for a small OS-kernel buffer cache, not because "small caches are most vulnerable."
- [A3] **Cross-directory rename crash window: "identical to Linux ext2" is unverified** — §Design claims the cross-dir rename temporary-hardlink window is "identical to Linux ext2's behavior (journal-free) and is documented." This is asserted, not cited. Pillai et al., "All File Systems Are Not Created Equal," OSDI 2014, is the direct empirical audit of per-filesystem persistence properties (their BOB tool and ALICE framework specifically measured ext2/ext3/ext4/btrfs crash behaviors) and should be cited here. More pointedly: Pillai et al. found that "safe" atomicity assumptions about Linux ext2 rename are often violated because the orderings of buffer-cache writeback and on-disk block layout leak through. Recommend: either (a) cite Pillai 2014 and state which of their ext2 orderings the RFC's `sync_dirty_buffer` pairing preserves, or (b) weaken the claim to "matches the intended ext2 on-disk invariant, modulo writeback ordering, which `e2fsck` reconciles." Also relevant: Chidambaram et al., "Optimistic Crash Consistency," SOSP 2013, for the optimistic-fsync pattern, and Ganger & Patt, "Soft Updates" (TOCS 18(2), 2000) for the canonical three metadata-ordering rules — the RFC cites Ganger but does not name the three rules or state which ones the `sync_dirty_buffer` policy actually satisfies.
- [A4] **Fast symlink "< 60 bytes" is off-by-one** — §Design does not give an explicit bound, but Workstream D says "fast symlink (`i_blocks == 0`, ≤60 chars, stored inline)" while the Abstract / Alternatives paragraphs' spirit (and the RFC task text given to reviewers) suggest "< 60". `i_block` is 15 × u32 = 60 bytes exactly; Linux's `ext2_symlink` uses `<= 60` (strict ≤, including the case where the entire 60 bytes are name bytes with no NUL terminator, because ext2 symlinks are not NUL-terminated on disk). Recommend: explicit "≤ 60 bytes (`i_block` array is exactly 60 bytes wide; names are not NUL-terminated on disk)" in §Design and a matching pjdfstest-level unit test for the name_len == 60 edge case. Getting this off by one is a silent data-corruption bug, not just a cosmetic spec issue.
- [A5] **FSCQ as "semantic target" is an overclaim** — Chen et al., "Using Crash Hoare Logic for Certifying the FSCQ File System" (SOSP 2015) specifies a **write-ahead-log-backed** filesystem (log-bitmap-inode-dir) and proves a subset of POSIX (the paper's own evaluation lists the verified syscalls explicitly — no `rename`, no `fsync`-durability spec, no orphan list, no setuid-exec semantics). ext2's semantics diverge materially: bitmap-allocator writeback races, orphan list as a first-class on-disk data structure, cross-dir rename without a log. Recommend: either (a) demote the FSCQ reference to "FSCQ demonstrates machine-checked proofs are tractable for a POSIX subset" and drop "semantic target" framing, or (b) explicitly enumerate which FSCQ POSIX clauses the vibix ext2 driver intends to match and which are knowingly out of scope (probably `rename`, `fsync` semantics, and the orphan list). BilbyFs (Amani et al., ASPLOS 2016, CoGENT) is a closer bitmap-allocator analog and deserves higher billing than FSCQ as the "semantic target."
- [A6] **Orlov "~20–30% fragmentation reduction per LWN/folklore"** — the Alternatives section acknowledges this is folklore, which is honest. The original LWN article (https://lwn.net/Articles/14633/) actually reports "time required to traverse a Linux kernel tree was reduced by 30% or so" — this is **access time reduction on an aged tree**, not a direct fragmentation-percentage measurement, and the metric is walk-time not fragmentation. If the RFC keeps the claim at all, recommend quoting LWN verbatim: "traversal time on an aged tree reduced ~30% (LWN 2002)" rather than "fragmentation reduction ~20–30%."
- [A7] **pjdfstest "237 .t files · ~8,600 tests" is close but not the current number** — the current upstream `pjd/pjdfstest` master counts ~235 `.t` files and ~8,800 test cases on ext4 (JuiceFS and OpenBenchmarking's recent audits report 8,813). The 237 / 8,600 figure in the RFC likely comes from the research brief and was accurate at some past pin. Recommend: pin a specific commit of `pjd/pjdfstest` (or `saidsay-so/pjdfstest`) in Workstream G and derive the count from that pin rather than quoting folklore numbers. There is no academic audit of pjdfstest's coverage of POSIX itself — that would be a nice follow-up paper, but is not a reason to block this RFC.
- [A8] **POSIX.1-2017 §4.4.2 "first-match terminates" framing is precise but deserves a sharper cite** — POSIX.1-2017 §4.4.2 ("File Access Permissions") does indeed specify that the file permission bits are checked in owner → group(s) → other order and that earlier matches are terminal (so an owner whose UID matches but who is denied by owner bits is not rescued by looser group bits). The RFC states this correctly. One nuance to pin down in the design: "additional file access control mechanisms" (POSIX §4.4 general) — vibix has no ACLs, so nothing to worry about there, but it would be worth an explicit "no ACLs / no POSIX.1e `acl_*`" sentence under Alternatives so reviewers don't expect `getfacl`/`setfacl` to be covered by the "permission-aware" claim.
- [A9] **Formal-methods survey is light on the middle ground** — the RFC cites FSCQ and BilbyFs but omits Sigurbjarnarson et al., "Push-Button Verification of File Systems via Crash Refinement" (Yggdrasil, OSDI 2016), which is arguably the closest published verification work to a bitmap-allocator filesystem and which specifically targets crash refinement rather than full functional correctness. If the RFC is going to position itself against the verification literature at all, Yggdrasil belongs alongside FSCQ/BilbyFs.

**Verdict:** LGTM

dburkart:
## Defense cycle 1 — blocking findings addressed

All 22 blocking findings across the four CHANGES REQUESTED reviews are addressed in `b03af7e`. Mapping below.

### Security Researcher

- **[B1] Orphan chain validation** — §Design / *unlink-while-open + orphan list* now specifies a bounded walk capped at `s_inodes_count`, a `BitVec` visited set for cycle detection, bounds `[EXT2_GOOD_OLD_FIRST_INO, s_inodes_count]` on every chain link, and explicit rejection of root/reserved inodes. On a bad chain → force-RO, no panic, no loop. §Security Considerations now has a dedicated "Orphan-chain and directory-entry confusion" bullet restating these as normative.
- **[B2] Fast-symlink path confusion** — §Key data structures fixes the gate to `S_ISLNK(i_mode) && i_blocks == 0 && i_size <= 60` (all three), and the user-copy is `min(i_size, 60, user_buflen)`. Workstream D task updated to match.
- **[B3] Indirect-block pointer bounds** — §Indirect-block walker now requires every pointer (direct + single/double/triple-indirect) to be validated against `[s_first_data_block, s_blocks_count)` *and* a metadata-forbidden bitmap computed at mount (rejects pointers aliasing the superblock, BGDT, bitmaps, or inode tables — closing the confused-deputy aliasing attack). Failure → `EIO` + force-RO. §Security has a matching bullet.
- **[B4] A-before-B merge gate** — §Security bullet 1 rewritten as a normative CI gate: Workstream A syscalls gate behind `#[cfg(feature = "vfs_creds")]`, plus a grep CI check on A's PRs that fails if B's `Task::credentials` is absent from main. Workstream A task list preamble restates this.
- **[B5] Dirent validation** — §Directory operations now specifies a normative per-record validation list that includes `name_len > 0` on live records, `file_type` ∈ known set under `INCOMPAT_FILETYPE`, and unconditional skip of `inode == 0` at any slot (not just first). Live dirents with `inode < EXT2_GOOD_OLD_FIRST_INO` (not root) or `> s_inodes_count` are `EIO`.

### OS Engineer

- **[B1] Spin-lock / blocking-lock inversion** — new §Lock ordering subsection publishes the total order and pins the invariant: "no VFS lock is ever held across a synchronous block I/O wait." `BufferHead.data` is released before `sync_dirty_buffer`'s device wait. Workstream C acquires this as a task-level invariant assertion.
- **[B2] Orphan list strong ref** — §In-memory types changed: `orphan_list: Mutex<BTreeMap<u32, Arc<Inode>>>` (strong refs), plus `Ext2Inode::unlinked: AtomicBool` that vetoes `gc_queue` eviction. Ownership contract documented inline.
- **[B3] Rename lock order** — §Lock ordering extends the existing ramfs/super_block contract (`sb.rename_mutex` > ino-ordered `dir_rwsem`s) down through per-Ext2Fs and buffer-cache locks, with explicit ordering of all the per-Ext2Fs locks.
- **[B4] BufferHead eviction** — §Buffer cache now has four normative eviction invariants: skip pinned (`Arc::strong_count > 1`), skip DIRTY+LOCKED_IO, never sync-flush from `bread` (returns `ENOMEM` instead), single-cache-entry invariant (insert under `entries.write()`).
- **[B5] Sync-on-eviction hazard** — Closed by eviction invariant #3 above: `bread` never writes. Writeback is only ever performed by the background daemon or explicit `sync_fs`. Workstream C task updated.
- **[B6] Allocation/free rollback** — new §Create write-ordering subsection specifies the soft-updates sequence (bitmap → inode → dirent, each `sync_dirty_buffer`), plus explicit rollback if step 3 fails. Rollback-of-rollback failure → force-RO + clear valid-FS flag. §rename and truncate paths pick up the same discipline.

### User Space Staff Engineer

- **[B1] mmap scope** — §Kernel-Userspace Interface now has an explicit "File-backed mmap scope" subsection: both `MAP_SHARED` and `MAP_PRIVATE` return `ENODEV`; `execve` rejects `PT_INTERP` ELFs from ext2-backed inodes with `ENOEXEC` so dynamic-link failure is unambiguous. Epic's userspace target is statically-linked binaries only. Workstream F task list adds the execve gate.
- **[B2] `utimensat` semantics** — new subsection pins every case: `UTIME_NOW` (write OR owner), `UTIME_OMIT` (leave), explicit times (owner-only), `AT_SYMLINK_NOFOLLOW`. Errno table added.
- **[B3] `chmod`/`chown` SUID/SGID clearing** — new subsection: `chown` always clears `S_ISUID` on success and `S_ISGID` when `S_IXGRP` is set (matching Linux); `chmod` silently clears `S_ISGID` when caller is not a group member. Enforced in `Ext2Inode::setattr`.
- **[B4] Per-syscall errno tables** — full table added covering rename/unlink/rmdir/link/symlink/mkdir/truncate/chmod/chown/mount/umount2/setuid/setgid/setgroups/utimensat/readlink/execve plus common errors (`ENAMETOOLONG`, `EROFS`, `EIO`).

### Filesystem Engineer

- **[B1] `s_first_data_block` bitmap math** — §Key data structures / "Bitmap-to-absolute-block math" spells out `s_first_data_block + G * s_blocks_per_group + i` and mandates a mount-time assertion on the 1 KiB vs ≥2 KiB invariant. Partial-last-group bits past `s_blocks_count`/`s_inodes_count` treated as set.
- **[B2] `i_dtime` dual-use** — §unlink-while-open + orphan list now opens with a "`i_dtime` dual-use rule" that makes the chain-next-pointer-vs-wall-clock sequencing normative.
- **[B3] Orphan-list op ordering** — normative final-close sequence: (1) unlink from on-disk chain + sync, (2) free data/indirect blocks + sync bitmap, (3) free inode bit + sync, (4) *only now* write wall-clock `i_dtime` + zero out inode slot + sync. Reversal → `e2fsck` complaints or worse.
- **[B4] `i_blocks` update rule** — §Key data structures spells out `(block_size/512)` per allocated block including indirect blocks. §Indirect-block walker and Workstream E tasks updated.
- **[B5] Dirent `inode==0`** — §Directory operations specifies universal skip (not just first slot) by all iterators, and insert-path reuse of any zero-inode slot.
- **[B6] Rename link-count ordering** — §rename step sequence is now: `i_links_count++` + sync → write new dirent + sync → delete old dirent + sync → `i_links_count--` + sync. Overwrite path rewrites target dirent ino in place. Workstream E task updated.
- **[B7] Create write-ordering** — §Create write-ordering subsection makes the soft-updates sequence (bitmap → inode → dirent, each `sync_dirty_buffer`) normative.

### Advisory polish also addressed

- RMW inode writes preserving `l_i_uid_high`/`l_i_gid_high` and unknown reserved fields (Filesystem A1, A3).
- Nanosecond timestamp truncation documented (Filesystem A2).
- Writeback daemon lifecycle with `SbActiveGuard` + `draining` interlock (OS A3).
- `s_state` clear skipped on RO mounts; RW mounts sync the superblock before `mount(2)` returns (OS A4, Filesystem A7).
- Real-vs-effective UID distinction for `access`/`faccessat2` (Userspace A2).
- Rootfs image reproducibility + hash verification (Security A2).
- SMP scope bound with `#[cfg(not(smp))]` assertion; follow-up RFC required before SMP (Security A4, OS A5).
- `BlockDevice` trait introduced as Workstream C task (OS A1).
- Supplementary-groups-on-uid-transition pinned to Linux rule (Security A3).

### Not adopted (with reason)

- `NGROUPS_MAX = 32` remains for MVP; moved to Open Questions to revisit during Workstream B. Raising it is a one-line change with no other implications.
- Card §3.2 misattribution and CLOCK-Pro small-cache framing (Academic A1, A2) — softened in §Background / §Design; Academic verdict was LGTM so these don't affect the cycle.

### Re-review scope

Academic issued LGTM on cycle 0 and is not re-spawned. Security, OS Engineer, User Space Staff Engineer, and Filesystem Engineer will re-review the updated RFC and post new verdicts superseding their previous ones.

dburkart:

**Summary:** All five cycle-0 blocking findings are now closed by normative text in the RFC. The Security Considerations section, the Directory operations section, the Indirect-block walker section, and the on-mount orphan-chain validation block now spell out concrete bounds, cycle detection, metadata-aliasing rejection, and the A-before-B sequencing gate in a way that an implementer cannot accidentally weaken. Cleared to merge.

**Blocking findings:**

(none — all five from cycle 0 are resolved)

- ~~[B1] Orphan chain validation~~ — Closed. §unlink-while-open + orphan list / "On-mount orphan-chain validation" pins the bounded walk (`s_inodes_count` cap), `BitVec` visited set for cycle detection, ino bounds `[EXT2_GOOD_OLD_FIRST_INO, s_inodes_count]`, and explicit refusal of root + reserved 1..10 inodes; cycle/bad-chain → force-RO with no panic, no infinite loop. §Security Considerations restates the same as a normative bullet.
- ~~[B2] Fast-symlink gate~~ — Closed. §Security Considerations / "Fast-symlink path confusion" requires `S_ISLNK(i_mode) && i_blocks == 0 && i_size <= 60` (all three predicates) and clamps the user copy to `min(i_size, 60, user_buflen)`. Workstream D task list mirrors this verbatim. The leak-of-`i_block[]`-bytes and adjacent inode-table memory paths are both closed.
- ~~[B3] Indirect-block pointer bounds~~ — Closed. §Indirect-block walker rejects `p < s_first_data_block`, `p >= s_blocks_count`, *and* any pointer that aliases a metadata region (superblock, BGDT, group bitmaps, inode tables) via a mount-time "metadata-forbidden" bitmap. Failure → `EIO` + force-RO. §Security Considerations names the confused-deputy attack explicitly. This closes the user-data-write-overwrites-`i_mode` privilege-escalation path.
- ~~[B4] A-before-B workstream sequencing~~ — Closed. §Security Considerations bullet 1 is now belt-and-suspenders: every Workstream A syscall dispatch arm is gated behind `#[cfg(feature = "vfs_creds")]` (returns ENOSYS until B flips the feature), AND a CI grep check on A's PRs fails if `Task::credentials` is not yet on main. The "no merge ordering exposes an unauthenticated DAC-bypass window, even transiently" claim is now defensible.
- ~~[B5] rec_len / dirent validation~~ — Closed. §Directory operations / "Per-record validation" requires `rec_len >= 8` + 4-byte aligned, `rec_len + cursor <= block_end`, `name_len + 8 <= rec_len`, and additionally `name_len > 0` on live entries, `file_type` ∈ known set under `INCOMPAT_FILETYPE`, universal `inode == 0` skip at any slot (lookup, getdents64, insert scan, rename walk), `inode < EXT2_GOOD_OLD_FIRST_INO` (excluding ROOT_INO) → EIO, `inode > s_inodes_count` → EIO, and `Ext2Fs::iget(0)` is statically unreachable so a stray ino-0 cannot index the inode table at a negative offset.

**Advisory findings:**

- [A1] Metadata-forbidden bitmap memory cost — On large volumes the per-block "is-metadata" bitmap is `s_blocks_count / 8` bytes (e.g. 32 MiB for a 256 GiB / 1 KiB-block volume; 4 MiB for 256 GiB / 4 KiB-block). For MVP single-image workloads this is fine, but worth noting in §Performance Considerations or as an Open Question whether to switch to a sorted-range representation later. Non-blocking — easy to revisit when SMP/large-volume work begins.
- [A2] Errno disclosure oracle for indirect-block aliasing — The current §Security text folds all "off-disk / malformed / metadata-alias" cases to `EIO`, which is correct, but a perf-counter or dmesg log line per metadata-aliasing rejection could itself become a probe oracle if exposed to unprivileged users. Recommend rate-limiting + dmesg-only (root-readable) for these specific log paths during Workstream D.
- [A3] Force-RO transition observability — Several paths (bad orphan chain, bad indirect pointer, dirent corruption) silently force-RO. Userspace processes already holding write fds will start seeing `EROFS` mid-syscall with no other signal. Consider a normative requirement that force-RO transitions emit a single high-priority kmsg with the triggering inode + reason, so post-mortem on a hostile-image incident is tractable. Non-blocking; can land with Workstream D.

**Verdict:** LGTM

dburkart:

**Summary:** All four cycle-0 blocking findings are closed. The new "File-backed mmap scope", "`utimensat(2)` semantics", "`chmod`/`chown` SUID/SGID clearing", and "Per-syscall errno table" subsections in §Kernel-Userspace Interface are normative, internally consistent, and correctly cross-referenced from the Workstream A/E/F task lists. Userspace can confidently bind to the listed errnos and pjdfstest will not ping-pong on ambiguous returns.

**Blocking findings:**
_(none — all four cycle-0 blockers resolved)_

- [B1] mmap scope — closed. ENODEV is now scoped explicitly to both `MAP_SHARED` and `MAP_PRIVATE`, the static-binary-only epic boundary is documented, and the up-front `execve(PT_INTERP)` → `ENOEXEC` gate (§Workstream F task list) replaces the previous "opaque mapping failure deep inside `execve`" failure mode. This is exactly option (b) from cycle 0; acceptable.
- [B2] `utimensat` — closed. Each of `times==NULL` / per-field `UTIME_NOW` / `UTIME_OMIT` / explicit-time / `AT_SYMLINK_NOFOLLOW` has its semantics and per-case permission rule spelled out. The owner-only requirement for explicit-time is correctly justified as anti-forensics. The nanosecond-truncation note matches §Key data structures. `EACCES` vs `EPERM` split for the two permission classes is unambiguous.
- [B3] SUID/SGID clearing — closed. `chown` always-clear-`S_ISUID` + conditional-`S_ISGID`-on-`S_IXGRP`, `chmod` silent-`S_ISGID`-clear-for-non-group-member, and the `write(2)` SUID/SGID-clear-on-success rule (matching Linux) are all normative. Enforcement is pinned to `Ext2Inode::setattr` with the line "misses here are CVEs, not bugs," which is the right framing.
- [B4] Per-syscall errno table — closed. The MVP table covers rename/unlink/rmdir/link/symlink/mkdir/truncate/chmod/chown/mount/umount2/setuid/setgid/setgroups/utimensat/readlink/execve plus the universal `ENAMETOOLONG`/`EROFS`/`EIO` rows. The "POSIX permits either; we pick one and commit" framing (e.g., `rename` → `ENOTEMPTY` over `EEXIST`) is exactly the contract userspace needs. The setresuid/setreuid saved-set-uid prose at the end is a nice bonus.

**Advisory findings:**
- [A1] Errno table — `EOVERFLOW` on `chown` for uid/gid > u32 (mentioned in §Key data structures bullet on `l_i_uid_high`/`l_i_gid_high`) is missing from the table row. Add a `chown` / `EOVERFLOW` row so userspace doesn't have to cross-reference two sections.
- [A2] Errno table — `ELOOP` for symlink-loop path resolution (POSIX-mandated, fires after `SYMLOOP_MAX = 8`/40 hops) is not enumerated for any of the path-taking syscalls. Probably belongs as a row under "Any path" alongside `ENAMETOOLONG`. Not strictly an ext2 epic concern (it's a VFS resolver concern), but pjdfstest will exercise it.
- [A3] Errno table — `link` is missing the very common `EACCES` (no write permission on the target directory) and `ENOENT` (source/target dir component missing) rows; `mkdir` is missing the same. The "MVP" framing covers this implicitly, but listing the high-frequency-from-userspace rows would shorten implementer review.
- [A4] `chown` to a different group by a non-privileged caller who *is* a member of the target group: POSIX permits this; the table only documents `EPERM` for "different user by non-privileged" — please confirm in the table that "different group, caller is member" is allowed (or call it out as `EPERM` if not). Affects `pjdfstest` `chown/00.t`.
- [A5] `execve` `ENOEXEC` for ext2-backed `PT_INTERP` is the right call but should also surface a kernel-printed warning string at first-occurrence so that someone running a stock `/bin/sh` and seeing `ENOEXEC` knows immediately to rebuild static rather than chase a corrupted-binary phantom. Boot-time scan is overkill; first-failure log line is enough.
- [A6] `getgroups`/`setgroups` errno rows could note `EFAULT` for a bad userspace buffer pointer; the "Any path / `EFAULT`" pattern from `utimensat` doesn't generalize to non-path syscalls. Minor.

**Verdict:** LGTM

dburkart:
### Review: Filesystem Engineer

**Summary:** All seven of my previous blocking findings are now addressed normatively in the RFC. The new "Bitmap-to-absolute-block math", "i_dtime dual-use rule", "Create write-ordering", and updated unlink/orphan + rename sequences match the on-disk contracts that `e2fsck -fy` expects. A handful of small bookkeeping items remain (advisory only).

**Blocking findings:**

(none)

**Advisory findings:**

- [A1] Superblock-level free counters not explicitly named alongside BGDT counters — §Create write-ordering and §unlink-while-open + orphan list both call out `bg_free_inodes_count` / `bg_free_blocks_count` updates and sync, but never mention the corresponding `s_free_inodes_count` / `s_free_blocks_count` superblock fields. They must be incremented/decremented symmetrically with every bitmap mutation, or `e2fsck -fy` will rewrite the superblock on every touched image with "Free inodes count wrong" / "Free blocks count wrong". Recommend adding "and `s_free_{inodes,blocks}_count` mirrored under the same sync" to the create, unlink-final-close, and rollback steps, plus a Workstream E task line.
- [A2] §rename overwrite path — relative ordering of source-side `i_links_count--` (step 4) and victim `i_links_count--` (step 5) and their respective `sync_dirty_buffer`s is left implicit. If both inodes share an inode-table block they fold to one write; if they don't, two separate syncs are needed. A one-line clarification ("each decrement is followed by `sync_dirty_buffer(inode_table_bh)` for that inode's block; if the two inodes share a block, one sync suffices") would close it.
- [A3] §Create write-ordering — for `mkdir`, the new directory's own `i_links_count` initial value should be pinned to 2 (one from the dirent in the parent, one from its own `.`) before step 3's sync, not left implicit under "i_links_count >= 1" in step 2. `e2fsck` will fix it but it's worth being normative.
- [A4] §unlink-while-open + orphan list, sequence step 1 ("Unlink from on-disk orphan chain first") — when the inode being freed is the chain head, the protocol updates `sb.s_last_orphan := next` and syncs the superblock; when it's mid-chain, the predecessor's `i_dtime` is rewritten and the predecessor's inode-table block is synced. Good. But if the predecessor's inode is itself in the in-memory orphan list and being freed concurrently, the two finalizers can race on the chain. Recommend stating that the orphan-chain mutation is serialized under `Ext2Fs.orphan_list` (the existing lock-order entry), so only one CPU at a time walks/edits the chain. (Probably fine on UP; pin it anyway for SMP follow-up.)
- [A5] §Mount step 7 walks the on-disk orphan chain "before allowing userspace access" — good. Worth stating that on a successful drain, `sb.s_last_orphan` is set to 0 and synced as the final step of recovery, otherwise a subsequent mount sees a pre-drained chain and runs the validation pass redundantly (cheap, but a clean post-recovery state simplifies fsck reasoning).

**Verdict:** LGTM

dburkart:
### Review: OS Engineer

**Summary:** All six previous blocking findings are adequately closed. The new §Lock ordering subsection, the four normative buffer-cache eviction invariants, the strong-ref orphan list with `unlinked`-AtomicBool veto, and the §Create write-ordering rollback rule together resolve every concurrency / lifecycle / inversion hazard I called out in cycle 0. Two small advisory notes on phrasing and the SMP-future strong_count check.

**Blocking findings:**

(none — all six previous B-findings are closed)

- [B1-closed] Spin-lock / VFS-lock inversion — §Lock ordering pins "no VFS lock is ever held across a synchronous block I/O wait", with `BufferHead.data` released before `sync_dirty_buffer`'s device wait and Workstream C carrying it as a task-level invariant assertion. Closes the priority-inversion hazard.
- [B2-closed] Orphan-list ownership — `orphan_list: Mutex<BTreeMap<u32, Arc<Inode>>>` now holds strong refs, plus the `Ext2Inode::unlinked: AtomicBool` veto in `Inode::drop` provides belt-and-suspenders against an `evict_inode` call racing the orphan-final-close free path.
- [B3-closed] Rename lock order — §Lock ordering publishes a total order from `sb.rename_mutex` down through ino-ordered `dir_rwsem` pair, per-inode `meta` / `block_map`, `Ext2Fs.bgdt`, `orphan_list`, `BlockCache.entries`, and `BufferHead.data`. Extends the existing ramfs/super_block ino-order contract cleanly.
- [B4-closed] BufferHead eviction race — four normative invariants in §Buffer cache: skip `Arc::strong_count > 1`, skip `DIRTY + LOCKED_IO`, no sync flush from `bread` (returns `ENOMEM`), and single-cache-entry insert under `entries.write()`. The coexistence of two `Arc<BufferHead>`s for the same `(dev, blk)` is now structurally impossible.
- [B5-closed] Sync-on-eviction inside `read()` — eviction invariant #3 forbids `bread` from calling `sync_dirty_buffer`; writeback is daemon-only or explicit `sync_fs`. The block-device spin lock can no longer be acquired transitively under a VFS lock.
- [B6-closed] Alloc/free rollback symmetry — §Create write-ordering specifies the soft-updates `bitmap → inode → dirent` sequence (each `sync_dirty_buffer`), explicit step-3-failure rollback of step 1, and rollback-of-rollback failure → force-RO + clear valid-FS flag. The same discipline is extended to `rename` and `setattr`-for-truncate.

**Advisory findings:**
- [A1] `unlinked` AtomicBool veto wording — §In-memory types reads "if set, eviction/`evict_inode` is vetoed (the orphan-list-final-close path owns the free)." Read literally this risks a leak: after the §final-close sequence step 5 drops the `Arc<Inode>` from `orphan_list`, `unlinked` is still `true`, and Inode::drop would veto. The intended reading is that the veto skips only the FS-level `evict_inode` hook (on-disk free already done by the orphan path) — the in-memory destructor still runs. Worth one explicit sentence in §In-memory types or §unlink-while-open clarifying that the veto is on the evict_inode hook, not on Drop.
- [A2] `Arc::strong_count > 1` check is racy under the future SMP RFC — the lock-ordering invariant says §Lock ordering is the normative contract for "any future SMP landing", but the eviction sweep that consults `strong_count` must take that read while holding `entries.write()` (otherwise a concurrent `bread` on another CPU can clone the Arc between the check and the unlink-from-map). Invariant #4 already requires `entries.write()` for insert; please state explicitly that the sweep's strong_count check is also under `entries.write()`. Trivially correct on UP today; flagged so the SMP follow-up RFC inherits the invariant in writing.
- [A3] Lock-order entry for `Ext2Fs.bgdt` vs `orphan_list` — final-close sequence steps 2/3 free blocks and inode bits (touching `bgdt`) while the inode is on `orphan_list`. The lock order shows `bgdt` above `orphan_list`, which is the intended direction (acquire bgdt first when both are needed); please add a one-liner that final-close releases the `orphan_list` Mutex before re-acquiring `bgdt` for the bitmap path, so a future implementer doesn't accidentally invert this. Non-blocking.

**Verdict:** LGTM

Issue #531, opened 2026-04-18
repro-fork harness runs ~15,000 cycles per boot in CI — every soak iteration hits HARD_CAP


dburkart:
## Motivation

The 100× smoke soak on main (run [24595210929](https://github.com/dburkart/vibix/actions/runs/24595210929)) fails every iteration at exactly the `REPRO_HARD_CAP = 900s` wall-clock ceiling. Post-cancellation artifact analysis shows the repro-fork harness is running **~14,988 fork+exec+wait cycles per boot**, not the advertised 500-cycle default. At ~60 ms/cycle under un-accelerated CI QEMU (TCG, no KVM) that produces 900s of work per boot, the xtask kills QEMU at HARD_CAP, the soak records the run as `fail`, and the job wastes a full 6-hour timeout with 0% pass rate.

This makes smoke-soak pass-rate accounting meaningless and blocks the epic #501 acceptance gate (`SOAK_MIN_PASS_RATE=80` is unreachable when every run is a 907s hard-cap fail).

## Evidence — run-001.log from cancelled run 24595210929

| Signal | Count |
|---|---|
| `tasks: reaped task N` lines | 14,988 (task IDs 5 → 14,981) |
| `ring3-iretq: rip=0x400000` (hello re-entry post-exec, 1 per cycle) | 14,987 |
| `hello: hello from execed child` | 14,988 |
| `repro: cycle K alive` heartbeat lines (should fire every 50 cycles) | **0** |
| `repro: fork loop complete cycles=N` success marker | **0** |
| `repro: starting fork loop cycles=` banner | 1 (but `cycles=` number is interleaved with first child's `ring3-iretq` output on the shared serial, so the CYCLES value in the banner is not legible) |
| summary.tsv | run 1: 907.4s fail / run 2: 907.3s fail |

14,988 cycles × ~60 ms ≈ 899 s — arithmetic is consistent: the harness runs at expected per-cycle speed, just for **30× more iterations than the 500-cycle default**.

## Suspected root causes (in rough likelihood order)

[...]

PR #583, opened 2026-04-18
kernel/idt: skip ring3-first-fault log for resolved user #PFs

### LINKED ISSUE #527: kernel: residual fork flake manifests as ring3 user-stack-top #PF at rip=0x4002c6

dburkart:
Closes #527

## Summary

- The `ring3-first-fault:` one-shot diagnostic was firing on the first CPL=3 page fault of the session — which in steady state is always a legitimate CoW first-touch on the parent's post-fork stack page. The fault is handled correctly by `cow_copy_and_remap`, but the log line made the 100× soak output look like a live flake and blocked the epic #501 three-green-nightlies gate per issue #527.
- Move the `#PF` call to `log_first_ring3_fault` into the access-violation fall-through branch so the marker only fires for faults the kernel could not resolve (real SIGSEGV-class bugs — the original #478 intent). SMAP / RSVD / USER_VA_END panic paths each emit their own dedicated EXCEPTION line already; `#UD` / `#GP` still log unconditionally because neither has a legitimate ring-3 resolution path.
- No userspace changes — PR #532 already fixed the inline-asm clobber bug that had been conflating the runaway-cycles failure mode with this fault.

## Root cause (1 sentence)

The latch at the top of the `#PF` handler emitted on every first CPL=3 fault regardless of outcome, so the normal post-fork CoW fault on the parent's stack (`rip` just past `syscall`, `cr2` at `rsp+0`) tripped it on cycle 1 of every boot — issue #527's "smoking gun" was the diagnostic, not a bug in the fault path.

## Test plan

- [x] `cargo xtask build` — ok.
- [x] `cargo xtask smoke` — all 33 markers present, 0 `ring3-first-fault` lines in serial log.
- [x] `cargo xtask repro-fork` — 500 cycles complete, 0 `ring3-first-fault` lines, every `repro: cycle K alive` heartbeat at K=50..500 fires.
- [x] `cargo xtask test` — 377 host unit tests + integration tests pass. The `page_fault.rs` test still drives `test_hook::expect_page_fault` (which runs before the moved log call), so its exit-on-match contract is unchanged.
- [x] `cargo test -p xtask` — 34 tests pass.
- [x] Longer `REPRO_FORK_CYCLES=50000 cargo xtask repro-fork` run makes forward progress past cycle 13200 with zero `ring3-first-fault` markers before hitting an unrelated heartbeat stall from accumulated stack-VA-slot leaks (separate #501 sub-issue — the `stack VA slot ... leaked` line this PR does not chase).
- [ ] CI green on fmt+build / host unit tests / integration tests (QEMU) / smoke.
- [ ] Post-merge manual `workflow_dispatch` of `smoke-soak` on main (count=100, min_pass_rate=80), expected pass rate ≥ 80 %.

### DIFF
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -14,16 +14,29 @@ use crate::arch::x86_64::interrupts::{
 };
 use crate::serial_println;
 
-/// #478 diagnostic latch: the first CPL=3 fault of the session logs a
-/// `ring3-first-fault:` line with full frame state before the handler
-/// runs its normal path. Subsequent ring-3 faults are common (SIGSEGV
-/// delivery etc.) and must not flood the log. One-shot by design.
+/// #478 / #527 diagnostic latch: the first *unexpected* CPL=3 fault of the
+/// session logs a `ring3-first-fault:` line with full frame state before
+/// the handler runs its normal path. Subsequent ring-3 faults are common
+/// (SIGSEGV delivery etc.) and must not flood the log. One-shot by design.
+///
+/// "Unexpected" here means a fault that the kernel could not resolve via
+/// its normal dispatch (CoW first-touch, demand-page fill, growsdown stack
+/// extension, etc.).  The `#PF` path only calls this helper on the
+/// fall-through / access-violation branch; the successfully-handled
+/// CoW / demand-page returns skip the log entirely.  That keeps the marker
+/// as a high-signal first-unexpected-fault breadcrumb instead of firing on
+/// every boot's legitimate first-fork CoW fault (issue #527).
 static FIRST_RING3_FAULT_LOGGED: AtomicBool = AtomicBool::new(false);
 
-/// Emit one `ring3-first-fault:` line on the first CPL=3 fault ever
-/// observed. `extra` is vector-specific trailer (e.g. `code=0x0` for #GP,
-/// `cr2=0x400000 code=PROTECTION` for #PF) — no user memory is touched
-/// while formatting, only scalars copied out of the hardware frame.
+/// Emit one `ring3-first-fault:` line on the first *unresolved* CPL=3 fault
+/// ever observed. `extra` is vector-specific trailer (e.g. `code=0x0` for
+/// `#GP`, `cr2=0x400000 code=PROTECTION` for `#PF`) — no user memory is
+/// touched while formatting, only scalars copied out of the hardware frame.
+///
+/// Callers: `#UD` / `#GP` call unconditionally (no legitimate resolution
+/// path for those in ring 3). `#PF` calls only after the VMA-backed
+/// resolver and growsdown extension have both declined to handle the
+/// fault — so a routine CoW or demand-page fault is silent.
 fn log_first_ring3_fault(vector: &str, frame: &InterruptStackFrame, extra: core::fmt::Arguments) {
     if (frame.code_segment.0 & 0b11) == 0 {
         return;
@@ -136,11 +149,12 @@ extern "x86-interrupt" fn page_fault(mut frame: InterruptStackFrame, code: PageF
     }
     let addr_u64 = x86_64::registers::control::Cr2::read_raw();
 
-    log_first_ring3_fault(
-        "#PF",
-        &frame,
-        format_args!("cr2={:#x} code={:?}", addr_u64, code),
-    );
+    // Note: log_first_ring3_fault is intentionally NOT called here.  The
+    // diagnostic latch fires only on CPL=3 faults that reach the
+    // access-violation fall-through below, so a routine CoW first-touch
+    // or demand-page fault on the parent's post-fork stack page (issue
+    // #527) stays silent.  SMAP, RSVD, and USER_VA_END panics log their
+    // own dedicated exception lines, so they don't need the marker.
 
     if let Some(expected) = crate::test_hook::take_page_fault_expectation() {
         use crate::test_harness::{exit_qemu, QemuExitCode};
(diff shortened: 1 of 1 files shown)


dburkart:
Scope redirect before this lands — the technical change is fine, but the framing overreaches.

## What's correct

The core claim holds: the one-shot `ring3-first-fault:` diagnostic as written fires on the *first* CPL=3 `#PF` of the boot regardless of outcome, and in steady state that's always a legitimate CoW first-touch on the post-fork stack (or similar handled fault). Moving the emit into the access-violation fall-through is the right shape for the diagnostic — real SIGSEGV-class bugs still log, legitimate CoW doesn't. This is a useful hygiene change and worth landing on its own merits.

## What's wrong

This PR should **not** `Closes #527`. That issue was effectively resolved by #531 → #532 (the SysV clobber fix that eliminated the harness cycle-counter corruption which was the actual cause of the flake whose smoking gun *looked* like the log line). #583 is diagnostic hygiene *about* #527's report — not a second independent fix. Claiming closure lets the PR inherit credit for work that already landed, and obscures the fact that #527's acceptance ("50k repro-fork cycles complete without hang") was met in the 100× soak [24596626665](https://github.com/dburkart/vibix/actions/runs/24596626665) under commit `b3571a5`, before #583 was written.

Also: #478 (silent init hang at `rip=0x400000` post-iretq, no diagnostic at all) is a separate, currently-live class of flake — evidence in runs [24597818762](https://github.com/dburkart/vibix/actions/runs/24597818762) and [24597457842](https://github.com/dburkart/vibix/actions/runs/24597457842). #583 does not and cannot address it, since the failure mode of #478 is that **no fault is taken**. Please make that clear so a future reviewer doesn't read "fork/exec flake closed" and assume CI is clean.

## Requested changes before merge

1. Drop `Closes #527`. Replace with `Refs #527 — diagnostic hygiene follow-up to #531/#532`.
2. Retitle to something like `kernel/idt: tighten ring3-first-fault log to unhandled faults only` (removes the bug-fix framing).
3. Rewrite the summary's first sentence to describe it as diagnostic tightening, not a bug fix. Explicitly note:
   - #527's underlying flake was fixed by #532.
   - #478's silent hang is out of scope and still open.
4. Integration tests are currently **FAILURE** on this PR. Before merging, confirm that's not a regression introduced by this change (it likely isn't — a moved log call shouldn't affect integration). If it's the #478 flake catching this run, that's expected but please re-run and confirm the second run is green; if integration stays red on reruns, stop and investigate before merging.

Happy to land this once the framing is fixed and integration is confirmed green.

dburkart:
Superseded as part of the #501 quality sprint (phase-4 orchestration, 2026-04-23).

The chosen path is to stack the #527 CoW fix on top of PR #595's diagnostic-marker branch (`m478-v3`) rather than suppress the `ring3-first-fault` log for resolved #PFs. Rationale:

- PR #595's markers make #527 a deterministic repro (not 50% intermittent). That repro is hard-won — dropping it and re-establishing it from main would waste that signal.
- Log-suppression here masks the actual CoW permission miss described in #527's body (PTE present-but-RO on the child's first user-stack store). The underlying fault needs a real fix in the VMA / stack-slot path, not a quieter log.
- Closing this PR keeps main's smoke from being poisoned by a band-aid while the real fix lands in a single merge that subsumes #478/#484/#527.

Closing without merging. #583's diff is preserved in git history if the log-tidy-up turns out to be independently wanted later.

Issue #553, opened 2026-04-18
block: bread/mark_dirty/sync_dirty_buffer/release API with CLOCK-Pro eviction


dburkart:
## Motivation

Implements RFC 0004 — Workstream C (Block buffer cache, wave 1). The main cache API
plus CLOCK-Pro eviction. The four normative invariants from the RFC MUST be enforced.

## Work

- [ ] `bread(device_id, block_no) -> Result<Arc<BufferHead>, BlockError>` — fetches the
      buffer; if miss, allocates + reads from device; if cache is full, runs CLOCK-Pro
      eviction first. **Returns `ENOMEM` rather than synchronously flushing dirty
      buffers** (normative invariant 3).
- [ ] `mark_dirty(&Arc<BufferHead>)` — sets DIRTY bit + inserts into `dirty` set
- [ ] `sync_dirty_buffer(&Arc<BufferHead>)` — sets LOCKED_IO, releases inner `BufferHead.data` lock,
      calls `device.write_at`, clears DIRTY and LOCKED_IO atomically, removes from dirty set
- [ ] `release(&Arc<BufferHead>)` — optional hint; drops one strong ref; real GC is via Arc drop
- [ ] CLOCK-Pro eviction (Jiang 2005):
  - hot-queue + cold-queue + non-resident metadata
  - rotate `clock_hand` through the entries map
  - **skip any `Arc::strong_count > 1` buffer** (normative invariant 1 — external refs pin)
  - **skip any DIRTY + LOCKED_IO buffer** (normative invariant 2)
  - evict cold-queue victims first; promote hot-queue on reference-bit hit
- [ ] **Single-cache-entry invariant** (normative invariant 4): `bread` must never return two
      different `Arc<BufferHead>` for the same `(DeviceId, block_no)`; guard via `entries` map lookup
      inside the insertion critical section
- [ ] Unit tests for: hit path, miss path, eviction preferring cold-queue, refusal to evict pinned buffer,
      refusal to evict DIRTY+LOCKED_IO, ENOMEM when max_buffers reached and all entries are pinned

## Out of scope

- `sync_fs` and writeback daemon (next C issues).
- Invariant assertion (last C issue).

## Context

- Parent epic: #537
- Depends on: #552 (BufferHead/BlockCache skeleton)
- RFC 0004 §Buffer cache eviction (CLOCK-Pro, normative invariants 1–4)
[...]

PR #587, opened 2026-04-18
block: BufferHead + BlockCache skeleton keyed on (DeviceId, u64)

### LINKED ISSUE #552: block: BufferHead + BlockCache skeleton keyed on (DeviceId, u64)
## Motivation

Implements RFC 0004 — Workstream C (Block buffer cache, wave 1). The core data
structures for the cache: `BufferHead` wraps a block-sized slab with state bits, and
`BlockCache` is keyed on `(DeviceId, u64)` so multiple mounts share one cache without
aliasing.

## Work

- [ ] Add `kernel/src/block/cache.rs`:
      ```rust
      pub struct BufferHead {
          pub data: RwLock<Box<[u8]>>,   // block-sized
          pub state: AtomicU8,           // bitmask: VALID | DIRTY | LOCKED_IO
          pub clock_ref: AtomicBool,     // CLOCK-Pro reference bit
      }
      pub struct BlockCache {
          device: Arc<dyn BlockDevice>,
          block_size: u32,
          entries: Mutex<BTreeMap<(DeviceId, u64), Arc<BufferHead>>>,
          dirty: Mutex<BTreeSet<(DeviceId, u64)>>,
          clock_hand: Mutex<Option<(DeviceId, u64)>>,
          max_buffers: usize,
      }
      ```
- [ ] State constants: `VALID = 1`, `DIRTY = 2`, `LOCKED_IO = 4`
- [ ] `DeviceId` as an opaque index type registered by `BlockCache::register_device`
- [ ] `BlockCache::new(max_buffers)` constructor and `Arc<BlockCache>` singleton for the default instance
- [ ] `block_size` carried on the cache (not per-entry) so the walker never has to look up a device
- [ ] Unit test: insert, look up, remove round-trips without the eviction/read path

## Out of scope

- `bread`/`mark_dirty`/`sync_dirty_buffer` — next C issue.
- Eviction — next C issue.

## Context

- Parent epic: #537
- Depends on: #551 (BlockDevice trait)
- RFC 0004 §Key data structures (BlockCache)

dburkart:
Closes #552

## Summary

Lands the data-structure skeleton for the buffer cache described in RFC 0004 §Buffer cache (Workstream C, wave 1). Scope is intentionally narrow — the map key, state bits, device registry, and the insert / lookup / remove primitives — so the CLOCK-Pro eviction issue (#553) and the `bread` / `mark_dirty` / `sync_dirty_buffer` write-back path can land on top without touching `BufferHead` layout or the cache key shape.

- `BufferHead { data: RwLock<Box<[u8]>>, state: AtomicU8, clock_ref: AtomicBool }` with the RFC-mandated bitmask `VALID = 1 | DIRTY = 2 | LOCKED_IO = 4` exposed as public constants.
- `BlockCache` owns `Arc<dyn BlockDevice>`, carries `block_size` on the cache struct (not per-entry), and keys its `BTreeMap` on `(DeviceId, u64)` so two mounts of the same device can't alias.
- `DeviceId` is an opaque `Copy + Ord` newtype handed out by `BlockCache::register_device`.
- `init_default_cache` / `default_cache` expose the `Arc<BlockCache>` singleton hook the VFS mount path will consume.

Explicitly **out of scope** and tracked in follow-ups: `bread`, `mark_dirty`, `sync_dirty_buffer`, and CLOCK-Pro eviction (#553+). `clock_hand` and the `dirty` set are wired now so the eviction / write-back issues land as additive edits rather than refactors.

## Test plan

- [x] `cargo xtask build` — clean (no new warnings on `kernel/src/block/cache.rs`)
- [x] `cargo xtask test` — 391/391 host unit tests pass, including the 8 new `block::cache::tests::*`
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy --lib -p vibix` — no new warnings (pre-existing `build.rs` `is_multiple_of` lint is unrelated)

New tests cover:

- `state_bit_layout_matches_rfc` — pins `VALID=1 | DIRTY=2 | LOCKED_IO=4`.
- `buffer_head_new_is_block_sized_and_clear` — slab sizing, zero-state init, `state_has` mask semantics.
- `register_device_allocates_disjoint_ids` — monotonic, non-colliding device ids.
- `insert_lookup_remove_roundtrip` — the core skeleton contract from the issue body.
- `device_ids_do_not_alias_on_same_block_number` — two `DeviceId`s sharing `blk=42` get independent `BufferHead`s.
- `block_size_is_carried_on_cache_not_entry` — verifies the RFC requirement that block size lives on the cache struct.
- `default_cache_singleton_round_trip` — `init_default_cache` / `default_cache` idempotence under `spin::Once`.

### DIFF
--- a/kernel/src/block/cache.rs
+++ b/kernel/src/block/cache.rs
@@ -157,7 +157,28 @@ impl BlockCache {
     /// Construct an empty cache bound to `device` with entries sized at
     /// `block_size`. Caps the resident entry count at `max_buffers`; the
     /// cap is enforced by the eviction follow-up (#553), not here.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `block_size` is zero or not a whole multiple of
+    /// `device.block_size()`. Every future `bread` / `sync_dirty_buffer`
+    /// call would translate a cache-block index into a byte offset of
+    /// `blk * block_size`, and the device's `read_at` / `write_at`
+    /// reject any non-device-aligned offset with
+    /// [`BlockError::BadAlign`](super::BlockError::BadAlign). Catching
+    /// the geometry mismatch at construction is the difference between
+    /// "cache can never do I/O" (silent) and "misuse is caught at mount
+    /// time" (loud). Panicking is acceptable: `BlockCache::new` runs
+    /// only on the boot / mount paths, never inside a syscall.
     pub fn new(device: Arc<dyn BlockDevice>, block_size: u32, max_buffers: usize) -> Arc<Self> {
+        assert!(block_size > 0, "BlockCache: block_size must be non-zero");
+        let dev_bs = device.block_size();
+        assert!(
+            dev_bs > 0 && block_size % dev_bs == 0,
+            "BlockCache: cache block_size ({}) must be a whole multiple of device block_size ({})",
+            block_size,
+            dev_bs,
+        );
         Arc::new(Self {
             device,
             next_device_id: AtomicU32::new(0),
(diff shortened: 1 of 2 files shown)


dburkart:
Thanks for the review. Both findings addressed in `fae3dc9`:

1. **Geometry validation at construction** — `BlockCache::new` now panics if `block_size` is zero or not a whole multiple of `device.block_size()`. Catching the mismatch at mount time instead of on first `bread`'s `BadAlign` is the correct severity; `BlockCache::new` only runs on the boot/mount path, never in a syscall.

2. **`DeviceId` wraparound** — went with the narrower fix: `register_device` now panics on `u32::MAX` overflow rather than widening to `AtomicU64`. Keeping `DeviceId` at `u32` preserves the narrow `(DeviceId, u64)` key width — 12 bytes vs 16 — which matters because that pair is the `BTreeMap` node key on the hot block-cache path. 2³² mount registrations is an extreme upper bound we're not remotely near; catch-on-overflow gives the same correctness guarantee as widening, without bloating every key.

Tests added: `new_rejects_zero_block_size`, `new_rejects_incompatible_block_size`, `new_accepts_compatible_block_sizes`.

dburkart:
Second round of review addressed in `1e2b06d`:

1. **`register_device` wrap-before-panic** — switched to `fetch_update` so the counter isn't bumped on the overflow path. A panic-catcher would otherwise observe `0` on the next call and alias with whichever mount originally owned `DeviceId(0)`.
2. **Singleton test race** — `default_cache_singleton_round_trip` no longer does a `is_none()` check before init. That check was racy under parallel host tests (the `Once` slot is process-wide). Now unconditionally calls `init_default_cache` and asserts the guarantee that actually holds across the race: repeated `default_cache()` calls return the same `Arc`.

Issue #239, opened 2026-04-14
task: add per-process cwd with chdir/getcwd wiring


dburkart:
Implements RFC 0002 item 11/15. Per-process current working directory is a POSIX Confirmation in the RFC (not an open question). Needs a Task-struct field and fork/exec inheritance.

Work:
- Add cwd: Arc<Dentry> to the Task struct.
- fork copies cwd from parent; exec preserves cwd.
- Wire sys_chdir(path) and sys_getcwd(buf, len) syscalls.
- getcwd walks up via Dentry.parent until the root; returns ENAMETOOLONG if the full path exceeds buf.
- Host-side unit test.

Context: docs/RFC/0002-virtual-filesystem.md §Confirmations. Depends on #230, #238. Merged PR #224. ~100 LOC.

PR #589, opened 2026-04-18
kernel/syscall: wire unlink(2) and unlinkat(2) to VFS InodeOps

### LINKED ISSUE #539: sys: wire unlink/unlinkat to InodeOps::unlink

dburkart:
Closes #539

## Summary

- Wires SYS_unlink (87) and SYS_unlinkat (263, including AT_REMOVEDIR) through a shared unlinkat_impl that dispatches to InodeOps::unlink or InodeOps::rmdir. Enforces the POSIX pre-checks the FS-driver impls must not each reinvent: AT_REMOVEDIR-only flags whitelist, non-AT_FDCWD + relative rejected (until per-fd cwd lands in #239), trailing-slash directory semantics, leaf mountpoint EBUSY, kind dispatch (EISDIR/ENOTDIR), and parent sticky-bit S_ISVTX ownership enforcement. Leaf lookup uses parent.ops.lookup -- never path_walk -- so unlink("sym") never follows a symlink and deletes the wrong file.
- Follows the convention #585 set for mkdir: the dispatch arms are gated behind #[cfg(feature = "vfs_creds")] so ring-3 callers see -ENOSYS until Workstream B lands Task::credentials and flips the gate on (#546). The impl entry points (sys_unlink_impl / sys_unlinkat_impl) are always compiled and tests reach them directly. The vfs_creds feature, the default-off Cargo gate, and the CI A-before-B ordering job already exist on main from #585.
- Adds integration test syscall_unlink covering: syscall_dispatch(SYS_unlink, ...) == -ENOSYS (anti-regression anchor for the A-before-B ordering until #546 flips the gate on); sys_unlink_impl removes a regular file with subsequent stat ENOENT; the RFC 0004 errno table (EISDIR, ENOTDIR, ENOENT, EINVAL on unknown flags); unlinkat(AT_FDCWD, "/abs", 0) round-trip; unlinkat(AT_FDCWD, "/tmp/dir", AT_REMOVEDIR) removes an empty dir.

## Test plan

- [x] cargo fmt --all -- --check clean.
- [x] cargo xtask build clean (only pre-existing is_guard_hit dead-code warning).
- [x] cargo xtask test --test syscall_unlink -- all 8 tests pass.
- [x] cargo xtask smoke -- all 33 markers present.
- [x] cargo build --features vfs_creds --test syscall_unlink --target x86_64-unknown-none -Zbuild-std=... compiles cleanly.

Note: PR was rebased onto current main (mkdir #585 landed in parallel). Force-pushed with --force-with-lease. CodeRabbit re-review will trigger automatically.

Generated with Claude Code (https://claude.com/claude-code)

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -291,6 +291,10 @@ name = "syscall_rmdir"
 harness = false
 
 [[test]]
+name = "syscall_unlink"
+harness = false
+
+[[test]]
 name = "ntty_sigint_flush"
 harness = false
 
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -818,6 +818,19 @@ pub unsafe extern "C" fn syscall_dispatch(
         // InodeOps::rmdir on the parent after path-walking to the leaf.
         RMDIR => super::syscalls::vfs::sys_rmdir_impl(a0),
 
+        // unlink(path) — remove a non-directory from its parent. Same
+        // RFC 0004 A ↔ B gate as MKDIR/MKDIRAT: with `vfs_creds` off
+        // the arm is absent and the dispatcher's default returns
+        // `-ENOSYS`. Integration tests exercise the impl directly via
+        // `sys_unlink_impl` once the feature is on.
+        #[cfg(feature = "vfs_creds")]
+        UNLINK => super::syscalls::vfs::sys_unlink_impl(a0),
+
+        // unlinkat(dfd, path, flags) — unlink or rmdir (AT_REMOVEDIR)
+        // relative to `dfd`. Same gate as UNLINK.
+        #[cfg(feature = "vfs_creds")]
+        UNLINKAT => super::syscalls::vfs::sys_unlinkat_impl(a0 as i32, a1, a2 as u32),
+
         // poll(fds, nfds, timeout_ms) — wait for readiness on a set of fds.
         POLL => crate::poll::syscalls::sys_poll(a0, a1, a2 as i64),
 
@@ -1413,6 +1426,8 @@ pub mod syscall_nr {
     pub const MKDIR: u64 = 83;
     pub const MKDIRAT: u64 = 258;
     pub const RMDIR: u64 = 84;
+    pub const UNLINK: u64 = 87;
+    pub const UNLINKAT: u64 = 263;
     pub const POLL: u64 = 7;
     pub const SELECT: u64 = 23;
     pub const PSELECT6: u64 = 270;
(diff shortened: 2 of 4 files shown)


dburkart:
Parking this PR for human review.

**Diagnosis:** Branch is in `CONFLICTING` / `DIRTY` merge state — main has advanced since the last push and the branch needs a rebase. All CI checks green (fmt+build, host unit, integration, smoke, CodeRabbit pass, vfs-creds gate, soak eligibility) at head `d95567e`, but GitHub reports the branch cannot be merged cleanly against the current `main`.

Merge-watcher rules forbid touching the code (rebase), so handing off.

**Also FYI from CodeRabbit's latest review (non-blocking, actionable but nitpick-level):**
- `kernel/tests/syscall_unlink.rs` ~L225 — `unlink_removes_regular_file` passes `&mut sb` from the kernel stack to `stat`, which `stat_into_user` rejects as a bad user pointer; the post-unlink assertion is exercising pointer validation instead of the intended `-ENOENT` path. Consider staging the `Stat` buffer in the user VMA like sibling tests.
- Suggestion to also assert `SYS_UNLINKAT` is behind the `vfs_creds` dispatcher gate in `unlink_dispatcher_is_enosys`.
- Suggestion to do parent-dir W+X permission check in `vfs.rs` unlink/rmdir path rather than leaving it to backends.

CI run for reference: https://github.com/dburkart/vibix/actions/runs/24606521054

PR #595, opened 2026-04-18
userspace/init: pre/post-write diagnostic markers around first write(1) (#478)

### LINKED ISSUE #484: userspace init: add fine-grained diagnostic markers around first write syscall (#478 split 1/2)

dburkart:
Subsumes #530 (same diagnostic content). Closes #484.

## Summary

Brackets `init::_start`'s first `write(1, HELLO_MSG)` with two additional diagnostic markers, both routed through the existing direct-syscall `write()` helper (so they exercise the identical syscall path):

- `init: pre-write marker` — emitted on fd=2 as the very first userspace action, immediately after `iretq` into ring 3 and before any other write.
- `init: post-write marker` — emitted on fd=1 right after the first `write(1, HELLO_MSG)` returns.

Registered in `xtask::SMOKE_MARKERS` with a three-way marker-localisation table so a flaked-out #478 boot can be classified from the serial log alone:

| Present | Absent | What's broken |
|---|---|---|
| `ring3-iretq:` | `init: pre-write marker` | First SYSCALL / entry trampoline silently faulted |
| `init: pre-write marker` | `init: hello from pid 1` | fd=1 write path or state between writes |
| `init: hello from pid 1` | `init: post-write marker` | SYSRET / user-context restore |

No behavioral change beyond the two extra `write()` calls.

## Repro status (on this branch, post-rebase onto main @ `b16ec78`)

A local bounded 30x TCG soak (`timeout 900 cargo xtask smoke`) reproduced #478 at ~10% flake rate. The failing iterations terminate with:

```
init: entering ring-3 entry=0x400000 rsp=0x7fffff58
init: iretq to ring-3
-----------------------
Error: "smoke: missing markers [\"init: hello from pid 1\", \"init: post-write marker\", \"init: pre-write marker\"]"
```

Three signals pin this to the **entry-trampoline class**:

1. `ring3-iretq:` fires — kernel reached the 5-word IRETQ frame push cleanly.
2. No `init: pre-write marker` — userspace never successfully completed its first SYSCALL.
3. No `ring3-first-fault:` — the IDT's one-shot CPL=3 fault latch never tripped, so whatever is happening is not a ring-3 #PF / #GP / #UD (each of which runs `log_first_ring3_fault`).

The follow-up fix (root-causing what stalls between `iretq` and the first user SYSCALL without taking any CPU exception) continues on #478. This PR lands the classification vehicle so the next flake leaves more evidence.

## Test plan

- [x] `cargo xtask build` on this branch — succeeds.
- [x] `cargo xtask smoke` — passes on green iterations with all 35 markers; the two new markers fire immediately after `ring3-iretq:` on the happy path.
- [x] Local 30-iter TCG soak reproduces #478 with the marker set (~10% flake rate) and pins the classification.
- [ ] CI: fmt + build, host unit tests, integration tests (QEMU), smoke test.
- [ ] Post-merge: 3 consecutive all-green main CI runs (tracked on #478 — gating criterion for the root-cause fix, not this diagnostic PR).

## Test plan (boilerplate checklist)

- [ ] CI `fmt + build` green
- [ ] CI `host unit tests` green
- [ ] CI `integration tests (QEMU)` green
- [ ] CI `smoke test` green

### DIFF
--- a/userspace/init/src/main.rs
+++ b/userspace/init/src/main.rs
@@ -74,9 +74,33 @@ const HELLO_MSG: &[u8] = b"init: hello from pid 1\n";
 /// Emitted after the parent collects the child's exit status.
 const DONE_MSG: &[u8] = b"init: fork+exec+wait ok\n";
 
+/// Diagnostic marker for #478. Emitted on fd=2 as the very first userspace
+/// action — i.e. after `iretq` into ring 3 but before the first
+/// `write(1, HELLO_MSG)`. If the kernel-side `ring3-iretq:` marker fires
+/// but this one doesn't, the first SYSCALL instruction or the entry
+/// trampoline silently faulted before the handler ran. If this fires but
+/// `init: hello from pid 1` doesn't, the failure is specific to the fd=1
+/// write path, not ring-3 entry.
+const PRE_WRITE_MSG: &[u8] = b"init: pre-write marker\n";
+
+/// Diagnostic marker for #478. Emitted on fd=1 immediately after the
+/// first `write(1, HELLO_MSG)` returns. If `init: hello from pid 1` fires
+/// but this marker doesn't, the write syscall return path (SYSRET /
+/// restore of user context) is the culprit, not the write itself.
+const POST_WRITE_MSG: &[u8] = b"init: post-write marker\n";
+
 #[no_mangle]
 pub extern "C" fn _start() -> ! {
+    // Pre-write diagnostic marker — see #478. Emitted on fd=2 so it
+    // exercises the syscall path but writes to a distinct stream from
+    // HELLO_MSG, making it easy to grep from the serial log.
+    write(2, PRE_WRITE_MSG);
+
     write(1, HELLO_MSG);
+
+    // Post-write diagnostic marker — see #478. Emitted immediately after
+    // the first `write(1, HELLO_MSG)` returns, before any further work.
+    write(1, POST_WRITE_MSG);
 
     // fork() — child PID returned to parent; 0 returned to child.
     let fork_ret: i64;
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -106,7 +106,24 @@ const SMOKE_MARKERS: &[&str] = &[
     // `ring3-first-fault:` diagnostic line emitted by the IDT fault
     // handlers in that case).
     "ring3-iretq: rip=",
+    // #478 diagnostic: emitted by userspace init on fd=2 as the very first
+    // userspace action, immediately after ring-3 entry and before the first
+    // `write(1, HELLO_MSG)`. Three-marker localization:
+    //   - `ring3-iretq:` present, `pre-write marker` absent:
+    //     first SYSCALL instruction or the kernel syscall-entry trampoline
+    //     silently faulted before the Rust handler ran.
+    //   - `pre-write marker` present, `init: hello from pid 1` absent:
+    //     fd=1 write path (or state between the two writes) is the culprit,
+    //     not ring-3 entry.
+    //   - `init: hello from pid 1` present, `post-write marker` absent:
+    //     SYSRET / user-context restore is the culprit, not the write itself.
+    "init: pre-write marker",
     "init: hello from pid 1",
+    // #478 diagnostic: emitted on fd=1 immediately after the first
+    // `write(1, HELLO_MSG)` returns, before any further work (fork etc.).
+    // Missing this while `init: hello from pid 1` fires is a signal that the
+    // SYSRET / user-context restore after the first write is broken.
+    "init: post-write marker",
 ];
 
 /// Markers for the fork+exec+wait flow. These are flakey under un-accelerated


dburkart:
Auto-engineer status update — parking this PR for human review.

## Context

Assigned to stack a fix for #527 (fork CoW #PF at `rip=0x4002c6 cr2=0x7fffff0c`) onto this branch. Attempted a reproducer + investigation; status below.

## What was tried

1. **Reproduced the failure locally** via `cargo xtask smoke` on `m478-v3` HEAD (`ba6f809`). Local failure signature matches CI run `24608250754` exactly:

   ```
   init: iretq to ring-3
   -----------------------
   Error: "smoke: missing markers [...]"
   ```

   No `ring3-first-fault:`, no `EXCEPTION:` line, no `#PF` handler entry. Userspace never reaches the first SYSCALL.

2. **Symptom classification.** The failure on this PR's CI is the **#478 signature** (silent stall between `iretq` and first user SYSCALL — no exception fires), not the **#527 signature** (user write-fault at `cr2=0x7fffff0c rip=0x4002c6` which is handled by the existing CoW resolver and logs `ring3-first-fault:`). The task brief conflated the two, but the CI evidence shows the marker PR is making #478 reproducible, not #527.

3. **Code audit of the candidate root causes in #527's list:**
   - `fork_address_space` (kernel/src/mem/addrspace.rs:750) — correctly W-strips parent PTEs, bumps refcount, maps read-only in child, restores on error.
   - `cow_copy_and_remap` (kernel/src/mem/paging.rs:388) — allocates new frame, copies PTE-pointed source, releases old PTE ref.
   - `#PF` handler CoW branch (kernel/src/arch/x86_64/idt.rs:252) — takes `is_prot && is_write && pte_flags.WRITABLE`. The stack VMA from `init_process.rs:77-104` carries `prot_pte = PRESENT | WRITABLE | USER | NX`, so the branch is entered correctly.
   - `current_vma_lookup` uses `aspace.try_write()` which returns `None` on contention. Parent and child have **separate `Arc<RwLock<AddressSpace>>`** (created at `fork_current_task` line 218), so cross-task contention on one aspace is not possible in the single-CPU kernel.
   - None of these audits produced a concrete bug hypothesis I can confidently land a fix for.

## Why I'm parking instead of pushing a speculative fix

The CI failure mode on this PR is `iretq → silence` with no exception. Nothing in the #PF / CoW fault paths can explain a silent stall — those paths either resolve and return, or log+hang. The smoking gun from #527 (`ring3-first-fault: cr2=0x7fffff0c`) isn't appearing on this branch's CI; something earlier in the first user instruction sequence is stalling. Possible culprits that would need hardware-level QEMU tracing to discriminate:

- User instruction fetch at `0x400000` faults silently (TLB/PML4 inconsistency from the `flush.ignore()` paths in `map_in_pml4` when `Cr3 != pml4_frame`).
- IRETQ pushes consume a previously-unused frame but a preceding IRQ delivery clobbered it.
- A GDT / TSS / MSR config drift between the setup in `init::launch` and the ring-3 transition in `init_ring3_entry`.

Pushing a commit that guesses at one of these without confirmation would either be another band-aid (like #583, which was explicitly called out in the brief as the wrong approach) or code churn with no signal. Per the auto-engineer skill's "Parking the PR" contract, I'm stopping after a single fix round rather than burning review cycles on guesses.

## Next step for a human reviewer

Run `cargo xtask repro-fork --cycles 20000` with `-d int,cpu_reset` added to the QEMU cmdline in `xtask/src/main.rs`'s repro-fork harness to capture the faulting instruction trace around `rip=0x400000` at failure time. That output will pin #478 (and likely #527 along with it) to a concrete cause inside 20 minutes of trace inspection. Without that trace I don't have enough signal to land a defensible change.

## Unrelated observation

Integration-test CI is red on `main` since commit `cdf1d6e` (PR #610) with `ksymtab blob 262892 bytes exceeds reservation 262144`. Tracked as #615 with a parallel auto-engineer assigned. This PR's rebase onto post-#615 main is a prerequisite for CI-green merge regardless of the #527 question.

Labeling this PR `needs-human-review` and leaving the branch in place.

dburkart:
## Auto-engineer trace-capture update (WS-B, #478)

Rebased onto main (now including #621 wait4 timeout fix). Added `-d int,cpu_reset,guest_errors` + in_asm tracing to a direct QEMU loop (bypassing xtask scaffolding). Captured 4 failing traces + 1 passing baseline.

**Repro**: unchanged — `init: iretq to ring-3` + `ring3-iretq: rip=0x400000 rsp=0x7fffff58 cs=0x23 ss=0x1b rflags=0x202` prints, then serial goes silent.

## What the trace rules out

- **No #GP / #PF / #DF / #UD in the CPU trace** on failing runs (`grep -E "v=0[0-9a-f]"` → zero exception vectors after the two early-BIOS resets). So this is **not** an IRETQ fault, **not** a ring-3 instruction-fetch #PF, **not** a double-fault, **not** a triple-fault (no late `CPU Reset`).
- **CPL=3 count = 0** on failing runs, vs 1378+ on passing. So iretq never observably transfers to user mode, or transfers and is instantly re-entered by something the int-trace doesn't log.
- **debugcon shows `e0 e1`** (not `e0 e1 e0 e1`): the init task reaches `jump_to_ring3` asm, writes both markers, *then* iretq runs — there's nothing else after iretq before control is lost.

## What the trace points at

**Kernel is alive but timer-starved.** Over a 120 s failing run only 24 vector-0x20 (timer) ints fire on the failing path vs 5531 on the passing path. Adding `-d in_asm` shows the CPU executing a **spin loop in `vibix::process::current_pid`** for the entire tail of the run:

```
0xffffffff8005ef30:  pause
0xffffffff8005ef32:  movzbl (%rbx), %eax     ; %rbx = TABLE mutex lock byte
0xffffffff8005ef35:  test   %al, %al
0xffffffff8005ef37:  jne    0xffffffff8005ef30
```

That's the `spin::Mutex::lock()` spin for `process::TABLE` (at `0xffffffff802fe2d8`). It's entered with IF=1 (the preceding `sti` at 0x8005eef7 restores the saved flags after the `SCHED.lock()` / `IrqLock` critical section). So timer interrupts *should* fire and preempt the spinner, letting the holder release — but they apparently don't, or the holder cannot make progress.

Earlier timer samples in the same trace show the bootstrap task (SP = `0xffff80000ff52…`, kernel-main stack) **inside `process::register_init`** at pc=`0x8005f9cc`, about 30 s into the run, which is almost certainly where TABLE ends up held. Register_init is called exactly once from `init_process::launch`; if the bootstrap task is preempted between `TABLE.lock()` and the matching drop, any concurrent `current_pid()` caller spins, and we're observing exactly that spin.

## Hypothesis (unconfirmed)

This is a **single-CPU TABLE-lock contention / priority-inversion** between the init bootstrap path and one of the kernel tasks on the ready queue (shell / cursor_blink / the newly-spawned `init_ring3_entry`). The likely trigger is lock-ordering around `TABLE` + `SCHED`:
- `current_pid()` takes SCHED (IrqLock, IF=0) → releases → takes TABLE (spin, IF=1).
- Several existing callers (`sys_getsid`, `sys_getpgid`, `sys_setpgid`, etc.) take TABLE first, then call `task::current_id()` which takes SCHED. TABLE→SCHED order is *documented as forbidden* in `kernel/src/process/mod.rs:6-11`, but is still present in multiple call sites.
- The diagnostic markers on `m478-v3` tighten the boot timing just enough to expose a window where TABLE can remain held across the first `current_pid()` contention.

However, the actual *holder* at the hang point could not be confirmed from the int-trace alone — the spinner is the only task the trace samples once the kernel enters the stuck state.

## Artifacts (not pushed — too large for the repo)

Saved locally under `target/repro-478/`:
- `serial-{5,11,13}.log`, `debugcon-{5,11,13}.log` — 3 failing captures.
- `trace-{5,11,13}.log` (2.6–2.8 MB each) — `-d int,cpu_reset,guest_errors`.
- `target/repro-478-extended/fail-trace.log` — 120 s run with `+in_asm`, which is what pinpointed the current_pid spin.

## Recommendation

**Parking PR #595 again.** The diagnostic markers are doing their job — they make #478 deterministic. The trace strongly suggests a `TABLE` spinlock issue under boot-window contention, but pinning the exact holder needs either:
1. A debug build of `spin::Mutex` that logs the owner task id on each acquire (so the trace at the hang point shows who's sitting on it), or
2. A GDB-stub session (`-s -S`) so a human can step through the init sequence and snapshot all task kernel stacks the moment `current_pid` starts spinning.

Either is appropriate for a human expert; I don't think an auto-engineer should speculate a fix here (per PR #583's lesson).

State of the linked issues:
- **#478**: trace captured, hypothesis narrowed to a TABLE-lock / lock-ordering issue (not ring-3 entry). Not fixed.
- **#484**: subsumed by #478 (same signature). Not fixed.
- **#527**: likely a separate signature (ring-3 #PF on fork CoW at cr2=0x7fffff0c), but could also be a secondary effect of the same lock issue if the CoW path takes TABLE/SCHED in an incompatible order. Left open; no follow-up filed because #527 is not reproducing on the `m478-v3` local loop (only #478 is).

Issue #548, opened 2026-04-18
sys: wire setuid/setreuid/setresuid + setgid/setregid/setresgid (POSIX saved-set-uid)


dburkart:
## Motivation

Implements RFC 0004 — Workstream B (Credential enforcement, wave 1). Identity-change
syscalls. Must correctly implement POSIX §2.4 saved-set-uid semantics so a setuid
binary can drop privilege and optionally restore it.

## Work

- [ ] Wire `setuid(uid)` — if euid == 0, set ruid/euid/suid all to `uid`; else require `uid == ruid || uid == suid` and set euid to `uid`
- [ ] Wire `setreuid(ruid, euid)` — `-1` leaves field unchanged; non-root requires each target to be one of the current {ruid, euid, suid}; update suid to new euid when euid changes or ruid was changed
- [ ] Wire `setresuid(ruid, euid, suid)` — `-1` leaves field unchanged; non-root requires each target ∈ current {ruid, euid, suid}
- [ ] Mirror all three for gid: `setgid` / `setregid` / `setresgid`
- [ ] **Supplementary groups NOT cleared on uid/gid transitions** — follow Linux rule, documented in RFC 0004
- [ ] Atomic update: write the new `Arc<Credential>` under the task's `BlockingRwLock`; readers continue to see the old snapshot until they re-read
- [ ] Errno coverage: `EAGAIN` (if we cap counts), `EINVAL`, `EPERM` per RFC errno table
- [ ] Integration test matrix:
  - root setuid drops all three to non-zero (verify cannot regain)
  - setuid binary's drop-and-restore dance via setreuid(ruid, suid)
  - setresuid with -1 preserves fields
  - setregid(-1, new_egid) transitions egid only

## Out of scope

- `setfsuid` / `setfsgid` (Linux-only filesystem identity) — deferred.
- `CAP_SETUID` / `CAP_SETGID` capabilities — explicit non-goal.

## Context

- Parent epic: #537
- Depends on: #546 (Credential struct)
- POSIX.1-2017 §2.4.4, §setuid/setgid, §setresuid
- RFC 0004 §Kernel-Userspace Interface (errno table)

PR #596, opened 2026-04-18
kernel/syscall: wire chmod/chown family to InodeOps::setattr (#541)

### LINKED ISSUE #541: sys: wire rename/renameat/renameat2(RENAME_NOREPLACE)
## Motivation

Implements RFC 0004 — Workstream A (POSIX write syscalls, wave 1). `rename` is the
atomic-swap primitive that higher-level tools (editors, package managers, rsync) depend
on. Without it, userspace can't perform atomic file replace, and the ext2 write path
can't be tested through its rename code path.

## Work

- [ ] Wire `rename(oldpath, newpath)` + `renameat(olddirfd, oldpath, newdirfd, newpath)`
- [ ] Wire `renameat2(..., flags)` supporting `RENAME_NOREPLACE` only (no `EXCHANGE`, no `WHITEOUT` in this epic)
- [ ] Cross-directory rename: lock parent dirs in ascending inode order (see RFC §Concurrency)
- [ ] Sticky-bit check on source parent (`S_ISVTX`) — symmetric with unlink
- [ ] Errno matrix per RFC 0004 §Kernel-Userspace Interface: `EACCES`, `EBUSY`, `EDQUOT`, `EEXIST` (w/ NOREPLACE), `EINVAL`, `EISDIR`, `ELOOP`, `EMLINK`, `ENOTDIR`, `ENOTEMPTY`, `EPERM`, `EROFS`, `EXDEV`
- [ ] Reject rename across mount points (`EXDEV`)
- [ ] Integration test: same-dir rename, cross-dir rename, `RENAME_NOREPLACE` conflict, sticky-bit block

## Out of scope

- Actual ext2 rename implementation with POSIX atomicity + link-count ordering (Workstream E).
- `RENAME_EXCHANGE`, `RENAME_WHITEOUT` — deferred (Open Question in RFC).

## Context

- Parent epic: #537
- RFC 0004 §Concurrency and locking, §Kernel-Userspace Interface errno table
- Ref: `kernel/src/fs/vfs/ramfs.rs` — reference rename op

dburkart:
## Summary

Wires the seven POSIX metadata-mutation syscalls that were stubbed to `-ENOSYS` while per-task credentials (#546) landed:

- `chmod(2)` = 90 / `fchmod(2)` = 91 / `fchmodat(2)` = 268
- `chown(2)` = 92 / `fchown(2)` = 93 / `lchown(2)` = 94 / `fchownat(2)` = 260

All seven route through one inner helper per op (`do_chmod` / `do_chown`), read the per-task `Credential` snapshot once, build a minimal `SetAttr`, and call `InodeOps::setattr`. Dispatch arms are `#[cfg(feature = \"vfs_creds\")]`-gated to preserve the A↔B gate; impl functions compile unconditionally so every branch type-checks.

Closes #541. Unblocker: #546. Workstream A of RFC 0004 ext2 epic #537.

## POSIX semantics worth reviewer attention

- **chmod**: allowed iff `euid == 0` or `euid == st_uid`; else `-EPERM`.
- **chown uid**: root-only. Non-root `-EPERM` the moment a non-sentinel uid is supplied.
- **chown gid**: root any. Owner may set to any supplementary group they currently belong to. Non-owner non-root always `-EPERM`.
- **`-1` sentinel**: `uid == u32::MAX` / `gid == u32::MAX` means "don't change". `fchown(fd, -1, -1)` is a permitted no-op returning 0 even for non-owners (matches Linux).
- **setuid-bit clearing on non-root chown of regular files**: `S_ISUID` always cleared; `S_ISGID` only cleared when `S_IXGRP` is also set — preserves the historic mandatory-lock marker per SUSv4.
- **fchmodat/fchownat flags**: `AT_SYMLINK_NOFOLLOW` is the only recognised bit; others `-EINVAL`. `lchown` == `fchownat(AT_FDCWD, path, uid, gid, AT_SYMLINK_NOFOLLOW)`.
- **Path walks** run under the caller's credential (new `resolve_inode_as` helper), not root, so search-denied components on the way to a writable leaf return `-EACCES`. Legacy `resolve_inode` keeps root semantics for read-only syscalls that predate the cred work.

Also adds `task::set_current_credentials` as the single write-side entry for the upcoming setuid/setgid family (#548) and the tests below. Builds a fresh `Arc<Credential>` rather than mutating in place so in-flight read-side snapshots stay consistent.

## Test plan

- [x] `cargo xtask test-integration` — all 17 new `syscall_chmod_chown` cases pass; full suite green except one cold-cache `wait4_condvar_race` flake that rerun cleanly.
- [x] `cargo xtask test-unit` — 409 host unit tests pass.
- [x] `cargo xtask smoke` — all 33 markers present (1 cold-cache miss on first run; clean on rerun, matches documented flake pattern).
- [x] `cargo xtask build` — clean build with rebased main.
- [ ] CI: wait for full gauntlet (build / unit / integration / smoke / soak) before merge.

Test coverage (`kernel/tests/syscall_chmod_chown.rs`):

- dispatcher gate: ENOSYS without `vfs_creds`, non-ENOSYS with.
- chmod by owner / by root succeed; by non-owner `-EPERM`.
- fchmod / fchmodat mirror chmod; fchmod on non-VFS fd `-EBADF`; fchmodat with `AT_FDCWD` resolves relative paths.
- fchmodat/fchownat reject unknown flag bits with `-EINVAL`.
- chown uid change by non-root `-EPERM`; by root succeeds.
- chown gid change by owner to a supplementary group succeeds; to a non-member group `-EPERM`.
- `fchown(fd, -1, -1)` no-op success as non-owner.
- setuid-bit clearing: S_ISUID always cleared, S_ISGID only cleared when S_IXGRP also set (verified on two chmods).
- lchown applies `AT_SYMLINK_NOFOLLOW` (regular-file case; symlink case deferred to the symlink syscall PR).

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -299,6 +299,10 @@ name = "syscall_getuid_family"
 harness = false
 
 [[test]]
+name = "syscall_chmod_chown"
+harness = false
+
+[[test]]
 name = "ntty_sigint_flush"
 harness = false
 
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -831,6 +831,41 @@ pub unsafe extern "C" fn syscall_dispatch(
         #[cfg(feature = "vfs_creds")]
         UNLINKAT => super::syscalls::vfs::sys_unlinkat_impl(a0 as i32, a1, a2 as u32),
 
+        // chmod / chown family (issue #541, RFC 0004 §Permission model).
+        // Every arm is gated behind `vfs_creds` — these syscalls make DAC
+        // decisions against the caller's credentials, so until Workstream
+        // B flips the feature on they must remain unreachable from ring-3.
+        // Integration tests call `sys_*_impl` directly, mirroring the
+        // mkdir/unlink convention.
+
+        // chmod(path, mode) — change the mode bits of `path`.
+        #[cfg(feature = "vfs_creds")]
+        CHMOD => super::syscalls::vfs::sys_chmod_impl(a0, a1),
+
+        // fchmod(fd, mode) — change the mode bits of an open fd's file.
+        #[cfg(feature = "vfs_creds")]
+        FCHMOD => super::syscalls::vfs::sys_fchmod_impl(a0, a1),
+
+        // fchmodat(dfd, path, mode, flags) — `*at` form of chmod.
+        #[cfg(feature = "vfs_creds")]
+        FCHMODAT => super::syscalls::vfs::sys_fchmodat_impl(a0 as i32, a1, a2, a3 as u32),
+
+        // chown(path, uid, gid) — change owner/group of `path`.
+        #[cfg(feature = "vfs_creds")]
+        CHOWN => super::syscalls::vfs::sys_chown_impl(a0, a1, a2),
+
+        // fchown(fd, uid, gid) — change owner/group of an open fd's file.
+        #[cfg(feature = "vfs_creds")]
+        FCHOWN => super::syscalls::vfs::sys_fchown_impl(a0, a1, a2),
+
+        // lchown(path, uid, gid) — change owner/group, stop on a symlink.
+        #[cfg(feature = "vfs_creds")]
+        LCHOWN => super::syscalls::vfs::sys_lchown_impl(a0, a1, a2),
+
+        // fchownat(dfd, path, uid, gid, flags) — `*at` form of chown.
+        #[cfg(feature = "vfs_creds")]
+        FCHOWNAT => super::syscalls::vfs::sys_fchownat_impl(a0 as i32, a1, a2, a3, a4 as u32),
+
         // poll(fds, nfds, timeout_ms) — wait for readiness on a set of fds.
         POLL => crate::poll::syscalls::sys_poll(a0, a1, a2 as i64),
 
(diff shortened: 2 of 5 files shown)


dburkart:
Parking this PR — `cargo fmt --check` is failing on the first CI run.

Failed job: https://github.com/dburkart/vibix/actions/runs/24607725250/job/71956673371

Fmt wants reflows in:
- `kernel/src/arch/x86_64/syscalls/vfs.rs` (around lines 1230 and 1401)
- `kernel/tests/syscall_chmod_chown.rs` (around lines 88, 302, 344, 472, 528, 551)

This is a mechanical `cargo fmt --all` fix, but it's outside the merge-watcher's scope (the watcher is spec'd not to fix code). No known flake attribution applies — rerunning would not help.

Other checks so far: RFC 0004 vfs_creds gate PASS, soak-eligibility PASS, host unit tests PASS, 100× smoke SKIPPED (as expected). `integration tests (QEMU)` and `smoke test` are still in progress; CodeRabbit pending.

Next step for a human / respawned auto-engineer: run `cargo fmt --all` locally, push, and let CI re-run.

Applying `needs-human-review` and exiting.

dburkart:
Applied `cargo fmt --all`; rekicking CI.

Issue #570, opened 2026-04-18
fs/ext2: InodeOps link and symlink (inline fast-symlink + slow-symlink paths)


dburkart:
## Motivation

Implements RFC 0004 — ext2 filesystem driver (workstream E, write-path).

Hard links bump `i_links_count` and add a new dirent pointing at the same inode; symbolic links are a distinct inode whose target path is stored either inline in `i_block[]` (fast symlink, ≤ 60 bytes) or in a single data block (slow symlink). Both paths must share the create/add_link ordering from #568.

## Work

- [ ] `Ext2InodeOps::link(old_dentry, parent, new_name, cred)`:
  - Permission-check parent with `MAY_WRITE | MAY_EXEC`.
  - Refuse if `old_dentry` is a directory (POSIX: only superuser historically; vibix rule: always refuse).
  - Refuse if `i_links_count == u16::MAX`.
  - Increment `i_links_count`, sync inode, add dirent in parent, sync parent block. On add_link failure, decrement `i_links_count` and revert.
- [ ] `Ext2InodeOps::symlink(parent, new_name, target, cred)`:
  - Follows the create ordering in #568, but:
    - If `target.len() <= 60`: store in `i_block[0..=14]` as raw bytes, `i_size = target.len()`, `i_blocks = 0` (fast symlink).
    - Else: allocate a data block, write target into it, `i_size = target.len()`, `i_blocks` updated (slow symlink).
  - `i_mode` = `S_IFLNK | 0o777` (symlinks are always 0777; POSIX clause §4.10).
- [ ] Tests: link cycle detection (refuse ELOOP up-front? — VFS already handles; ext2 does not re-check); link past u16::MAX; symlink at exactly 60 bytes (inline) and 61 bytes (slow); symlink read-back matches.

## Out of scope

- The `readlink` op itself is the read-path counterpart already in #563; this issue only covers writing the target.

## Context

- Parent epic: #537
- RFC: `docs/RFC/0004-ext2-filesystem-driver.md` (§Hard links, §Symbolic links, §Fast symlink format)
- Depends on: create/mkdir/mknod (this epic), dirent walker (#560).
- Blocks: rename (this epic) indirectly — rename across directories uses the same add_link helper.

PR #605, opened 2026-04-19
sys: wire link/linkat + symlink/symlinkat + readlink/readlinkat (#540)

### LINKED ISSUE #540: sys: wire link/linkat + symlink/symlinkat + readlink/readlinkat
## Motivation

Implements RFC 0004 — Workstream A (POSIX write syscalls, wave 1). The VFS
`InodeOps::link`/`symlink`/`readlink` hooks exist but there is no syscall arm calling
them. Required for hard-link testing and for symlink resolution during path walk.

## Work

- [ ] Wire `link(oldpath, newpath)` + `linkat(olddirfd, oldpath, newdirfd, newpath, flags)` — support `AT_EMPTY_PATH`, `AT_SYMLINK_FOLLOW`
- [ ] Wire `symlink(target, linkpath)` + `symlinkat(target, newdirfd, linkpath)` — `target` stored as bytes, not NUL-terminated
- [ ] Wire `readlink(path, buf, bufsize)` + `readlinkat(dirfd, path, buf, bufsize)` — **do NOT NUL-terminate the output** (POSIX returns bytes written, caller adds NUL)
- [ ] `link` rejects linking across mount points (`EXDEV`) and rejects directory source (`EPERM`)
- [ ] `link` enforces `LINK_MAX` (ext2 = 32000) — returns `EMLINK`
- [ ] Errno coverage: `EACCES`, `EEXIST`, `EFAULT`, `EIO`, `ENOENT`, `EXDEV`, `EMLINK`, `ENOSPC`, `EPERM` per RFC errno table
- [ ] Integration test: create symlink, `readlink` it back, `link` a file and verify hard-link count via `stat`

## Out of scope

- Fast-symlink on-disk storage for ext2 (Workstream E).
- Symlink loop detection limits — covered by existing path-walk code.

## Context

- Parent epic: #537
- RFC 0004 §Kernel-Userspace Interface for errno table
- `kernel/src/fs/vfs/` for trait signatures

dburkart:
## Summary

Closes #540. Part of RFC 0004 epic #537, Workstream A (POSIX write
syscalls, wave 1). Mirrors the syscall-wiring structure of #585
(mkdir), #589 (unlink), and #596 (chmod/chown).

- Wires `link`/`linkat`, `symlink`/`symlinkat`, `readlink`/`readlinkat`
  through the existing `InodeOps::link`/`symlink`/`readlink` trait
  methods. Dispatch arms are gated behind `#[cfg(feature =
  \"vfs_creds\")]` per the A-before-B convention; `sys_*_impl`
  entry points are always compiled so integration tests exercise
  them regardless of feature state.
- Shaping done in the shared bodies: `EPERM` on directory-hard-link,
  `EXDEV` across superblocks, `EEXIST` / `ENOTDIR` / `ENOENT` /
  `EINVAL` per RFC 0004 §Kernel-Userspace Interface. `readlink`
  is **not** NUL-terminated (POSIX); walk uses default `LookupFlags`
  (not `NOFOLLOW`) so a terminal symlink resolves to itself instead
  of returning `ELOOP`.
- Adds `EMLINK = -31` errno for the `link` nlink-overflow table
  entry. Actual enforcement lives in ext2's `link` body (#570,
  separate workstream).

## Test plan

- [x] Host unit tests pass (`cargo xtask test-unit`), including the
      syscall-number / errno ABI regression tables with the new
      entries.
- [x] Integration test `syscall_link_symlink` — 13 cases, all pass:
      dispatcher-ENOSYS anchors for SYS_link/symlink/readlink,
      symlink→readlink round-trip, no-NUL-terminator check,
      bufsize truncation, `EINVAL` on non-link / bufsize-0,
      `ENOENT` on missing path, `link` nlink increment + shared-ino
      verification, `EPERM` on directory source, `symlinkat` and
      `readlinkat` with `AT_FDCWD` + absolute paths.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -305,6 +305,10 @@ name = "syscall_unlink"
 harness = false
 
 [[test]]
+name = "syscall_link_symlink"
+harness = false
+
+[[test]]
 name = "syscall_getuid_family"
 harness = false
 
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -853,6 +853,41 @@ pub unsafe extern "C" fn syscall_dispatch(
         #[cfg(feature = "vfs_creds")]
         UNLINKAT => super::syscalls::vfs::sys_unlinkat_impl(a0 as i32, a1, a2 as u32),
 
+        // link/linkat/symlink/symlinkat/readlink/readlinkat (issue #540,
+        // RFC 0004 Workstream A wave 1). Same A-before-B gate as the
+        // other write syscalls — with `vfs_creds` off the arms are
+        // absent and the dispatcher's default returns `-ENOSYS`.
+        // Integration tests call `sys_*_impl` directly regardless of
+        // feature state, mirroring the mkdir/unlink convention.
+
+        // link(oldpath, newpath) — create a hard link at `newpath`.
+        #[cfg(feature = "vfs_creds")]
+        LINK => super::syscalls::vfs::sys_link_impl(a0, a1),
+
+        // linkat(olddfd, oldpath, newdfd, newpath, flags) — *at form
+        // of link; honors AT_SYMLINK_FOLLOW.
+        #[cfg(feature = "vfs_creds")]
+        LINKAT => super::syscalls::vfs::sys_linkat_impl(a0 as i32, a1, a2 as i32, a3, a4 as u32),
+
+        // symlink(target, linkpath) — create a symbolic link at
+        // `linkpath` whose contents are `target`.
+        #[cfg(feature = "vfs_creds")]
+        SYMLINK => super::syscalls::vfs::sys_symlink_impl(a0, a1),
+
+        // symlinkat(target, newdfd, linkpath) — *at form of symlink.
+        #[cfg(feature = "vfs_creds")]
+        SYMLINKAT => super::syscalls::vfs::sys_symlinkat_impl(a0, a1 as i32, a2),
+
+        // readlink(path, buf, bufsize) — read the target of a symlink.
+        // Output is NOT NUL-terminated; the return value is the byte
+        // count written (POSIX).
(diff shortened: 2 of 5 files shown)


dburkart:
Thanks CodeRabbit — addressing the three findings:

**1. Default `link()` on terminal symlink returning ELOOP (critical) — fixed.**
Commit fef847b: `linkat_impl` now walks with `LookupFlags::default()` when `AT_SYMLINK_FOLLOW` is clear, so the terminal symlink inode is returned as-is (same trick `readlinkat_impl` already uses). Added regression test `link_on_symlink_default_hard_links_the_symlink` that links a symlink with a deliberately-nonexistent target — proves the walk never dereferences.

**2. DAC checks use `Credential::kernel()` instead of caller creds — out of scope for #540.**
This matches the established convention in the four sibling Workstream-A wave-1 PRs that already landed (mkdir #585 line 853, unlink #589, chmod/chown #596 line 1092). The dispatch arms are gated behind `#[cfg(feature = "vfs_creds")]` and the feature is not default-on (enforced by the `RFC 0004 vfs_creds gate` CI check), so `Credential::kernel()` is never reached from ring-3 today. Plumbing per-task creds through every `sys_*_impl` path is Workstream B's job (#546) and will land atomically with the gate flip, per RFC 0004. Filing a follow-up separately if one doesn't exist — doing it piecemeal per syscall would defeat the A-before-B design.

**3. Test VMAs missing `USER_ACCESSIBLE` — out of scope nit.**
All nine existing `kernel/tests/syscall_*.rs` integration tests use the exact same `prot_pte = PRESENT | WRITABLE | NO_EXECUTE` without `USER_ACCESSIBLE` (mkdir, chmod_chown, dup_family, fcntl, getdents, lseek, mkfifo, open_flags, open_mmap). The pattern works because the test-runner's paging setup grants user access via a different mechanism. Changing it in this PR alone would diverge from the convention — if the finding is real, it wants a sweep PR across all nine sibling tests.

Issue #564, opened 2026-04-18
fs/ext2: on-mount orphan-chain validation (bounded walk, cycle detection)


dburkart:
## Motivation

Implements RFC 0004 — ext2 filesystem driver (workstream D, read-path).

ext2's on-disk orphan list (`s_last_orphan` head threaded via `i_dtime` per inode) can be corrupted or point at reserved inodes. On mount the driver must walk the chain defensively — bounded iteration, cycle detection, reserved-ino rejection — and either drain (RW mount) or log (RO mount) surviving entries. An unchecked walk is a trivial DoS / EIO vector on a malicious image.

## Work

- [ ] Implement `Ext2SuperBlock::replay_orphan_list(&self, rw: bool) -> Result<(), Errno>`, called from `SuperOps::mount` after the root inode is loaded.
- [ ] Maintain a `BitVec` indexed by inode number to detect cycles; refuse any `ino < EXT2_GOOD_OLD_FIRST_INO` except `EXT2_ROOT_INO` (2) and reject any `ino >= s_inodes_count`.
- [ ] Bound the walk at `s_inodes_count` iterations — anything longer is a cycle or corruption → force read-only, log, return.
- [ ] On RW mount: for each surviving entry, truncate-to-zero (if `i_links_count == 0`) and free the inode + blocks via the allocator; remove from `s_last_orphan` chain.
- [ ] On RO mount: log each surviving entry with its i_mode/i_size, leave the chain untouched.
- [ ] Tests: craft a small image with (a) a valid 3-entry orphan chain, (b) a chain with a cycle, (c) a chain pointing at ino 0, (d) chain length > `s_inodes_count`. Verify correct drain, cycle detection, rejection, and forced-RO behaviour.

## Out of scope

- The orphan-add / orphan-remove sides used during unlink (covered in the unlink/rmdir issue).
- Kernel task or timer that periodically sweeps stale entries at runtime (mount-time is sufficient).

## Context

- Parent epic: #537
- RFC: `docs/RFC/0004-ext2-filesystem-driver.md` (§Orphan List on-disk + §Mount-time recovery)
- Depends on: inode read path (#561), block layout walker (#559).
- Blocks: all write-path work in Workstream E (unlink/rmdir reuse the orphan-add/remove helpers but rely on recovery for consistency).

Issue #560, opened 2026-04-18
fs/ext2: indirect-block walker with bounds-checked pointers + per-inode cache


dburkart:
## Motivation

Implements RFC 0004 — Workstream D (ext2 read path, wave 2). ext2's data-block
addressing uses 12 direct + 1 single-indirect + 1 double-indirect + 1 triple-indirect
pointers in `i_block[15]`. Resolving a logical block index to an absolute LBA has to
walk the correct chain. Every pointer read off disk is attacker-controlled if the
image is hostile, so every pointer MUST be bounds-checked.

## Work

- [ ] `resolve_block(inode: &Ext2Inode, logical_block: u32) -> Result<Option<u64>>`
      returning `Ok(None)` for a sparse hole and `Ok(Some(absolute_block))` otherwise
- [ ] Four paths: direct (< 12), single (< 12 + N), double (< 12 + N + N²), triple (else);
      `N = block_size / 4`
- [ ] Each indirect block read goes through the buffer cache (`bread`)
- [ ] **Bounds check every pointer**: pointer value ∈ `[s_first_data_block, s_blocks_count)` OR zero
      (sparse). Zero is the hole sentinel. Any pointer into the metadata-forbidden range
      (superblock sector, BGDT blocks, inode bitmap / block bitmap / inode-table blocks
      of any group) is rejected with `EIO` and the FS is forced RO.
- [ ] **Per-inode walk cache**: small (4–8 entries) recent (logical_block → absolute)
      cache on the inode, invalidated by an epoch stamp that bumps on truncate / write-extend
- [ ] Unit tests: direct, single-indirect, double-indirect, triple-indirect resolution
      against a fixture image with known layouts; sparse hole returns None; out-of-range
      pointer returns EIO and sets the mount to RO

## Out of scope

- Allocating blocks (Workstream E).
- Extents (ext4 feature) — not applicable.

## Context

- Parent epic: #537
- Depends on: #559 (Ext2Inode), Workstream C (buffer cache)
- RFC 0004 §Indirect-block walker, §Security (bounds checking)

Issue #559, opened 2026-04-18
fs/ext2: Ext2Inode + iget + inode cache + orphan_list strong-ref mirror


dburkart:
## Motivation

Implements RFC 0004 — Workstream D (ext2 read path, wave 2). iget loads an inode from
the inode table and returns a VFS `Arc<Inode>`. A cache on top (`Weak<Inode>` keyed by
ino) avoids reloading on every path-walk step. The `orphan_list` mirrors the on-disk
`s_last_orphan` chain with `Arc<Inode>` strong refs so an unlinked-but-open inode
can't be evicted before its last close.

## Work

- [ ] `struct Ext2Inode { fs: Weak<Ext2Fs>, ino: u32, meta: BlockingRwLock<Ext2InodeMeta>, block_map: BlockingRwLock<Option<BlockMap>>, unlinked: AtomicBool }`
- [ ] `Ext2InodeMeta` carries the decoded inode fields (uid, gid, mode, size, timestamps, `i_blocks` — in 512-byte units, `i_links_count`, raw `i_block[15]`)
- [ ] `impl InodeOps::getattr` — populate `stat` struct from `Ext2InodeMeta`
- [ ] `impl InodeOps::lookup` — directory-entry search for a single name (cooperates with Workstream D getdents64 issue)
- [ ] `iget(fs, ino)` — lookup by ino: inode-table block = `bg_inode_table + (ino_in_group * inode_size) / block_size`; offset within block = `(ino_in_group * inode_size) % block_size`
- [ ] `inode_cache: Mutex<BTreeMap<u32, Weak<Inode>>>` — hand out the same `Arc<Inode>` if it's still live
- [ ] **orphan_list**: `Mutex<BTreeMap<u32, Arc<Inode>>>` held on `Ext2Fs`, strong refs so the inode is not dropped while unlinked+open
- [ ] `Inode::drop` vetoes eviction when `unlinked == true` — the orphan path is the only path that sets it, and the orphan_list's strong ref keeps it alive
- [ ] Integration test: iget root (ino=2), iget a regular file, verify cache returns the same Arc

## Out of scope

- Indirect-block walker (next D issue).
- Directory iteration / lookup (D6).
- Orphan-chain replay on mount (D8).
- Adding/removing from orphan list (Workstream E's unlink path).

## Context

- Parent epic: #537
- Depends on: #557 (disk types), #558 (Ext2Fs/mount)
- RFC 0004 §Key data structures (Ext2Inode), §Orphan list

[...]

Issue #558, opened 2026-04-18
fs/ext2: Ext2Fs implements FileSystem + SuperOps; mount with feature-flag gate


dburkart:
## Motivation

Implements RFC 0004 — Workstream D (ext2 read path, wave 2). The superblock read and
mount-time validation is the gate between "this is an ext2 disk we understand" and
"abort or force-RO". Must exactly match ext2's feature-flag semantics so we don't
mount and silently corrupt a filesystem that uses features we don't implement.

## Work

- [ ] `kernel/src/fs/ext2/mod.rs` — `struct Ext2Fs` matching RFC 0004 §Key data structures
      (device, cache, sb, bgdt, inode_cache, orphan_list, mount_flags)
- [ ] `impl FileSystem for Ext2Fs` — driver registration + `mount(&dev, &flags)` entry
- [ ] `impl SuperOps for Ext2Fs` — `statfs`, `sync_fs`, `unmount`, etc.
- [ ] Mount sequence:
  1. `bread` superblock sector (disk offset 1024), validate `s_magic == 0xEF53`
  2. Read block group descriptor table
  3. Validate feature flags:
     - unknown `INCOMPAT_*` bit set → reject with `EINVAL`
     - unknown `RO_COMPAT_*` bit set → force `MS_RDONLY`
  4. If mount is RW: synchronously write `s_state := ERROR_FS` before returning (so a crash leaves a marker; `s_state` is cleared on clean umount by Workstream E)
  5. If mount is RO: **never** write the superblock
- [ ] `mount_flags` on the `Ext2Fs`: track `MS_RDONLY`, `MS_NOSUID`, `MS_NODEV`, `MS_NOATIME`
- [ ] Mount flag `MS_NOSUID` enforced by the execve gate + by stripping SUID bits on
      files being opened (that's the existing VFS's job; document assumption)
- [ ] Integration test against a host-produced `mkfs.ext2` fixture image: mount RO,
      mount RW (observing `s_state = ERROR_FS`), mount with bogus INCOMPAT (rejected),
      mount with bogus RO_COMPAT (forced RO)

## Out of scope

- umount's `s_state := clean` write (Workstream E).
- mount(2) syscall entry (Workstream F) — this issue only implements the driver;
  the boot path calls it directly for now.

## Context

- Parent epic: #537
- Depends on: #557 (disk types), Workstream C (buffer cache)
- RFC 0004 §Mount, §Feature-flag gate

PR #606, opened 2026-04-19
fs/ext2: Ext2Inode + iget + inode cache + orphan_list (#559)

dburkart:
## Summary

Lands Workstream D wave 2 of RFC 0004 (issue #559):

- **`Ext2Inode`** — driver-level inode with `meta: BlockingRwLock<Ext2InodeMeta>`, `block_map` slot for #560's walker, and an `unlinked: AtomicBool` the orphan-list path will flip.
- **`iget(super, sb, ino)`** — per-group + block-in-table arithmetic, reads the inode-table slot through the per-mount `BlockCache`, builds the VFS `Arc<Inode>`, publishes `Weak<Inode>` in the cache, races are winner-take-all.
- **`inode_cache: Mutex<BTreeMap<u32, Weak<Inode>>>`** — lookup shortcut, not an ownership root. Same `Arc<Inode>` returned while at least one live.
- **`orphan_list: Mutex<BTreeMap<u32, Arc<Inode>>>`** — strong refs that pin unlinked-but-open inodes per RFC 0004 §Orphan-list residency invariant. Constructed empty; populated by Workstream E's unlink path and mount-time replay in #564.
- **`InodeOps::getattr`** populated from the decoded meta; `InodeOps::lookup` stubbed to `ENOENT` (the dir-entry walker is the next issue).
- **`sb.root` wired at mount** via `iget_root(EXT2_ROOT_INO)`; mount failure there releases the single-mount latch.
- **`Ext2Fs::current_super()`** returns `Option<Arc<Ext2Super>>` so tests and `/proc/mounts` can reach the concrete per-mount type without a dyn-Any downcast.

Primary file is `kernel/src/fs/ext2/inode.rs` (new); `fs.rs` grew two fields on `Ext2Super` plus the root-inode wiring. No changes outside the ext2 module aside from a new `[[test]]` entry.

## Test plan

- [x] Host unit tests (`cargo test -p vibix --lib --features ext2`) — 432 pass, includes the new pure-function tests on `inode_kind_from_mode` / `Ext2InodeMeta::from_disk` / `vfs_meta_from`
- [x] New QEMU integration test `kernel/tests/ext2_inode_iget.rs` — 5 tests: getattr on root (mode/uid/size/blocks/nlink), iget of lost+found (non-root arithmetic), cache identity under live Arcs, EINVAL on `ino == 0` / `ino > s_inodes_count`, empty orphan list on fresh mount
- [x] `kernel/tests/ext2_mount.rs` still green (no regression to #558)

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -352,3 +352,8 @@ required-features = ["ext2"]
 [[test]]
 name = "block_writeback"
 harness = false
+
+[[test]]
+name = "ext2_inode_iget"
+harness = false
+required-features = ["ext2"]
--- a/kernel/src/fs/ext2/fs.rs
+++ b/kernel/src/fs/ext2/fs.rs
@@ -152,6 +152,12 @@ pub struct Ext2Fs {
     /// Self-reference so `Ext2Super::unmount` can upgrade it and clear
     /// the latch. `Arc::new_cyclic` fills this in at construction.
     self_ref: Weak<Ext2Fs>,
+    /// Weak pointer to the currently-mounted `Ext2Super`, set on a
+    /// successful `mount` and cleared on `unmount`. Exposed via
+    /// [`Ext2Fs::current_super`] so integration tests (and the future
+    /// `/proc/mounts` consumer) can reach the concrete per-mount type
+    /// without a dyn-Any downcast on `Arc<dyn SuperOps>`.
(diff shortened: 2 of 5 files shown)


dburkart:
smoke keeps tripping #478; reran. Will wait for indirect flake to clear.

Issue #563, opened 2026-04-18
fs/ext2: fast symlink read (inline) + slow symlink read


dburkart:
## Motivation

Implements RFC 0004 — Workstream D (ext2 read path, wave 2). ext2 stores symlinks
whose target is ≤ 60 bytes inline in `i_block[]` (the "fast" symlink); longer targets
are stored in the first data block. Both paths need to work, and readlink must clamp
correctly so we don't leak inode internals through an oversize read.

## Work

- [ ] Detect fast symlink: `S_ISLNK(mode) && i_blocks == 0 && i_size <= 60`
  - Note: `i_blocks` is in 512-byte units per ext2 spec
- [ ] Fast path: copy from `i_block[0..15]` bytes; length = `min(i_size, 60, user_buflen)`
- [ ] Slow path: `bread` the first data block; copy `min(i_size, block_size, user_buflen)` bytes
- [ ] `InodeOps::readlink` returns the copied bytes; **do not NUL-terminate**
- [ ] Path-walk resolution: when the walker encounters a symlink, call this and recurse
      (existing path-walk infrastructure already handles loop limits)
- [ ] Integration test: fast symlink with 10-byte target; slow symlink with 200-byte
      target; symlink cycle detection (already in path-walk)

## Out of scope

- Writing symlinks (Workstream E).

## Context

- Parent epic: #537
- Depends on: #559 (Ext2Inode/getattr), #560 (walker for slow path)
- RFC 0004 §Symlinks (fast/slow boundary at 60 bytes)

dburkart:
Closing as already implemented. PR #602 landed the fast-symlink path and an interim slow-symlink read; PR #645 wired the slow-symlink path through the indirect walker (closing follow-up #603). Both legs of the issue checklist are now in main.

Issue #561, opened 2026-04-18
fs/ext2: FileOps::read through the buffer cache


dburkart:
## Motivation

Implements RFC 0004 — Workstream D (ext2 read path, wave 2). Pulls together the
indirect-block walker and the buffer cache to service a `read(fd, buf, n)` against an
ext2 file. First user-visible sign the driver works.

## Work

- [ ] `impl FileOps::read for Ext2OpenFile` (or equivalent open-file state)
- [ ] Loop: for each block in the requested range, call `resolve_block`, then `bread`,
      then copy the slice into the caller buffer
- [ ] Sparse-hole handling: if `resolve_block` returns `None`, write zeros of block_size
      length into the output buffer (POSIX)
- [ ] Respect seek offset; advance by the number of bytes actually copied
- [ ] Return EOF when offset ≥ `i_size` (short read or 0)
- [ ] `atime` bump if mount is not `MS_NOATIME`; goes through the normal dirty-inode path
      (which does not exist until Workstream E — until then, `atime` updates are dropped
      and the test confirms this expectation)
- [ ] Integration test against fixture image: read a small file, read a large file
      crossing triple-indirect, read into a sparse file (verify the hole returns zeros)

## Out of scope

- Writing (Workstream E).
- Readahead / prefetch — Open Question in RFC, deferred.

## Context

- Parent epic: #537
- Depends on: #560 (indirect walker), #553 (bread)
- RFC 0004 §Read path

PR #608, opened 2026-04-19
fs/ext2: FileOps::getdents64 + directory lookup via Ext2DirEntry2 iteration (#562)

### LINKED ISSUE #562: fs/ext2: FileOps::getdents64 + directory lookup via Ext2DirEntry2 iteration
## Motivation

Implements RFC 0004 — Workstream D (ext2 read path, wave 2). Directory iteration is
the path-walk primitive: `lookup(parent, name)` iterates the same entries. Getting
the validation right is important — a malformed directory block can crash the walker
or leak kernel memory into userspace.

## Work

- [ ] `Ext2DirEntry2` iterator: walks the directory's data blocks, each entry is
      `{ inode: u32, rec_len: u16, name_len: u8, file_type: u8, name: [u8; name_len] }`
- [ ] **Skip inode == 0 entries universally** (tombstones for deleted entries; MUST
      NOT return them via getdents64 or lookup)
- [ ] Validation on every entry before accepting it:
  - `rec_len` must be ≥ `8 + name_len` and ≤ `block_size - offset_within_block`
  - `name_len` must be > 0 on live entries
  - `file_type` must be one of the known ext2 types (REG_FILE, DIR, CHRDEV, BLKDEV, FIFO, SOCK, SYMLINK, UNKNOWN) when `INCOMPAT_FILETYPE` is enabled
  - Reject `inode` values in the reserved range `[1, EXT2_GOOD_OLD_FIRST_INO)` except
    `EXT2_ROOT_INO (2)` and `s_last_orphan` is not a live entry anyway
- [ ] `FileOps::getdents64` emits `linux_dirent64` structs (existing format in the
      getdents64 syscall)
- [ ] `InodeOps::lookup` iterates and matches the name byte-for-byte (no Unicode normalization)
- [ ] Integration test: `ls /` on the fixture image, stat each returned entry, confirm
      malformed directory (rec_len overrun) is caught and returned as EIO

## Out of scope

- dir_index (HTree) — not supported.
- Directory mutation (create/rmdir/unlink entries) — Workstream E.

## Context

- Parent epic: #537
- Depends on: #559 (Ext2Inode/lookup), #561 (FileOps::read path shared helpers)
- RFC 0004 §Directory operations, §Security (dirent validation)

dburkart:
Closes #562.

## Summary

Workstream D, wave 2 of RFC 0004's ext2 driver. Adds `kernel/src/fs/ext2/dir.rs` carrying the directory iteration primitives — a per-block `DirEntryIter` with full RFC 0004 §Security validation, plus `lookup` and `getdents64` entry points that walk a directory's data blocks through the shared `indirect::resolve_block` walker.

- `DirEntryIter` — skips tombstones (ino == 0); rejects `rec_len` values that are too small, unaligned, or straddle the block; rejects zero-length names on live records; rejects unknown `EXT2_FT_*` codes when `INCOMPAT_FILETYPE` is set; and rejects reserved inos in `[1, EXT2_GOOD_OLD_FIRST_INO)` except `EXT2_ROOT_INO` (2).
- `dir::lookup(super, &Ext2Inode, name)` returns the matching ino or `ENOENT`.
- `dir::getdents64(super, &Ext2Inode, buf, &mut cookie)` emits Linux `linux_dirent64` records with the same byte layout ramfs/devfs/tarfs use.

Per the issue's scope, wiring these into `InodeOps::lookup` / `FileOps::getdents` on `Ext2Inode` is a follow-up — this PR doesn't touch `inode.rs` to avoid colliding with the parallel #561 work on the read path.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo test --lib --features ext2 fs::ext2::dir` — 15 host unit tests (tombstone skip, every corrupt-record flavour, reserved-ino range, file_type handling w/ and w/o `INCOMPAT_FILETYPE`, golden-image dot/dotdot decode, `dt_from_file_type`, `linux_dirent64` emission, short-buffer behaviour).
- [x] QEMU integration test `ext2_dir_ops` — 5 tests over the 64 KiB `golden.img`: `lookup(root, "lost+found") == 11`, `lookup(root, "missing") == ENOENT`, `lookup(root, "") == ENOENT`, `getdents64` single-call coverage, and `getdents64` with a 32-byte buffer forcing cookie-driven multi-call iteration.

## Follow-ups

- Integration PR that swaps `Ext2Inode::lookup` / `Ext2FileOps::getdents` stubs for calls into this module. Deferred to avoid colliding with #561 / #563 on `inode.rs` this wave.
- `MetadataMap` construction from the BGDT (walker currently uses `MetadataMap::empty()`; bounds validation via `Geometry` remains enforced).

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -380,3 +380,8 @@ required-features = ["ext2"]
 name = "ext2_block_alloc"
 harness = false
 required-features = ["ext2"]
+
+[[test]]
+name = "ext2_dir_ops"
+harness = false
+required-features = ["ext2"]
--- a/kernel/src/block/mod.rs
+++ b/kernel/src/block/mod.rs
@@ -48,6 +48,7 @@ pub const SECTOR_SIZE: usize = 512;
 /// | `NotInitialized`| `ENODEV` |
 /// | `Timeout`       | `EIO`    |
 /// | `OutOfRange`    | `EINVAL` |
+/// | `ReadOnly`      | `EROFS`  |
 /// | `NoMemory`      | `ENOMEM` |
 ///
 /// Kept deliberately small — the cache and ext2 driver only need to
--- a/kernel/src/block/mod.rs
+++ b/kernel/src/block/mod.rs
@@ -75,6 +75,12 @@ pub enum BlockError {
     Enospc,
     /// The requested byte range lies outside `[0, capacity())`.
     OutOfRange,
+    /// Write attempted against a read-only backend (e.g. a RO-mounted
+    /// image or a test harness deliberately rejecting writes). Distinct
+    /// from [`BlockError::DeviceError`] because the caller knows the
+    /// backend is intentionally immutable and can surface `EROFS` /
+    /// equivalent rather than a generic I/O error.
+    ReadOnly,
     /// Cache is full and eviction could find no victim buffer to reclaim.
     /// Every resident entry was either pinned (external `Arc` handle held
     /// by a caller) or mid-I/O (`DIRTY | LOCKED_IO`). Returned by
--- a/kernel/src/fs/ext2/dir.rs
+++ b/kernel/src/fs/ext2/dir.rs
@@ -381,6 +381,14 @@ mod block_backed {
 
         let block_count = size.div_ceil(block_size as u64);
         for logical in 0..block_count {
+            // Cap record parsing to the directory's logical size. The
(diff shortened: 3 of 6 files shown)


dburkart:
Smoke hit the known #478 flake (debugcon captured `e0 e1` — pre-write markers fired, post-write missing; init never completed first write). Retriggered smoke.

PR #618, opened 2026-04-23
kernel/exec: gate execve on PT_INTERP with ENOEXEC (#578)

### LINKED ISSUE #578: kernel/exec: execve gate — refuse PT_INTERP (dynamic linking) with ENOEXEC until ld.so lands
## Motivation

Once vibix mounts a real ext2 root, userspace binaries on disk will occasionally be dynamically linked (ELF `PT_INTERP` segment pointing at `/lib/ld-linux.so`). vibix has no dynamic linker, no shared-library loader, and no TLS setup; running such a binary would crash or, worse, jump to an unloaded pointer. Until `ld.so` lands, `execve` must refuse these binaries cleanly with `ENOEXEC`.

## Work

- [ ] In `kernel/src/binfmt/elf.rs::load_elf`, iterate program headers; if any `PT_INTERP` is present, return `Errno::ENOEXEC` immediately — do not partially map.
- [ ] Log at `info!` level with the binary path so test harnesses can see "dynamic linking not supported yet" without it looking like a crash.
- [ ] Allow statically-linked binaries through as before (no `PT_INTERP`, no `PT_DYNAMIC` with DT_NEEDED entries — the latter requires reading the dynamic table; the header-level check is sufficient for vibix's current binfmt policy).
- [ ] Tests: exec a statically-linked ELF → loads; exec a dynamically-linked ELF (any pre-built glibc binary) → ENOEXEC without crashing; log contains the "dynamic linking not supported" message.

## Out of scope

- Actual `ld.so` support — tracked separately when `track:posix` dynamic linking work begins. This issue is just the safety gate.

## Context

- Parent epic: #537
- RFC: `docs/RFC/0004-ext2-filesystem-driver.md` (§Execve gate for dynamic binaries)
- Depends on: boot-path root mount (this epic) — before that, there was no way to get a dynamic binary onto the FS.
- Blocks: nothing in this epic, but prevents a user-visible regression once ext2 is the root.

dburkart:
Closes #578

## Summary

- Until a userspace `ld.so` lands, vibix cannot run dynamically-linked ELFs. With the ext2 rootfs about to mount as `/`, host binaries referencing `/lib/ld-linux.so.*` would crash on exec; this change refuses them cleanly with `-ENOEXEC` at the `exec_atomic` boundary, logging the interpreter path at info level so the failure is distinguishable from a generic malformed-ELF error.
- The gate sits in the execve syscall path rather than in `load_user_elf_with_vmas`, keeping the loader's existing PT_INTERP plumbing usable from kernel bootstrap flows that ship an interpreter as a Limine module.
- Bumps `KSYMTAB_BYTES` from 256 KiB to 384 KiB: the test-crate symbol set grew past the old cap on ext2 integration tests (observed 263912 bytes), which was stopping xtask before any test could run. Reservation is patched post-link, so headroom is free.

## Test plan

- [x] `cargo xtask test` — new `execve_rejects_ptinterp` passes; all existing tests (including `execve_atomic` atomicity test) pass.
- [x] `cargo xtask smoke` — 33/33 markers present.
- [x] New test synthesizes an ELF64 with `PT_INTERP="/lib/ld-linux.so.2"`, calls `exec_atomic`, asserts `-ENOEXEC` and that the caller's address-space Arc pointer is unchanged (gate runs before any staged PML4 commit).

## Out of scope

- Actual `ld.so` / dynamic loader support. Tracked separately under `track:posix` dynamic-linking work.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -193,6 +193,10 @@ name = "execve_atomic"
 harness = false
 
 [[test]]
+name = "execve_rejects_ptinterp"
+harness = false
+
+[[test]]
 name = "fork_cow"
 harness = false
 
--- a/kernel/linker.ld
+++ b/kernel/linker.ld
@@ -60,7 +60,7 @@ SECTIONS
     } :data :dynamic
 
     /*
-     * The 256 KiB ksymtab reservation lives as a plain static in
+     * The 384 KiB ksymtab reservation lives as a plain static in
      * `.rodata` — no custom output section. Giving it its own output
      * section tripped rust-lld into merging `.data`/`.bss` into the
      * read-only rodata PT_LOAD, leaving writable globals mapped RO.
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -1042,6 +1042,39 @@ pub fn exec_atomic(elf_bytes: &[u8]) -> Result<core::convert::Infallible, i64> {
     use x86_64::structures::paging::{Page, PageTableFlags, Size4KiB};
     use x86_64::VirtAddr;
 
+    // 0. Gate: refuse dynamically-linked binaries (issue #578).
+    //
+    //    Until a userspace dynamic linker (ld.so) is implemented, vibix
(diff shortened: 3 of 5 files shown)


dburkart:
@coderabbitai review

Issue #573, opened 2026-04-18
fs/ext2: orphan list final-close sequence — truncate-to-zero, free, remove from chain


dburkart:
## Motivation

Implements RFC 0004 — ext2 filesystem driver (workstream E, write-path).

When the last fd referencing an unlinked inode closes, the driver must: (1) truncate the inode to zero (free all data+indirect blocks), (2) free the inode number, (3) remove the inode from the on-disk orphan chain, (4) drop the in-memory orphan pin. This is the consumer of the orphan-list machinery set up in `unlink` / `rmdir` (#569) and validated on mount in #564.

## Work

- [ ] Hook `Ext2InodeOps::evict` (the VFS op invoked when the last `Arc<Inode>` drops). Check `Ext2Inode::unlinked` — if true:
  1. Truncate to zero via the setattr truncate path (#572).
  2. Call `free_inode(ino, was_dir)` to clear the inode bitmap bit.
  3. Unchain the inode from the on-disk orphan list by patching its predecessor's `i_dtime` (or `s_last_orphan` if it was the head).
  4. Set `i_dtime = now` on the inode table entry (tombstone); write-back buffer.
  5. Remove from `Ext2SuperBlock::orphan_pins`.
- [ ] Treat `Ext2Inode::unlinked` as a veto against drop-eviction of a clean inode cache entry — even if the inode is "clean," it must not be discarded until this sequence runs.
- [ ] Tests: unlink open file, close fd → blocks freed, orphan chain empty, free counts restored; power-fail between steps (2) and (3) — on next mount the replay (#564) completes the sequence.

## Out of scope

- Background orphan reaper thread — mount-time replay covers the crash case; there is no need for a runtime sweep.

## Context

- Parent epic: #537
- RFC: `docs/RFC/0004-ext2-filesystem-driver.md` (§Orphan list, §Final-close sequence)
- Depends on: setattr truncate (this epic), inode allocator free (this epic), unlink/rmdir (this epic), mount-time replay (#564).
- Blocks: `s_state` valid-FS flag (this epic) — can't mark clean until orphan list is known-drained.

PR #626, opened 2026-04-23
fs/ext2: InodeOps unlink/rmdir with orphan-list add on hit-zero (#569)

### LINKED ISSUE #569: fs/ext2: InodeOps unlink/rmdir with orphan-list add on last open reference
## Motivation

Implements RFC 0004 — ext2 filesystem driver (workstream E, write-path).

POSIX `unlink(2)` removes the name; the inode only becomes reclaimable when `i_links_count` drops to zero **and** the last open fd closes. Between those two events the orphan list must reference the inode so a crash doesn't leak its blocks. This is the classic unlink-while-open semantic — non-negotiable for POSIX conformance.

## Work

- [ ] Implement `Ext2InodeOps::unlink(parent, dentry, cred)`:
  1. Permission-check parent with `MAY_WRITE | MAY_EXEC`; sticky bit check (`S_ISVTX`) — if set, only the file owner or parent owner may unlink.
  2. Remove dirent in parent: either mark as deleted by setting `rec_len` of the previous dirent to swallow it (or clear `inode` field) per RFC §Directory block format.
  3. Decrement `i_links_count`; update `i_ctime`.
  4. If `i_links_count == 0`: push onto on-disk orphan list (`s_last_orphan`), store in in-memory `Ext2SuperBlock::orphan_pins` as `Arc<Inode>` strong-ref, set `Ext2Inode::unlinked = AtomicBool(true)`.
- [ ] Implement `Ext2InodeOps::rmdir`: refuse unless directory is empty (only `.` and `..` dirents remain); decrement parent's `i_links_count` (the `..` back-link); decrement `bg_used_dirs_count`; reuse the orphan path above.
- [ ] Tests: unlink open file then close → blocks freed; unlink open file then crash mount → orphan-replay frees the inode; rmdir non-empty → ENOTEMPTY; unlink with sticky-bit denied; rmdir of `/` → EBUSY.

## Out of scope

- Concurrent unlink-rename interaction is covered by the rename issue.
- Quota accounting on free — quotas are not in ext2 rev 0 without `QUOTA` compat flags (future P2).

## Context

- Parent epic: #537
- RFC: `docs/RFC/0004-ext2-filesystem-driver.md` (§Orphan list, §Unlink semantics)
- Depends on: inode allocator free path (this epic), block allocator free path (this epic), dirent walker (#560), orphan-chain mount-time replay (#564).
- Blocks: setattr/truncate (this epic), rename (this epic).

dburkart:
Closes #569.

## Summary

- Implements `InodeOps::unlink` and `InodeOps::rmdir` on `Ext2Inode` via a new `fs/ext2/unlink.rs` module (RFC 0004 §Unlink semantics).
- Dirent removal: previous-live `rec_len` extend, or first-in-block tombstone via zeroed inode field.
- Link-count bookkeeping: decrement child on-disk `i_links_count` (2 for rmdir, 1 for unlink) and stamp `i_ctime`; mirror into both `Ext2InodeMeta` and VFS `InodeMeta` so still-open fds see the update.
- For rmdir: empty-directory check (ENOTEMPTY otherwise), parent's `i_links_count` decrement for the lost `..`-back-link, and BGDT `bg_used_dirs_count` decrement on the child's group.
- On hit-zero: mark `unlinked`, push on the on-disk orphan list (`s_last_orphan` update + `i_dtime` repurposed as next-pointer) and pin `Arc<Inode>` in `Ext2Super::orphan_list`. Block/inode freeing deferred to #573's final-close path.
- Per-mount `Ext2InodeCache` (parallel `BTreeMap<u32, Weak<Ext2Inode>>`) added to `Ext2Super` so the trait entry points can recover `Arc<Ext2Inode>` from `Arc<Inode>` without an unsafe downcast. `iget` publishes to both caches.

## Error surface

| Condition | Errno |
|---|---|
| RDONLY / FORCED_RDONLY mount | EROFS |
| `.` or `..` | EINVAL |
| Name not in parent | ENOENT |
| `unlink` on directory | EISDIR |
| `rmdir` on non-directory | ENOTDIR |
| `rmdir` on non-empty dir | ENOTEMPTY |

## Test plan

- [x] `cargo xtask build`
- [x] `cargo xtask test` — 493 host unit tests + all QEMU integration tests green, including new `ext2_unlink` (6 sub-tests: rmdir-of-lost+found-and-orphans, unlink-of-directory-is-eisdir, unlink/rmdir-missing-name-is-enoent, unlink/rmdir-of-dot-dotdot-is-einval, ro-mount-refuses-both)
- [x] `cargo fmt --all`

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -399,3 +399,8 @@ required-features = ["ext2"]
 name = "ext2_file_write"
 harness = false
 required-features = ["ext2"]
+
+[[test]]
+name = "ext2_unlink"
+harness = false
+required-features = ["ext2"]
--- a/kernel/src/fs/ext2/fs.rs
+++ b/kernel/src/fs/ext2/fs.rs
@@ -444,6 +444,7 @@ impl FileSystem for Ext2Fs {
             ext2_flags,
             owner: self.self_ref.clone(),
             inode_cache: super::inode::new_inode_cache(),
+            ext2_inode_cache: super::inode::new_ext2_inode_cache(),
             orphan_list: super::inode::new_orphan_list(),
             alloc_mutex: spin::Mutex::new(()),
             self_ref: weak.clone(),
@@ -548,6 +549,12 @@ pub struct Ext2Super {
     /// `Arc<Inode>` so repeat lookups dedup. Wave 3 (#559) introduces
     /// this; Workstream E populates it on every inode read.
     pub inode_cache: super::inode::InodeCache,
+    /// Driver-private parallel cache of `Weak<Ext2Inode>` keyed on the
+    /// same ext2 ino as [`inode_cache`]. The unlink path (#569) and
+    /// future setattr/truncate paths consume this to recover the
+    /// concrete `Arc<Ext2Inode>` without a dyn-Any downcast on
+    /// `Arc<dyn InodeOps>`.
+    pub ext2_inode_cache: super::inode::Ext2InodeCache,
     /// Strong-ref orphan list. Holds `Arc<Inode>` for every
     /// unlinked-but-open inode on this mount so its blocks aren't
     /// freed before the last close (RFC 0004 §Orphan-list residency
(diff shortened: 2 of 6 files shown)


dburkart:
Thanks for the review!

- **dir_is_empty doc** — fixed in 3c5b9e1 (comment now correctly describes walking all allocated blocks).
- **Unused `sb_vfs` parameter** — dropped in 3c5b9e1.
- **RamDisk duplication across ext2 tests** — tracked as #627 so it can be done in one pass across all three current call sites (and absorb future ext2 integration tests as they land).

Issue #579, opened 2026-04-18
xtask: deterministic ext2 image builder (mkfs.ext2 -E hash-seed, fixed UUID, reproducible content)


dburkart:
## Motivation

Implements RFC 0004 — ext2 filesystem driver (workstream F, mount plumbing).

Smoke/CI needs a byte-identical ext2 image across runs so failures reproduce deterministically and image diffs are meaningful. `mkfs.ext2` defaults randomize UUID, hash-seed, and timestamps; we need to pin them via a small xtask that drives `mkfs.ext2` with `-E no_copy_xattrs -U <fixed> -E hash_seed=<fixed>` and then `debugfs` scripted inserts for /init and a known fixture tree.

## Work

- [ ] Add `xtask/src/ext2_image.rs::build()`:
  - Create a fixed-size (e.g., 64 MiB) zero-filled file.
  - Run `mkfs.ext2 -q -L vibix-root -U a1a2a3a4-0000-1111-2222-333344445555 -E hash_seed=... -E packed_meta_blocks=0 -t ext2 -r 0 -I 128 <image>`.
  - Scripted `debugfs`: `mkdir /bin`, `mkdir /tmp`, `write <built-init-binary> /init`, `chmod 0755 /init`.
  - Fix `s_mtime` / `s_wtime` / `s_lastcheck` to a known epoch (e.g., 1700000000).
  - Emit the image at `target/vibix-root.ext2`.
- [ ] Integrate with `cargo xtask smoke` so the QEMU run picks up the image via `-drive format=raw,file=target/vibix-root.ext2,if=none,id=drive0` + matching virtio-blk device line.
- [ ] Produce a stable sha256 of the image; commit the expected hash under `tests/fixtures/ext2_image.sha256` and fail xtask if the hash drifts.
- [ ] Tests: run xtask twice; images are byte-identical.

## Out of scope

- Embedding the mkfs.ext2 logic in Rust directly — shelling out to e2fsprogs is fine; the host toolchain already has it.

## Context

- Parent epic: #537
- RFC: `docs/RFC/0004-ext2-filesystem-driver.md` (§Deterministic image generation)
- Depends on: nothing in the driver code (this is pure host-side tooling), but landed last in Workstream F so the smoke harness can switch over in one shot.
- Blocks: Workstream G (pjdfstest) — the conformance suite needs a reproducible image to run against.

PR #630, opened 2026-04-23
kernel/boot: parse root=/rootflags= cmdline + select root fs with fallback

### LINKED ISSUE #577: kernel/boot: select root FS — ext2 on virtio-blk with tarfs/ramfs fallback, rootflags= parsing
## Motivation

Today vibix hardcodes the rootfs at init (either RamFs or TarFs, depending on build). To boot from a real ext2 disk image, the boot path needs to: (1) probe virtio-blk for a valid ext2 superblock, (2) mount it as `/` via the same FileSystem vtable used by `mount(2)`, (3) fall back to TarFs/RamFs cleanly when no ext2 device is present (so existing smoke tests don't regress).

## Work

- [ ] Add `kernel/src/init/rootfs.rs::select_root()` returning a `SuperBlock`.
- [ ] If `boot_args` contains `root=ext2://virtio-blk0` (or equivalent): probe LBA 2 (offset 1024 / block_size) for magic 0xEF53; if valid, call `Ext2FileSystem::mount()` and return its super.
- [ ] Else fall back to the existing TarFs/RamFs construction.
- [ ] Honor `rootflags=ro` / `rootflags=rw` on the kernel command line; default to `ro` until orphan replay (#564) has been exercised in CI.
- [ ] Ensure the existing `/dev`, `/tmp` mounts still layer on top without regressing smoke.
- [ ] Tests: xtask qemu-run with an ext2 image boots to userspace and executes `/init`; xtask qemu-run without an image still boots via tarfs (legacy path intact); `rootflags=rw` + induced mount error → clean fallback to RO.

## Out of scope

- Initramfs pivot (classic two-stage boot) — noted in RFC Alternatives; future P2.

## Context

- Parent epic: #537
- RFC: `docs/RFC/0004-ext2-filesystem-driver.md` (§Boot-path root selection)
- Depends on: mount(2) (this epic), Ext2FileSystem::mount (#557), s_state (#574), execve gate (this epic).
- Blocks: xtask deterministic image (this epic) — the CI needs both a populated image and a kernel that can mount it.

dburkart:
Closes #577

## Summary
- New host-testable `boot_cmdline` parser for `root=<source>` and `rootflags=<csv>`. Accepts `/dev/vda` / `ext2` / `tarfs-module` / `tarfs` / `ramfs`; csv flags `ro,rw,nosuid,noexec,nodev`. Unknown root values fall back to `Default` (boot never wedges on a typo). Last-`rootflags=` token wins, replacing prior flag state. 16 unit tests.
- `vfs::init` gains `init_with(RootArgs)`; old `init()` delegates via `RootArgs::auto()` so the 80+ existing integration tests keep their current behavior. New `try_mount_selected_root` walks `[VirtioBlk, TarfsModule, Ramfs]` by default, logging `vfs: mount <src> / failed: errno=<n> (trying next)` per attempt and `vfs: mounted <src> at / (requested=<orig>)` on success. VirtioBlk mounts RDONLY by default pending orphan replay (#564); explicit `rootflags=rw` overrides.
- `main.rs` now parses and logs the cmdline result (pure observability — prod boot path does not itself call `vfs::init::init_with` yet; tests do). The existing `writeback::parse_cmdline` path is untouched.
- `xtask --root=ext2` builds the deterministic image from #579 and wires it through virtio-blk with `root=/dev/vda`. New `inject_limine_cmdline` helper rewrites the Limine config stanza (append or replace); 4 new unit tests cover its edge cases.
- Host/target split: local `MountFlags` newtype in `boot_cmdline` mirrors `vfs::dentry::MountFlags`; compile-time asserts pin the layout on target and a `From` impl handles the conversion. Parser is `#![cfg_attr(not(test), no_std)]`-clean.

## Test plan
- [x] `cargo xtask test` — 509 host unit tests pass (adds 16 for parser + 4 for inject_limine_cmdline)
- [x] `cargo xtask smoke` — 33/33 markers
- [x] Full integration suite boots with virtio-blk attempt → ext2 errno=-19 → tarfs fallback → root mounted; observed in kernel serial log: `vfs: mount ext2 (virtio-blk) / failed: errno=-19 (trying next)` → `vfs: mounted tarfs (module) at / (requested=Default)`

### DIFF
--- a/kernel/src/boot_cmdline.rs
+++ b/kernel/src/boot_cmdline.rs
@@ -1,0 +1,373 @@
+//! Kernel-command-line parsing for root-filesystem selection.
+//!
+//! Issue #577 (RFC 0004 Workstream F). Extracts `root=<source>` and
+//! `rootflags=<csv>` from the Limine-provided cmdline string so the
+//! boot path can choose between mounting ext2 on virtio-blk, falling
+//! back to the Limine-delivered tarfs module, or a throwaway ramfs.
+//!
+//! The parser is deliberately tolerant: unknown tokens are ignored
+//! (future knobs coexist without a dispatch switch here), and a second
+//! occurrence of `root=` or `rootflags=` wins (last-writer-wins, same
+//! policy as `writeback_secs`). An empty or missing `root=` resolves
+//! to [`RootSource::Default`], which the mount path interprets as
+//! "prefer ext2 on the default block device, fall back to tarfs then
+//! ramfs."
+//!
+//! # Accepted `root=` values
+//!
+//! | Token            | Resolves to           |
+//! |------------------|-----------------------|
+//! | `/dev/vda`       | `RootSource::VirtioBlk` (try ext2 on virtio-blk) |
+//! | `ext2`           | `RootSource::VirtioBlk` (alias, no `/dev/` prefix) |
+//! | `tarfs-module`   | `RootSource::TarfsModule` |
+//! | `ramfs`          | `RootSource::Ramfs`   |
+//! | *(unset)*        | `RootSource::Default` (auto-probe) |
+//! | *(other)*        | `RootSource::Default` (unknown → defer to auto-probe + log) |
+//!
+//! # Accepted `rootflags=` tokens
+//!
+//! Comma-separated; whitespace around commas is stripped; unknown
+//! tokens are silently dropped (bubbles up as "nothing set").
+//!
+//! - `ro` → [`MountFlags::RDONLY`]
+//! - `nosuid` → [`MountFlags::NOSUID`]
+//! - `noexec` → [`MountFlags::NOEXEC`]
+//! - `nodev` → [`MountFlags::NODEV`]
+//!
+//! `rw` is accepted and explicitly clears `RDONLY` — this lets the
+//! cmdline override a default-RO policy (which #577 installs so the
+//! orphan-replay path can be exercised in CI before we flip to RW by
+//! default).
+
+// `MountFlags` from `crate::fs::vfs::dentry` is only compiled on
+// `target_os = "none"`, but the parser has to run on the host too
+// (unit tests). Rather than branch the struct, declare a local
+// newtype here whose bit layout matches `vfs::dentry::MountFlags`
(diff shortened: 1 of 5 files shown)


dburkart:
Addressed the major duplicate-`cmdline:` finding in cd7c022 — switched `inject_limine_cmdline` to a two-pass approach (pre-scan for an existing entry, then either replace-in-place or append, never both) and added an `out.matches("cmdline:").count() == 1` regression assert.

On the `VirtioBlk` → `TarfsModule` fallback nit: this is intentional for #577. The issue title explicitly asks for "ext2 on virtio-blk with tarfs/ramfs fallback," and operationally an explicit `root=/dev/vda` boot that can't mount ext2 (unformatted disk, bad superblock, missing virtio) is better served by landing on tarfs and emitting a clear error trail than by dropping straight to an empty ramfs. Happy to revisit if we want a `rootflags=nofail`-style opt-out later.

dburkart:
Addressed both findings on the latest round:

- **Multi-stanza `in_block` scoping (02b75ab)**: now clears `in_block` when a later column-0 non-blank non-comment line opens a new stanza, in both the pre-scan and emit passes. Added `does_not_mutate_later_stanzas` regression test with a trailing `/other` multiboot2 stanza.
- **`--root=` not wired as global flag (dade70c)**: new `take_value()` helper pulls `--root=<kind>` out of the pre-subcommand arg vector alongside `--release` etc., then threads `Option<&str>` through `run` → `run_with_root`. Drops the stale `env::args()` re-read. Now `cargo xtask --root=ext2 run` and `cargo xtask run --root=ext2` both work.

On **`RootArgs` not wired through the prod `_start` path**: acknowledged and deferred — will file a follow-up to call `vfs::init::init_with(root_args)` from `_start` after `block::init()` is ready. Keeping it out of this PR is deliberate: threading the mount into the prod boot path is a behavior change (vs. today's main.rs which doesn't mount a VFS at all), needs careful ordering around `block::init()`/`block::ready()`, and would invalidate the "observability-only" comment that explicitly documents the current state. Expanded the xtask comment to call this out as a QEMU-side verification knob for now.

Issue #575, opened 2026-04-18
kernel/syscall: implement mount(2) — bind FS type by name, enforce MS_RDONLY/MS_NOSUID/MS_NOEXEC


dburkart:
## Motivation

Implements RFC 0004 — ext2 filesystem driver (workstream F, mount plumbing).

There is no `mount(2)` syscall in vibix today — every mount is hardcoded at init. To load an ext2 disk image at runtime we need a real `mount` syscall that accepts a source (block device path), target, fstype, and flags, dispatches to the registered FileSystem vtable, and installs the superblock at the target mount point.

## Work

- [ ] Wire `__NR_mount` (Linux x86_64 syscall 165) in `kernel/src/arch/x86_64/syscalls/vfs.rs`.
- [ ] Signature: `sys_mount(source: *const u8, target: *const u8, fstype: *const u8, flags: u64, data: *const u8) -> Result<(), Errno>`.
- [ ] Require `cred.euid == 0` (only uid 0 may mount); plumb via the new Credential path (#546). Until setuid/etc. land, uid 0 is the init-task default.
- [ ] Look up the fstype by name in a registry (`FileSystemRegistry` — add a small static map, populated by `register_filesystem`).
- [ ] Resolve the source path to a block device (via existing `/dev` DevFs); reject non-block-devices with ENOTBLK unless the FS is `MS_NODEV`-like (no-source FSes like ramfs/tmpfs — source may be empty string).
- [ ] Resolve the target path to a dentry; reject ENOTDIR for non-dir; reject EBUSY if the dentry already has a superblock mounted on it.
- [ ] Respect flags: `MS_RDONLY` (0x01), `MS_NOSUID` (0x02), `MS_NOEXEC` (0x08), `MS_NODEV` (0x04). Store on `VfsMount` and enforce at syscall entry points (execve must check MS_NOEXEC; setuid bit must be masked on MS_NOSUID).
- [ ] Install the new superblock in the mount table at target dentry; the existing VFS lookup code already consults this when walking.
- [ ] Tests: mount ext2 image at /mnt; mount as non-root → EPERM; mount over a busy dentry → EBUSY; mount with MS_RDONLY; exec from an MS_NOEXEC mount → EACCES.

## Out of scope

- `MS_BIND`, `MS_MOVE`, `MS_SHARED` (bind mounts / mount propagation) — future P2.
- `mountinfo`/`/proc/mounts` — there is no procfs yet; `findmnt`-style discovery can land when procfs does.

[...]

PR #634, opened 2026-04-23
kernel/syscall: implement umount2(2) — MNT_FORCE, MNT_DETACH, busy check (#576)

### LINKED ISSUE #576: kernel/syscall: implement umount2(2) — MNT_FORCE, MNT_DETACH, busy-check
## Motivation

Counterpart to `mount(2)` (#575). Without `umount2` there is no way to flush dirty buffers, drain orphan pins (#573), and set `s_state = EXT2_VALID_FS` (#574) — every shutdown looks like a crash to the next mount. This is also what `shutdown` + init will call during reboot.

## Work

- [ ] Wire `__NR_umount2` (Linux x86_64 syscall 166) in `kernel/src/arch/x86_64/syscalls/vfs.rs`.
- [ ] Signature: `sys_umount2(target: *const u8, flags: u32) -> Result<(), Errno>`.
- [ ] Require `cred.euid == 0`.
- [ ] Default (flags == 0): if any inode on this mount has open fds or any dentry is in use, return EBUSY. Otherwise call `SuperOps::put_super`, which flushes the buffer cache and sets `s_state` clean.
- [ ] `MNT_FORCE` (0x01): not supported initially — return ENOTSUP (flagged in RFC Alternatives).
- [ ] `MNT_DETACH` (0x02): lazy umount. Detach from the dentry tree now; the real `put_super` runs when the last open fd/dentry drops.
- [ ] Refuse to umount `/` unless the rootfs is being pivoted (future work; today return EBUSY).
- [ ] Tests: mount → umount; umount with open fd → EBUSY; MNT_DETACH + open fd + close fd → delayed put_super; umount as non-root → EPERM.

## Out of scope

- `pivot_root(2)`, `chroot(2)` — separate follow-ups, not required for root ext2 mount.

## Context

- Parent epic: #537
- RFC: `docs/RFC/0004-ext2-filesystem-driver.md` (§umount2(2), §Lazy umount)
- Depends on: mount(2) (this epic); orphan final-close (#573); s_state (#574).
- Blocks: boot-path root mount (this epic) — clean reboot requires clean umount.

dburkart:
Closes #576

## Summary

Implements `SYS_umount2` (syscall 166) on top of the VFS mount table landed
in #575 (RFC 0004 Workstream F).

- **Default (flags=0)**: strict busy check — returns EBUSY if any
  dentry on the mount is pinned or if any `SbActiveGuard` is live.
- **MNT_FORCE (0x1)**: aborts pending I/O and tears down inline. Refuses
  with EBUSY when the target has nested child mounts (safer than glibc's
  "force-through-nested").
- **MNT_DETACH (0x2)**: lazy unmount. Always succeeds; removes the edge
  from the mount table immediately, but defers `SuperOps::unmount` until
  the last `SbActiveGuard` drops. New `PENDING_DETACH` list keeps the
  SuperBlock Arc alive; `SbActiveGuard::drop` calls `finalize_pending_detach`
  when the active count hits zero.
- `SuperOps::sync_fs` runs before teardown on writable mounts.
- Superuser-only: non-root → EPERM. Unknown flag bits → EINVAL.

Target-path resolution accepts either the mountpoint dentry or a path
that crosses into the mount (walks up via `GlobalMountResolver::mount_above`).

## Test plan

- [x] `cargo xtask test` — all umount2_flags tests pass (7/7); one
      pre-existing flaky QEMU timeout on `wait4_condvar_race` passes
      in isolation.
- [x] New integration test `kernel/tests/umount2_flags.rs` covers:
  - `default_unmount_no_pins_succeeds`
  - `default_unmount_with_active_guard_returns_ebusy`
  - `force_with_active_guard_tears_down`
  - `force_refuses_nested_child_mount`
  - `detach_defers_ops_unmount_until_guard_drops`
  - `detach_with_no_guards_finalizes_inline`
  - `detach_allows_nested_child_mount`

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -419,3 +419,7 @@ required-features = ["ext2"]
 name = "ext2_link"
 harness = false
 required-features = ["ext2"]
+
+[[test]]
+name = "umount2_flags"
+harness = false
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -954,6 +954,13 @@ pub unsafe extern "C" fn syscall_dispatch(
         #[cfg(feature = "vfs_creds")]
         MOUNT => unsafe { super::syscalls::vfs::sys_mount_impl(a0, a1, a2, a3, a4) },
 
+        // umount2(target, flags) — RFC 0004 §umount2(2). Issue #576.
+        // Superuser-only; MNT_FORCE = abort in-flight I/O + detach
+        // (refuses nested mounts with EBUSY); MNT_DETACH = lazy unmount
+        // (unlink now, finalize when last SbActiveGuard drops).
+        #[cfg(feature = "vfs_creds")]
+        UMOUNT2 => unsafe { super::syscalls::vfs::sys_umount2_impl(a0, a1 as u32) },
+
         // poll(fds, nfds, timeout_ms) — wait for readiness on a set of fds.
         POLL => crate::poll::syscalls::sys_poll(a0, a1, a2 as i64),
 
@@ -1645,6 +1652,7 @@ pub mod syscall_nr {
     pub const FCHMODAT: u64 = 268;
     pub const UTIMENSAT: u64 = 280;
     pub const MOUNT: u64 = 165;
+    pub const UMOUNT2: u64 = 166;
     pub const POLL: u64 = 7;
     pub const SELECT: u64 = 23;
     pub const PSELECT6: u64 = 270;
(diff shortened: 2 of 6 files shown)


dburkart:
Both CodeRabbit findings are legitimate but cross-cutting enough that they deserve their own PRs rather than blocking umount2 landing:

- Target-path canonicalization (`/mnt/subdir` → `/mnt`, `/` → EBUSY) deferred to #636.
- Dentry-pin counter on `SuperBlock` (covers `chdir`/fd-table-pinned dentries that are invisible to `sb_active`) deferred to #637.

Landing the scoped implementation now; follow-ups will tighten both gates.

Issue #646, opened 2026-04-27
kernel/task: stack VA slot leaks on fork+exec reap path


dburkart:
## Motivation

Every successful fork+exec+wait cycle emits a kernel log line of the form:

```
tasks: reaped task N (stack VA slot 0xffffd00000014000..0xffffd00000019000 leaked)
```

Reproduced on `main` at 6d30c36 in a single boot via `cargo xtask run`. The reaper observes that the stack VMA is freed but the *VA slot* in the per-process stack arena is not returned, so each fork permanently consumes a 20 KiB slot. Long-running smoke / soak runs exhaust the arena. Likely shares root cause with #527 (CoW user-stack-top fault) — the same arena bookkeeping that misses the slot return is plausibly mis-marking the CoW page perms.

## Work

- [ ] Identify where the stack-VA-slot is reserved on fork (`Task::new_forked` / address-space clone path).
- [ ] Identify why the reaper's leak-detection branch fires — is the slot bookkeeping owned by VmaList, by a per-task arena, or by a separate stack allocator?
- [ ] Fix the slot-return path so the leak line stops printing on the happy path.
- [ ] Add an invariant assertion: at task reap, every stack VA slot must be returned (panic / log_error if not).
- [ ] Confirm whether the fix also collapses #527 (CoW write-fault at stack top) or whether #527 is genuinely independent.

## Context

- Observed in boot log on `main` 6d30c36 (2026-04-27).
- Related: #527 (residual fork CoW fault — likely same arena), #501 (quality sprint).
- Likely files: `kernel/src/task/task.rs`, `kernel/src/mem/vma.rs`, `kernel/src/process/`.

PR #649, opened 2026-04-27
ci: nightly 1000x smoke soak gating on residual flake markers

### LINKED ISSUE #648: ci: nightly 1000x smoke soak gating on residual flake markers (#420 unblocker)
## Motivation

#420 demoted `hello: hello from execed child` and `init: fork+exec+wait ok` to `SMOKE_SOFT_MARKERS` so CI stays green during the fork/exec quality sprint (#501). Once #478 + #527 land, the markers move back to required (#420). But re-promoting markers is not enough on its own — the residuals are *intermittent*, so even a green smoke run is weak evidence the bug class is gone. We need a higher-N soak that can fail loud on residual recurrence.

#507 (already merged) wired a 100× soak as a sprint-acceptance gate. This issue extends that into a **standing nightly gate** keyed to the specific residual signatures so any future regression that revives the bug class is caught in ≤ 24 h instead of leaking to manual QA.

## Work

- [ ] Add a `nightly-soak` GitHub Actions workflow that runs the smoke kernel 1000× under unaccelerated TCG.
- [ ] Pattern-match each run's serial output for the residual signatures and fail the run on any hit:
  - `ring3-first-fault: #PF` (the #527 user-stack-top fault).
  - `init: iretq to ring-3` followed by no `init: hello from pid 1` within timeout (the #478 silent-stall).
  - `tasks: reaped task .* (stack VA slot .* leaked)` (the stack-VA-slot leak).
- [ ] Publish per-run pass/fail to a job summary so flake density is visible across nightlies.
- [ ] Auto-file an issue (via `/file-issue`) when the workflow fails three nights in a row.

## Context

- Blocked-by → unblocks: #420 (re-harden markers), #511 (tune HARD_CAP back).
- Sibling: #507 (one-shot 100× soak), #501 (quality-sprint epic).
- Likely files: `.github/workflows/`, `xtask/src/smoke.rs`.

dburkart:
Closes #648.

## Summary

- Adds `.github/workflows/nightly-soak.yml`: a standing nightly workflow that runs `cargo xtask smoke` 1000× under unaccelerated TCG, sharded into 10 parallel matrix runners (~100 runs/shard, well under the 6 h job cap).
- Each run's captured serial is pattern-matched for residual signatures from the #501 quality sprint:
  - `ring3-first-fault: #PF` (#527 user-stack-top fault)
  - `init: iretq to ring-3` not followed by `init: hello from pid 1` (#478 silent-stall)
  - `tasks: reaped task ... (stack VA slot ... leaked)` (#646 stack-VA-slot leak)
- The workflow fails on any residual hit in any of the 1000 runs. Per-shard and aggregate counts are published to the job summary so flake density is visible across nightlies.
- On a 3-nights-red streak (current run red + the 2 prior schedule runs also red), the `aggregate` job auto-files a `priority:P1` `area:debug` tracking issue via `gh issue create` (or comments on the existing open one to avoid nightly spam).
- Adds a small `VIBIX_SMOKE_SERIAL_LOG` env-var hook to `xtask/src/smoke.rs` so the soak driver can persist each run's captured serial to a file for pattern-matching, without re-routing QEMU stdio. No kernel-side changes — the signatures are markers PR #595 already added.

The existing `smoke-soak.yml` (#507's 100× soak) is intentionally unchanged: that workflow gates on overall pass-rate, this one gates on named residual signatures — they are complementary signals.

## Test plan

- [x] `cargo fmt --all` clean.
- [x] `cargo check -p xtask` clean (xtask compiles with the new env-var hook).
- [ ] Workflow YAML parses and the new file appears in the PR's checks list (no `pull_request` trigger on this workflow — it's `schedule` + `workflow_dispatch` only, so no PR-time signal is expected; landed-on-main correctness will be validated by the first nightly run or a manual `workflow_dispatch`).
- [ ] CI green on this PR (rustfmt, clippy, smoke, tests).

### DIFF
--- a/.github/workflows/nightly-soak.yml
+++ b/.github/workflows/nightly-soak.yml
@@ -1,0 +1,506 @@
+name: nightly-soak
+
+# Standing nightly 1000× smoke soak gating on residual flake markers
+# from the #501 quality sprint — see issue #648.
+#
+# Runs `cargo xtask smoke` 1000 times under unaccelerated TCG and fails
+# the workflow on any of these residual signatures appearing in the
+# captured serial of any single run:
+#
+#   - `ring3-first-fault: #PF`             — #527 user-stack-top fault.
+#   - `init: iretq to ring-3` followed by
+#     no `init: hello from pid 1`          — #478 silent stall.
+#   - `tasks: reaped task ... (stack VA
+#     slot ... leaked)`                    — #646 stack-VA-slot leak.
+#
+# All three signatures are diagnostic markers added in PR #595. The
+# soak runs as a *standing gate*: any future regression that revives
+# the bug class is caught within ≤ 24 h instead of leaking to manual
+# QA.
+#
+# Sharding: 1000 runs at ~30-60 s each under TCG would be ~8-16 h on a
+# single GitHub-hosted runner, blowing past the 6 h job cap. Split into
+# `shard_count` parallel matrix shards so each shard stays well under
+# the cap. The post-shard `aggregate` job sums per-shard results into
+# one job summary.
+#
+# Auto-filing: if the workflow has failed three nights in a row at the
+# moment a new run completes red, the `aggregate` job opens a tracking
+# issue via `gh issue create` (project label `priority:P1`,
+# `area:debug`) so the on-call human is paged loudly. See the
+# `auto-file-followup` step.
+
+on:
+  schedule:
+    # 03:42 UTC nightly — staggered away from `smoke-soak.yml` (08:17),
+    # `ci.yml`, and other daily jobs in this repo.
+    - cron: '42 3 * * *'
+  workflow_dispatch:
+    inputs:
+      total_runs:
+        description: 'Total number of smoke runs to execute across all shards'
+        required: false
+        default: '1000'
+      shard_count:
+        description: 'Number of parallel shards (must evenly divide total_runs)'
+        required: false
+        default: '10'
+
+permissions:
+  contents: read
+  # Required for the auto-file-followup step (`gh issue create`) and
+  # for reading recent workflow runs to detect 3-nights-red streaks.
+  issues: write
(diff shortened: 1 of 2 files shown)


dburkart:
Addressed CodeRabbit's three findings in 5b1338b:

1. **Heredoc would escape the YAML `run: |` block** — replaced the column-0 `cat <<EOF` body with a `printf`-fed temp file (`mktemp`), passed via `gh issue create --body-file`. No more YAML-block-scalar termination risk.
2. **Missing shard counters silently aggregated as fewer runs** — the aggregate job now hard-fails (with a `::error::` and a job-summary block) if `shards_seen < SHARD_COUNT` or `total < TOTAL_RUNS_EXPECTED`. A shard that crashed before writing its counters can no longer let a red night slip through green.
3. **Late serial lines could be lost from `accumulated`** — the smoke loop now drains the reader thread's mpsc receiver (`while let Ok(line) = rx.try_recv()`) after the reader is joined and before persisting `VIBIX_SMOKE_SERIAL_LOG`, so a residual signature emitted in the final ms before the pipe closes is captured.

dburkart:
Round 2 of CodeRabbit findings addressed in dc3f8a9:

- **`set -euo pipefail` for the auto-file step** — transient `gh api` / `jq` failures now abort the job instead of silently producing empty `runs_json` and skipping the 3-nights-red filing.
- **`setsid cargo xtask smoke`** — orphan QEMU prevention. `timeout`'s kill now reaches the entire process tree.
- **Per-run TSV → collapsible Markdown table in `GITHUB_STEP_SUMMARY`** — the per-run density called out in #648 is now visible in the Actions UI without downloading the artifact.
- **`cancel-in-progress` gated on `workflow_dispatch`** — comment now matches behavior; manual retriggers replace the in-flight run, scheduled runs remain independent.

Issue #547, opened 2026-04-18
sys: wire getuid/geteuid + getgid/getegid


dburkart:
## Motivation

Implements RFC 0004 — Workstream B (Credential enforcement, wave 1). Read-only identity
queries; trivial but needed before setuid/setgid can be tested, and by pjdfstest to
assert what the current identity is.

## Work

- [ ] Wire `getuid()` → returns `task.credentials.uid`
- [ ] Wire `geteuid()` → returns `task.credentials.euid`
- [ ] Wire `getgid()` → returns `task.credentials.gid`
- [ ] Wire `getegid()` → returns `task.credentials.egid`
- [ ] Read via `Arc` snapshot so a concurrent `setuid` doesn't tear; these calls should be wait-free
- [ ] Integration test: a task with non-zero ruid/euid sees the right values

## Out of scope

- Anything mutating (setuid/setgid) — next B issue.
- `gettid()` / `getpid()` — orthogonal, already wired.

## Context

- Parent epic: #537
- Depends on: #546 (Credential struct extension)
- RFC 0004 §Kernel-Userspace Interface

PR #650, opened 2026-04-27
sys: wire setgroups/getgroups with NGROUPS_MAX=32

### LINKED ISSUE #549: sys: wire setgroups/getgroups with NGROUPS_MAX=32
## Motivation

Implements RFC 0004 — Workstream B (Credential enforcement, wave 1). Supplementary
groups are the third input to POSIX group-permission checks (alongside `gid`/`egid`).
Without setgroups, a multi-group user can't act with their group memberships, and
pjdfstest's group-permission matrix can't be exercised.

## Work

- [ ] Wire `getgroups(size, list[])` — returns `task.credentials.groups.len()` if `size == 0`, else copies up to `size` entries and returns the count
- [ ] Wire `setgroups(size, list[])` — root-only in this epic; replaces `groups` with the given vector
- [ ] Cap `NGROUPS_MAX = 32` — return `EINVAL` if `size > 32`; this matches the RFC's Open Question and is revisitable
- [ ] `InodeOps::permission` already uses the group list via `is_in_group` helper — verify no regression
- [ ] Errno coverage: `EFAULT`, `EINVAL`, `EPERM` per RFC errno table
- [ ] Integration test: setgroups then fchmod a file to group-writable and verify the membership gate

## Out of scope

- `CAP_SETGID` — non-root setgroups callers are rejected with `EPERM` this epic.

## Context

- Parent epic: #537
- Depends on: #546 (Credential struct)
- RFC 0004 §Kernel-Userspace Interface, §Open Questions (NGROUPS_MAX value)
- POSIX.1-2017 §getgroups, §setgroups

dburkart:
Closes #549.

## Summary

Implements RFC 0004 Workstream B wave 1 supplementary-group syscalls so multi-group users can exercise the POSIX group-permission matrix:

- `getgroups(size, list)` (SYS 115): `size==0` returns the count without touching the buffer (POSIX query form); `size>0` copies up to `size` u32 group IDs out and returns the actual count; `EINVAL` when `size>0 && size<count`; `EFAULT` on a bad user pointer.
- `setgroups(size, list)` (SYS 116): root-only this epic (`CAP_SETGID` is out of scope per the issue); `EPERM` for non-root, `EINVAL` when `size > NGROUPS_MAX (=32)` or `size < 0`, `EFAULT` on a bad pointer. The supplementary list is replaced wholesale.
- `NGROUPS_MAX = 32` per RFC 0004 §Open Questions — intentionally tight, revisitable.

Both arms slot into the existing wait-free Arc-snapshot model from #547/#548: readers clone `Arc<Credential>` under a short read-lock; `setgroups` builds a fresh `Credential` and swaps via `replace_current_credentials`. Failed `setgroups` calls never mutate the snapshot.

The dispatch arms are unconditional, matching `setuid`/`setgid` from #598. The existing `creds.rs` module doc captures the rationale: credential syscalls operate on the caller's own task snapshot rather than dereferencing VFS state, so they don't need the `vfs_creds` A↔B gate.

`InodeOps::permission` already consults `cred.groups` via `is_in_group`; no changes needed there — `setgroups` simply makes the slot reachable from userspace.

## Test plan

- [x] `cargo xtask build` green
- [x] New `kernel/tests/syscall_setgroups_family.rs` (12 cases): size==0 query, copy path, buffer-too-small EINVAL, negative-size EINVAL on both calls, root replace + clear, non-root EPERM, `>NGROUPS_MAX` EINVAL, boundary at NGROUPS_MAX, failed setgroups doesn't mutate, getgroups round-trip
- [x] All 12 pass under QEMU
- [x] `syscall_setuid_family` (15 cases) and `syscall_getuid_family` (5 cases) regression-clean
- [ ] Full `cargo xtask test` matrix on CI

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -325,6 +325,10 @@ name = "syscall_setuid_family"
 harness = false
 
 [[test]]
+name = "syscall_setgroups_family"
+harness = false
+
+[[test]]
 name = "syscall_truncate"
 harness = false
 
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -1031,6 +1031,17 @@ pub unsafe extern "C" fn syscall_dispatch(
         // setresuid.
         SETRESGID => super::syscalls::creds::sys_setresgid(a0 as u32, a1 as u32, a2 as u32),
 
+        // getgroups(size, list[]) — POSIX.1-2017 §getgroups. Issue #549.
+        // size==0 returns the count without touching the buffer; size>0
+        // copies up to `size` u32 group IDs to user. EINVAL if size is
+        // non-zero and smaller than the count.
+        GETGROUPS => super::syscalls::creds::sys_getgroups(a0 as i32, a1),
+
+        // setgroups(size, list[]) — POSIX.1-2017 §setgroups. Issue #549.
+        // Root-only this epic (CAP_SETGID is out of scope per RFC 0004
+        // Workstream B wave 1); list bounded at NGROUPS_MAX=32.
+        SETGROUPS => super::syscalls::creds::sys_setgroups(a0 as i32, a1),
+
         _ => -38i64, // ENOSYS
     }
 }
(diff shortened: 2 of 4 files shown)


dburkart:
@coderabbitai thanks — addressed in eb45a67. The round-trip test now drives SYS_SETGROUPS through the real dispatcher (using the buffer-less size==0 clear path so we don't need a user-half pointer), and a new `setgroups_grants_group_class_permission` test verifies the per-task Credential snapshot surfaces the new gid for the group-class permission check. Full end-to-end via populated user buffer + ext2/pjdfstest will land when the userspace libc syscall stubs are wired in a later wave — kernel test code can't easily allocate in the lower canonical half required by copy_*_user.

PR #652, opened 2026-04-27
block: invariant assertion — no spin lock across block-I/O wait (#556)

### LINKED ISSUE #556: block: invariant assertion — no spin lock across block I/O wait
## Motivation

Implements RFC 0004 — Workstream C (Block buffer cache, wave 1). Normative invariant
from the RFC: "no spin lock is held across a block-I/O wait." Catching violations at
debug-build runtime prevents the whole class of "hold spinlock, wait on device, deadlock"
bugs that are devilish to reproduce after the fact.

## Work

- [ ] Add a per-CPU `DEBUG_HELD_SPINLOCKS: u32` counter, incremented/decremented by
      `SpinLock::lock`/`drop` in debug builds only (`#[cfg(debug_assertions)]`)
- [ ] `BufferHead.data` lock MUST be released before `sync_dirty_buffer` calls
      `device.write_at` — assertion point
- [ ] On every call that could block (e.g. `BlockDevice::write_at`, `read_at`, any
      `wait` primitive used by sync_dirty_buffer), assert `DEBUG_HELD_SPINLOCKS == 0`
- [ ] Same assertion at every `bread` internal wait point
- [ ] Panic message names both the held-lock site and the I/O site for triage
- [ ] Unit test: constructing a violation (holding a lock across a mock I/O) panics
      under `debug_assertions`; release builds do not pay the cost

## Out of scope

- Lockdep-style full dependency-graph tracking — Open Question in RFC, deferred.
- Asserting the same for non-block subsystems (signals, net, etc.) — out of scope here.

## Context

- Parent epic: #537
- RFC 0004 §Buffer cache (normative invariant "no spin across I/O"), §Lock ordering
- This issue is the audit tripwire for the rest of Workstream C, D, E

dburkart:
Closes #556

## Summary

Adds a debug-build tripwire for RFC 0004 §Buffer cache normative invariant: **"no spin lock is held across a block-I/O wait."** Holding a spinlock across a device wait can deadlock — the wait may take arbitrarily long, while contending tasks burn CPU and IRQs that touch the lock dead-spin on a CPU that has interrupts disabled.

### Mechanism (`kernel/src/debug_lockdep.rs`)

- A per-CPU `HELD_SPINLOCKS: u32` counter (single-CPU today; a natural seam to widen when SMP arrives).
- `inc_held_spinlocks` / `dec_held_spinlocks` / `assert_no_spinlocks_held(io_site)` are all `#[cfg(debug_assertions)]` with `#[inline]` no-op stubs in release builds — the optimiser deletes the calls entirely, so the wait paths pay nothing.
- A `SpinLock<T>` newtype, `#[repr(transparent)]` over `spin::Mutex<T>`, that bumps / decrements the counter on acquire / drop. `IrqLock` participates too (a held IRQ-masking spinlock is the worst possible thing to hold across an I/O wait).

### Wired call sites (`kernel/src/block/cache.rs`)

- `BlockCache.inner` and `BlockCache.dirty` migrated from `spin::Mutex` to the tracking `SpinLock`. These are the cache's hot spinlocks; the existing code already drops them before issuing device I/O (RFC 0004 §Buffer cache, OS-engineer B5 hazard) — the assertion is the loud tripwire.
- `assert_no_spinlocks_held("BlockCache::bread → device.read_at")` immediately before the populate-on-miss `read_at`.
- `assert_no_spinlocks_held("BlockCache::sync_dirty_buffer → device.write_at")` immediately before the writeback `write_at`.
- The writeback daemon flushes via `sync_fs` → `sync_dirty_buffer`, so the assertion above covers it transitively.

The panic message names both the I/O site and the held-lock count so triage is "look at the named caller" rather than "decode a generic deadlock."

## Test plan

- [x] Six new host unit tests (counter inc/dec, nested locks, `try_lock` semantics, panic-on-violation, guard deref).
- [x] `cargo xtask test` — 515/515 host tests pass; QEMU integration tests pass.
- [x] `cargo xtask smoke` — all 33 markers present.
- [x] `cargo fmt --all -- --check` — clean.

### DIFF
--- a/kernel/src/block/cache.rs
+++ b/kernel/src/block/cache.rs
@@ -68,9 +68,10 @@ use alloc::sync::Arc;
 use alloc::vec;
 use core::sync::atomic::{AtomicBool, AtomicU32, AtomicU8, Ordering};
 
-use spin::{Mutex, Once, RwLock};
+use spin::{Once, RwLock};
 
 use super::{BlockDevice, BlockError};
+use crate::debug_lockdep::{assert_no_spinlocks_held, SpinLock};
 
 /// Valid bit — the in-memory slab reflects the on-disk block as of its
 /// most recent read-back.
@@ -216,7 +217,7 @@ pub struct BlockCache {
     /// Cache body + CLOCK-Pro metadata. Guarded together so the
     /// eviction sweep sees a consistent view of residency + classes +
     /// hand.
-    inner: Mutex<CacheInner>,
+    inner: SpinLock<CacheInner>,
     /// Dirty-set mirror used by the writeback daemon (populated by
     /// `mark_dirty`, cleared by `sync_dirty_buffer`). Stored as keys,
     /// not `Weak` handles — the daemon re-looks-up the `Arc` out of
@@ -224,7 +225,7 @@ pub struct BlockCache {
     /// still flushes correctly. A separate mutex so the writeback
     /// daemon (future) can snapshot the dirty set without contending
     /// on the CLOCK-Pro critical section.
-    dirty: Mutex<BTreeSet<(DeviceId, u64)>>,
+    dirty: SpinLock<BTreeSet<(DeviceId, u64)>>,
     /// Cap beyond which eviction must start reclaiming. Observed by
     /// `bread` when deciding whether to run a CLOCK-Pro sweep.
     max_buffers: usize,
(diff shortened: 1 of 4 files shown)


dburkart:
Addressed CodeRabbit's parallel-test-interference finding in 7926995 — gated all counter-touching tests on a `std::sync::Mutex<()>` per the `gc_queue.rs` pattern. The reset now happens under the lock, and `Mutex` poisoning recovers via `into_inner` so the `should_panic` test doesn't deadlock siblings.

PR #660, opened 2026-04-27
sys: replace Credential::kernel() at VFS entry; flip vfs_creds default-on

### LINKED ISSUE #550: sys: replace Credential::kernel() at VFS entry; flip vfs_creds feature flag
## Motivation

Implements RFC 0004 — Workstream B (Credential enforcement, wave 1). This is the
terminal PR of Workstream B: walk every VFS syscall entry point in
`kernel/src/arch/x86_64/syscalls/vfs.rs` and replace `Credential::kernel()` with
`task.credentials.read().clone()`, then flip `vfs_creds` on in `Cargo.toml`. After
this lands, every DAC check throughout the kernel is driven by the real caller's
identity.

Must land **after** all of Workstream A's write-syscall PRs have merged — otherwise an
unauthenticated DAC-bypass window exists. The A-before-B ordering is enforced by the
CI grep check that each A issue wires up ("CI fails if `Task::credentials` is absent
from main"); flipping the flag before A lands would leave every new syscall running
without a credential check.

## Work

- [ ] Audit every syscall arm in `kernel/src/arch/x86_64/syscalls/vfs.rs` for `Credential::kernel()` or equivalent
- [ ] Replace with `task.credentials.read().clone()` (or snapshot an `Arc` and pass it through)
- [ ] Same sweep over `open`/`openat`/`stat`/`fstat`/`lstat`/`newfstatat`/`chdir`/`getcwd`/`mknod`/`read`/`write`/`fsync`/`lseek`/`getdents64`/`fcntl`/`dup3` — the already-wired syscalls
- [ ] Flip `vfs_creds = []` → default-on in `Cargo.toml` **in the same PR**
- [ ] Delete or mark `#[cfg(test)]` any remaining `Credential::kernel()` callers in kernel code
- [ ] Integration test: a task with `ruid != 0` cannot chmod a root-owned file (returns `EPERM`)
- [ ] Grep invariant: CI fails if `Credential::kernel()` appears outside `#[cfg(test)]` or an explicit `// AUDITED` comment after this lands

## Out of scope

- `CAP_*` capability checks — explicit non-goal.
- Auditing non-VFS syscalls (signals, process, net) — separate follow-up if needed.

## Context

- Parent epic: #537
- Depends on: #546, #547, #548, #549 (full credential surface), plus all of Workstream A (#538–#545)
- RFC 0004 §Credential model, §Security Considerations

dburkart:
Closes #550

## Summary

Terminal PR of RFC 0004 Workstream B. Walks every userspace-driven VFS syscall entry in `kernel/src/arch/x86_64/syscalls/vfs.rs` and replaces the `Credential::kernel()` placeholder with the caller's per-task credential snapshot via `crate::task::current_credentials()`. Then flips the `vfs_creds` feature on by default in `kernel/Cargo.toml` so the chmod / chown / utimensat / setgroups / faccessat / unlink / link / symlink / truncate dispatch arms wired by Workstream A become reachable from ring-3.

After this lands, every DAC check throughout the kernel runs against the real caller's identity.

### Sites updated in `vfs.rs`

- `resolve_inode()` — read-path resolver used by `open`/`openat`/`stat`/`fstat`/`lstat`/etc. now walks under the caller's cred, so an unsearchable ancestor fails with `-EACCES` instead of silently resolving as root and tripping a misleading errno downstream.
- `mkdir` DAC check on the parent directory.
- `unlinkat` sticky-bit check (`S_ISVTX`) — uses `euid` for ownership comparison, matching POSIX.
- `linkat` manual NOFOLLOW walk + DAC check on the new parent.
- `symlinkat` DAC check on the new parent.
- `readlinkat` path walk.

### Off-branch tests preserved

The `vfs_creds = []` feature gate stays in place so the off-branch regression tests in `kernel/tests/syscall_*.rs` (chmod, chown, truncate, utimensat, faccessat, unlink, link/symlink) continue to verify that the dispatcher returns `-ENOSYS` when the feature is explicitly disabled with `--no-default-features`. The on-branch arms of those same `dispatch_gate_enosys` helpers exercise the now-default-on path.

### CI gate

The `vfs-creds-gate` CI job already accepts a default-on `vfs_creds` when `Task::credentials` is present in the tree (added in #590). The gate's "Task::credentials present → satisfied" branch fires immediately and lets the default-on flip through.

## Test plan

- [x] `cargo xtask build` clean.
- [ ] CI green: `vfs-creds-gate`, `fmt-build`, `kernel-tests` (host + QEMU integration).
- [ ] Targeted `syscall_chmod_chown` integration test passes locally (ran clean before push).
- [ ] No `Credential::kernel()` callers remain in `kernel/src/arch/x86_64/syscalls/vfs.rs` outside doc-comments.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -24,7 +24,7 @@ pic8259 = "0.11"
 pc-keyboard = "0.8"
 
 [features]
-default = []
+default = ["vfs_creds"]
 # Enable to trigger a UD2 after init; verifies the invalid-opcode handler path.
 fault-test = []
 # Enable to fire a deliberate `panic!()` after init; useful for eyeballing
@@ -49,13 +49,15 @@ bench = []
 # never ship it. Enable with `cargo xtask run --features fork-trace` (or
 # `--features fork-trace` on any xtask subcommand).
 fork-trace = []
-# Workstream A ↔ B gate (RFC 0004). When off, POSIX write-side VFS
-# syscalls added by Workstream A (mkdir/mkdirat today; rmdir/unlink/…
-# in sibling PRs) compile to `-ENOSYS`. Workstream B flips this on in
-# the same PR that lands `Task::credentials`, so per-task credentials
-# are always available at every DAC-sensitive call site. Keeps a
-# partially-landed Workstream A from exposing an unauthenticated
-# DAC-bypass window.
+# Workstream A ↔ B gate (RFC 0004). On by default as of #550 — every
+# VFS syscall entry now reads the caller's per-task credential
+# snapshot via `crate::task::current_credentials()` (no more
+# `Credential::kernel()` fallback at any userspace-driven call site).
+# Disable explicitly with `--no-default-features` to compile the
+# Workstream A write-side syscalls back out to `-ENOSYS`; this is
+# only useful for re-running the dispatch-gate regression tests in
+# the `kernel/tests/syscall_*.rs` files that assert the off-branch
+# behaviour. Production builds always run with the feature on.
 vfs_creds = []
 # RFC 0004 Workstream D/E feature gate. When off, the ext2 on-disk
 # types (`fs::ext2::disk`) still compile (they're pure data, host unit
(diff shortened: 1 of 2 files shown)


dburkart:
@coderabbitai review

PR #661, opened 2026-04-27
fs/ext2: trigger orphan_finalize from last-fd-close (#638)

### LINKED ISSUE #638: fs/ext2: wire production trigger for orphan_finalize (OpenFile drop → evict_inode)
Follow-up of #573 (merged as da481d1).

## Motivation

`fs::ext2::orphan_finalize::finalize` (exposed as `finalize_orphan`) implements the RFC 0004 §Final-close sequence and is wired into `Ext2Super::evict_inode`, so the VFS `gc_queue` drainer will run it when the last `Arc<Inode>` for an orphan ino drops. Today that never actually happens in-kernel during a live mount because the per-super `orphan_list` keeps a strong `Arc<Inode>` pin — `Inode::Drop` can't fire until that pin goes away, and the pin only goes away *inside* `finalize`. So the current drivers of `finalize_orphan` are:

- mount-time orphan-chain replay (#564)
- integration tests (`kernel/tests/ext2_orphan_finalize.rs`)

A real production trigger needs an in-kernel last-fd-close path that lets the kernel finalize an orphan without waiting for umount/remount. RFC 0004 envisions a per-`OpenFile` drop that decrements a driver-side refcount and calls `finalize_orphan` when the count hits zero for an unlinked inode.

## Work items

- Design a driver-side "outstanding opens" refcount (probably on `Ext2Inode`, bumped by `Ext2Fs::open` / decremented by the `OpenFile` drop hook).
- Add the drop hook path (may need a small VFS surface change: `SuperOps::release_file` or `FileOps::release`).
- When the count hits zero **and** the ino is in `orphan_list`, call `finalize_orphan` directly (not via the `gc_queue`/`evict_inode` indirection, since the `Arc<Inode>` pin from `orphan_list` would still block eviction).
- Drop the `Arc<Inode>` orphan_list pin inside `finalize_orphan` (already happens) so subsequent VFS-cache drops are normal no-ops.
- Add an integration test that exercises the full `open → unlink → close → evict` sequence without unmounting.

## Context

- `kernel/src/fs/ext2/orphan_finalize.rs` — module docs have a "Wiring the production trigger" section that restates the above.
- `kernel/src/fs/ext2/fs.rs::Ext2Super::evict_inode` — current wire point; docs updated post-review to reflect the unchain-before-free ordering.
- RFC 0004 §Final-close sequence — normative spec.

Not urgent: until this lands, orphan inodes still get cleaned up either by an explicit `finalize_orphan` call (tests) or by mount-time replay on the next boot. But long-lived mounts with heavy unlink-while-open traffic will leak storage until remount.

Labels: `area:fs`, `track:filesystem`, `priority:P2` (quality-of-life, not correctness).

dburkart:
Closes #638.

## Summary

- Adds `FileOps::open` / `FileOps::release` lifecycle hooks to the
  VFS `FileOps` trait (default no-op). `OpenFile::new` calls `open`
  after the SB pin is migrated into `dentry_pin_count`;
  `OpenFile::Drop` calls `release` *before* dropping the SB pin so
  release-time finalize logic still sees a pinned super.
- Wires ext2's outstanding-opens refcount on `Ext2Inode`
  (`open_count: AtomicU32`). On the 1->0 release transition where
  `unlinked == true`, ext2 calls `orphan_finalize::finalize` directly.
  Resolves the chicken-and-egg in RFC 0004's orphan-list residency
  invariant: the orphan_list `Arc<Inode>` pin kept the inode alive
  forever, so VFS gc never reached `evict_inode` for orphans, so the
  pin was never released, and so on.
- Race fix in `unlink.rs` and `rename.rs`: `push_on_orphan_list` now
  happens *before* `unlinked.store(true)`. Previously a concurrent
  releaser could observe `unlinked == true` between the store and
  the push, finalize-with-ENOENT, and silently leak the orphan when
  the push completed.
- Updates `orphan_finalize` module doc to list the three production
  triggers (release hook #638, mount-time replay #564, evict_inode
  passthrough #573).

## Test plan

- [x] New `ext2_orphan_open_close` integration test:
  - `open_unlink_close_finalizes_orphan` -- holds an OpenFile across
    rmdir, asserts orphan pin survives, drops OpenFile, asserts
    orphan_list empty + `s_last_orphan == 0` + free counters bumped
    + on-disk tombstone (`i_links_count == 0`, `i_dtime != 0`).
  - `rmdir_with_no_openers_keeps_orphan_pinned` -- control case;
    confirms the trigger is open-count driven, not synchronous on
    rmdir.
- [x] `cargo xtask build` clean (only pre-existing dead-code warning).
- [x] `cargo xtask test` green (one pre-existing wait4_condvar_race
      flake on retry, unrelated; passes on second run).
- [x] `cargo xtask smoke` -- all 35 markers present.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -447,6 +447,11 @@ harness = false
 required-features = ["ext2"]
 
 [[test]]
+name = "ext2_orphan_open_close"
+harness = false
+required-features = ["ext2"]
+
+[[test]]
 name = "ext2_rename"
 harness = false
 required-features = ["ext2"]
--- a/kernel/src/fs/ext2/inode.rs
+++ b/kernel/src/fs/ext2/inode.rs
@@ -63,7 +63,7 @@
 use alloc::collections::BTreeMap;
 use alloc::sync::{Arc, Weak};
 use alloc::vec;
-use core::sync::atomic::AtomicBool;
+use core::sync::atomic::{AtomicBool, AtomicU32, Ordering};
 
 use spin::Mutex;
 
@@ -174,6 +174,20 @@ pub struct Ext2Inode {
     /// the surface exists so the unlink path and mount-time orphan
     /// replay (#564) have a well-known place to set it.
     pub unlinked: AtomicBool,
+    /// Driver-side outstanding-opens refcount. Bumped by
+    /// [`FileOps::open`] (one per successful `OpenFile::new`) and
+    /// decremented by [`FileOps::release`] (one per `OpenFile::Drop`).
+    /// When [`unlinked`](Self::unlinked) is set and this count
+    /// transitions from one to zero, the `release` hook calls
+    /// [`super::orphan_finalize::finalize`] directly to drive the
+    /// RFC 0004 §Final-close sequence — bypassing the VFS
+    /// `gc_queue`/`evict_inode` indirection that would otherwise be
+    /// blocked by the orphan-list `Arc<Inode>` pin (chicken-and-egg:
+    /// the pin is held by `orphan_list` and only released **inside**
+    /// `finalize`, so `Inode::Drop` for an orphan can never fire to
+    /// trigger eviction). See issue #638 / RFC 0004 §Wiring the
+    /// production trigger.
+    pub open_count: AtomicU32,
 }
 
 impl Ext2Inode {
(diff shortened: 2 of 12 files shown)


dburkart:
@coderabbitai review

PR #664, opened 2026-04-28
RFC 0005: Scheduler / IRQ Seam for Deterministic Simulation Testing

dburkart:
## Summary

RFC 0005 proposes extracting the scheduler core in `kernel/src/task/` from its direct dependence on PIT/HPET/APIC by routing time and IRQ-ack through two minimal injected traits — `Clock` and `IrqSource` — defined in a new `kernel/src/task/env.rs`. Production wires them to existing drivers via zero-sized adapters (no behavior change); a future host-side simulator (Phase 2, out of scope here) supplies seeded implementations to convert flaky concurrency bugs (#501, #527) into reproducible seeds.

This PR contains a design document only — no kernel code changes. The RFC will be reviewed by the four default archetypes (security researcher, OS engineer, user space staff engineer, academic), revised over up to 3 defense cycles, then merged. Implementation issues will be filed from the roadmap section after merge.

Source discussion: #394.

## RFC sections

- Abstract
- Motivation
- Background (incl. prior-art table — TigerBeetle VOPR, FoundationDB, Linux clocksource/clockevents, loom, commonware-runtime, Tokio test clock, Polar Signals DST, Hermit)
- Design (overview, exact trait shapes, production wire-up, generics-vs-trait-objects decision, test-time wiring, migration plan)
- Security Considerations
- Performance Considerations (hot-path cost analysis)
- Alternatives Considered (wider trait surface, generics, wait-for-SMP, loom-style cfg seam, abstract-everything)
- Open Questions (LAPIC-timer migration, per-CPU Clock, SMP, lifetime of trait objects)
- Implementation Roadmap (7 dependency-ordered, independently-landable items)

## Test plan

- [ ] All four scoped archetype reviewers post a review comment.
- [ ] All blocking findings addressed within 3 defense cycles.

### DIFF
--- a/docs/RFC/0005-scheduler-irq-seam.md
+++ b/docs/RFC/0005-scheduler-irq-seam.md
@@ -1,7 +1,7 @@
 ---
 rfc: 0005
 title: Scheduler / IRQ Seam for Deterministic Simulation Testing
-status: In Review
+status: Accepted
 created: 2026-04-27
 ---
 
@@ -491,8 +491,8 @@ incrementally and impossible to bisect when (not if) it regresses.
 
 ## Open Questions
 
-These are explicitly flagged for resolution during implementation, not
-before merge:
+All open questions are deferred to implementation per the resolutions
+recorded below. None block RFC acceptance.
 
 - **`Clock` resolution after LAPIC-timer migration.** Today `Tick` is PIT
   ticks (10 ms). When the kernel migrates to LAPIC-timer (likely as part
--- a/docs/RFC/0005-scheduler-irq-seam.md
+++ b/docs/RFC/0005-scheduler-irq-seam.md
@@ -1,7 +1,7 @@
 ---
 rfc: 0005
 title: Scheduler / IRQ Seam for Deterministic Simulation Testing
-status: Draft
+status: In Review
 created: 2026-04-27
 ---
 
--- a/docs/RFC/0005-scheduler-irq-seam.md
+++ b/docs/RFC/0005-scheduler-irq-seam.md
@@ -1,0 +1,403 @@
+---
+rfc: 0005
+title: Scheduler / IRQ Seam for Deterministic Simulation Testing
+status: Draft
+created: 2026-04-27
+---
+
+# RFC 0005: Scheduler / IRQ Seam for Deterministic Simulation Testing
+
+## Abstract
+
+Extract the scheduler core in `kernel/src/task/` from its direct dependence on
+the PIT timer counter, the LAPIC, and the global tick wakeup table by routing
+those calls through two minimal injected dependencies: a `Clock` trait
+(monotonic time + future-tick scheduling) and an `IrqSource` trait (interrupt
+acknowledgement + drain). Production builds wire these to the existing
+PIT/HPET/APIC drivers with zero observable behavior change; tests can supply
+mocks; a future host-side simulator (out of scope here) can supply seeded
+implementations to turn today's flaky concurrency bugs into reproducible
+seeds. This RFC covers Phase 1 only — the seam, not the simulator.
+
+## Motivation
+
+Two concrete pains motivate the seam now:
+
+- The long-running fork/exec/wait quality sprint (#501) and the residual
+  ring-3 user-stack `#PF` flake (#527) are both *intermittent* — the current
+  toolchain can only retry-and-hope. Every retry burns ~6–8 hours of CI
+  wallclock under `-Zbuild-std` + QEMU. Without a deterministic harness, a
+  flake reproduced once may not reproduce again, and root-causing happens by
+  staring at logs. A seeded simulator turns each flake into a `cargo test
+  -- --seed=0xDEADBEEF` — but only if the scheduler can be driven without
+  real hardware.
+- The SMP track (#30, #88–#91) will refactor the scheduler regardless. Doing
+  the testability seam *first* gives us a deterministic regression harness
+  before SMP starts churning the same code, rather than landing two
+  destabilizing rewrites back-to-back.
+
+The proposal here is a *refactor*, not a new feature. The end state for
+Phase 1 is identical kernel behavior, the same passing test suite, plus two
+new trait definitions and a single production wire-up site.
+
+## Background
+
+Today the scheduler is in `kernel/src/task/` (`mod.rs`, `scheduler.rs`,
+`switch.rs`, `task.rs`, `softirq.rs`, `priority.rs`). The PIT ISR calls
+`preempt_tick()` in `task::mod`, which:
+
+1. Reads `crate::time::ticks()` (a global `AtomicU64` advanced by the PIT
+   IRQ handler) to compute deadlines.
+2. Drains `crate::time::drain_expired(ticks())`, a global `BTreeMap<u64,
+   Vec<TaskId>>` of pending wakeups guarded by an `IrqLock`.
+3. Acks the IRQ via the architecture-specific path (PIC/LAPIC EOI in
+   `kernel/src/arch/x86_64/{pic,apic}.rs`).
+4. Hands control to the priority-aware ready bank
+   (`task/scheduler.rs::Scheduler::pop_highest`).
+
+Time and IRQ are reached through *crate-global* statics today —
+`time::TICKS`, `time::WAKEUPS`, the LAPIC MMIO `Once<LocalApic>` —
+which means there is no in-process way for a test to advance time
+without going through real hardware.
+
+### Prior art surveyed
+
+| Source | Lesson taken |
+|---|---|
+| TigerBeetle VOPR | Three traits — `Time`, `Network`, `Storage` — keep production and simulator on identical code paths. Discipline: *no globals* for these capabilities. |
+| FoundationDB simulator (Flow) | Determinism only works if *every* nondeterminism source goes through the runtime. The cost of one escape hatch is the entire harness. |
+| Linux clocksource / clockevents | Mature precedent for splitting "read time" (clocksource) from "set next event" (clockevent_device). Rating-based runtime selection across TSC/HPET/LAPIC validates that the seam shape proposed here is well-trodden. |
+| Tokio `loom` | cfg-gated type-aliasing in a project-local `sync` module is the lowest-friction Rust pattern; loom replaces `Mutex`, `Atomic*`, `thread::spawn`, `UnsafeCell`. Production code is unchanged. |
+| commonware-runtime (Rust) | Trait surface = `Clock` + `Spawner` + `Network` + `Storage` — confirms "minimum 2–3 traits" is sufficient for an async runtime DST harness. |
+| S2.dev / Tokio test clock | Time advances on `sleep()` calls; clocks include a "schedule event for the future" method, not just `now()`. Validates `Clock::schedule_tick(deadline)` belongs in the trait. |
+| Polar Signals (state-machine DST) | A maximalist "everything is a state machine" approach has high cognitive overhead and slows engineers. Argues for *not* abstracting more than the harness needs. |
+| Hermit (Microsoft Research) | Time, RNG, scheduling, and syscalls are the four nondeterminism sources. Phase 1 covers time + scheduling; RNG and syscall-replay are deferred. |
+
+## Design
+
+### Overview
+
+Two traits, defined in a new file `kernel/src/task/env.rs`. The scheduler
+core takes them by `&dyn` reference at the call boundary; production wires
+them to existing drivers via a single `static` per trait.
+
+```rust
+// kernel/src/task/env.rs
+
+/// Monotonic tick count since boot. Opaque newtype so we can swap the
+/// underlying width or units without ABI churn.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Debug)]
+pub struct Tick(pub u64);
+
+/// Source of monotonic time and one-shot future tick wakeups for the
+/// scheduler. The scheduler never reads a clock register or arms a timer
+/// directly — it goes through this trait.
+pub trait Clock: Sync {
+    /// Current monotonic tick count.
+    fn now(&self) -> Tick;
+
+    /// Drain all wakeup ids whose deadline is `<= now`. The implementor
+    /// owns the deadline structure (today: `time::WAKEUPS`); the
+    /// scheduler only consumes the drained ids.
+    fn drain_expired(&self, now: Tick) -> alloc::vec::Vec<usize>;
+
+    /// Enqueue task `id` for a wakeup at `deadline`. Idempotent on
+    /// `(deadline, id)` pairs.
+    fn enqueue_wakeup(&self, deadline: Tick, id: usize);
+}
+
+/// Source of preemption interrupts for the scheduler. Only the
+/// per-tick-hot-path operations live here; one-shot IDT install,
+/// IOAPIC redirection, and IPI send are *not* part of this trait
+/// (see Alternatives Considered).
+pub trait IrqSource: Sync {
+    /// Acknowledge the timer IRQ that drove the current `preempt_tick`
+    /// call. Today: LAPIC EOI write; on legacy PIC: PIC EOI.
+    fn ack_timer(&self);
+}
+```
+
+That is the entire surface for Phase 1. Two traits, four methods.
+
+### Production wire-up
+
+In `kernel/src/task/mod.rs`:
+
+```rust
+// Production-only: zero-sized adapters over the existing globals.
+struct HwClock;
+impl Clock for HwClock {
+    fn now(&self) -> Tick { Tick(crate::time::ticks()) }
+    fn drain_expired(&self, now: Tick) -> Vec<usize> {
+        crate::time::drain_expired(now.0)
+    }
+    fn enqueue_wakeup(&self, deadline: Tick, id: usize) {
+        crate::time::enqueue_wakeup(deadline.0, id)
+    }
+}
+
+struct HwIrq;
+impl IrqSource for HwIrq {
+    fn ack_timer(&self) { crate::arch::ack_timer_irq(); }
+}
+
+static HW_CLOCK: HwClock = HwClock;
+static HW_IRQ: HwIrq = HwIrq;
+
+/// Returns the production environment. Tests in `cfg(not(target_os =
+/// "none"))` builds and the future simulator override this.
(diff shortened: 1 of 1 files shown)


dburkart:

**Summary:** The seam is kernel-internal and does not change the privilege boundary. The Security Considerations section addresses the obvious surface (no new MMIO, no new IDT entries, no userspace exposure) honestly. One concrete blocker around `cfg`-gated mock code reachability, plus a couple of advisory items.

**Blocking findings:**
- [B1] Mock-impl `cfg` gate is too loose — `#[cfg(any(test, not(target_os = "none")))]` is the gate proposed for `MockClock` / `MockIrqSource`. The first half (`test`) is fine — `cfg(test)` excludes from `--release` *and* from non-test target builds. The second half (`not(target_os = "none")`) only excludes from kernel builds; it does *not* exclude from any host-target build of the kernel crate (e.g. `cargo build -p kernel --target x86_64-unknown-linux-gnu`), and the kernel crate is built host-side today for unit tests. That's fine for tests, but the RFC needs to commit that mocks are gated behind a marker that *also* excludes them from `cargo xtask build --release` for the kernel target — and that the gate is verified, not just asserted. Concrete ask: add a `#[cfg(any(test, feature = "sched-mock"))]` style gate (no `target_os` exception), and add a CI step that runs `nm` over the release kernel binary and fails if it contains `MockClock` symbols. Otherwise the "physically excluded from production" claim is unverified.

**Advisory findings:**
- [A1] Trait method auditability — `IrqSource::ack_timer` having no return value or error path means a buggy mock that silently no-ops in production (e.g. via the wrong `env()` wire-up at boot) would manifest as a wedged scheduler rather than a panic. Consider `#[must_use]` on the trait or a `debug_assert!` in production wire-up that confirms the production `env()` was called before the first IRQ. Not blocking — the wire-up is a single boot site.
- [A2] `Tick(u64)` with `pub` field — exposing the field publicly invites callers outside the scheduler to do tick arithmetic that bypasses the `Clock` trait. Make it `pub(crate)` or wrap arithmetic in `impl Tick { fn checked_add(self, other: u64) -> Self }`. Cheap hygiene that prevents the seam from leaking back into ad-hoc time math.
- [A3] `drain_expired` returning `Vec<usize>` allocates on every tick — not a security concern per se, but allocation in the IRQ-tail-context path is a documented pattern that has bitten kernel security (allocator reentrancy). Worth flagging that the allocation already happens today inside `time::drain_expired`; the trait is not introducing it, but it freezes the contract.

dburkart:
### Review: OS Engineer

**Summary:** The proposed seam shape is sound and matches well-trodden precedent (Linux clocksource/clockevents). I have substantive concerns about IRQ-context safety in the trait contract, the `env()` accessor's interaction with early-boot code, and the under-specification of *which* IRQ-ack site is being captured (LAPIC vs PIC vs spurious). The `&dyn` decision is correct for v1.

**Blocking findings:**
- [B1] IRQ-context contract is unspecified — `Clock::drain_expired`, `Clock::enqueue_wakeup`, and `IrqSource::ack_timer` are all called from interrupt context (the PIT/LAPIC ISR) *and* from task context (`sleep_ms` enqueues from a syscall). The trait does not say which methods are IRQ-callable, what their reentrancy contract is, or whether implementors must mask IRQs. Today `time::WAKEUPS` is an `IrqLock<BTreeMap>` for exactly this reason. The trait needs explicit doc-comments per method stating "callable from IRQ context" / "must be IRQ-safe" / "implementor must mask IRQs internally." Without that, a future `Clock` impl could deadlock on the first PIT IRQ that lands while `sleep_ms` holds the lock — the exact bug `IrqLock` was created to prevent.
- [B2] `env()` accessor before scheduler init — `env()` returns `&'static dyn Clock` over zero-sized statics, which works at any point in boot. But several boot-phase callers in `kernel/src/main.rs` call `time::ticks()` *before* `task::init()` runs. The RFC needs to specify whether `env()` is callable pre-`task::init` (it must be — calibration uses `time::ticks` during boot) and what happens if a future non-ZST `Clock` impl assumes init-order. Concrete ask: the trait contract must say "implementors must be safe to call from any point after global allocator init" or the RFC must enumerate which call sites are pre-task-init and exempt them.
- [B3] Step-3 of the migration plan is not actually CI-green — "Migrate `preempt_tick` and the IRQ-ack path … existing scheduler tests + fork/exec/wait smoke must stay green" — but the same step also tightens `time::*` to `pub(crate)` (per step 5). If step 3 lands without step 5, callers outside `task/` (signal delivery uses `time::ticks` per `kernel/src/signal/mod.rs` from the grep earlier) are temporarily reading time *both* through the seam and directly, with no enforcement. Either fold steps 3+5 into one item, or have step 3 add a `#[deprecated]` annotation on the `time::*` functions to surface stragglers in CI warnings. The roadmap as written can leave the tree in a half-migrated state for an arbitrary length of time.

**Advisory findings:**
- [A1] `IrqSource::ack_timer` only — the trait covers the timer IRQ but not the spurious-vector ack, the IPI-ack (future), or device IRQ acks (PIC/IOAPIC). Naming the trait `IrqSource` while it only handles the timer is misleading. Either rename to `TimerIrq` / `PreemptionIrq` (more honest scope), or document explicitly that "this trait is the *scheduler's* view of IRQs; device drivers continue to ack their own IRQs through `arch::*` directly." Strong preference for the rename.
- [A2] `enqueue_wakeup` taking `usize` for the task id — task ids are a typed concept elsewhere in `task/`. Use a typed `TaskId` newtype here so a future refactor (e.g. ids becoming generation-counted) doesn't silently break the trait.
- [A3] No mention of the `softirq` path — `kernel/src/task/softirq.rs` exists and the RFC doesn't say whether softirq processing goes through the seam or remains direct. If softirqs run "after IRQ ack but before scheduler dispatch", the trait may need a `post_ack_hook` or the softirq path needs to be explicitly out-of-scope. Worth one sentence either way.
- [A4] Performance claim "≲ 1 µs/sec at 100 Hz" is true but the *real* hot path under DST will be the simulator stepping the scheduler millions of ticks per simulated minute. At sim-time, the indirect call cost is paid per simulated tick, not per wall-clock tick. That's fine — the simulator doesn't care — but the RFC's perf section reads as if the production cost is the only cost worth analyzing. Add one sentence acknowledging the simulator-time amortization is what makes `&dyn` not a problem there either.

dburkart:

**Summary:** The RFC is explicit that this is a kernel-internal refactor with no syscall / `/proc` / `/sys` impact, which I confirm by reading the design — nothing here is reachable from userspace. With no userspace surface, my review is about the *downstream* userspace impact: does the seam constrain or enable POSIX semantics we'll need later? Mostly enables, with one concern about `CLOCK_MONOTONIC` semantics under simulation.

**Blocking findings:**
None — no userspace-visible interface is changing.

**Advisory findings:**
- [A1] `Clock::now()` returns `Tick`, not `ns` or a `Duration`. POSIX userspace will eventually want `clock_gettime(CLOCK_MONOTONIC, ...)` returning `struct timespec` (nanoseconds). The Open Question about `now_ns()` notes this, but the design today routes the entire scheduler through `Tick`-granularity time. When the LAPIC-timer migration adds `now_ns()`, will `clock_gettime` resolve to `clock.now_ns()` directly (good — automatically deterministic under simulation) or to a separate `time::now_ns()` global (bad — bypasses the seam)? Worth committing in the RFC: "all userspace-visible time eventually flows through `Clock`." Otherwise we ship a deterministic scheduler with non-deterministic `gettimeofday` and discover it during simulator bring-up.
- [A2] `sleep_ms` is the only userspace-facing-equivalent path migrated in the roadmap. POSIX has many more time-yielding syscalls — `nanosleep`, `clock_nanosleep`, `pselect` timeout, `ppoll` timeout, `epoll_wait` timeout, `futex` with timeout, `sigtimedwait`. Each becomes a separate seam-crossing later. The RFC implicitly assumes they all funnel through `enqueue_wakeup` once they exist. That's likely true (it's how Linux organizes hrtimer-based timeouts), but worth one sentence: "all timeout-bearing syscalls will route their wakeups through `Clock::enqueue_wakeup`; this RFC does not introduce them but commits that they will land here."
- [A3] No mention of `CLOCK_REALTIME` vs `CLOCK_MONOTONIC` semantics. The seam is monotonic-only (per `Tick` definition). Wall-clock time (CMOS RTC, eventually NTP) is correctly not in the seam. Worth a one-line "wall-clock time is intentionally out of scope; only monotonic time crosses the scheduler boundary."
- [A4] `Tick(pub u64)` — agree with the security reviewer that this should be `pub(crate)` field; userspace-facing time conversions should go through methods, not `Tick(deadline.0 + 1)` ad-hoc arithmetic.

**Verdict:** LGTM

dburkart:
### Review: Academic

**Summary:** The RFC's framing — "minimum viable seam now, grow when the harness demands it" — is well-grounded in the cited prior art (loom, commonware-runtime, TigerBeetle, Polar Signals' cautionary tale). The two-trait factoring matches Linux's mature clocksource/clockevents split, which is strong evidence it's the right primitive decomposition. I have one blocking concern about a missing formal property the RFC implicitly assumes, plus advisory items on prior art coverage.

**Blocking findings:**
- [B1] The "no behavior change" claim is unprovable as stated — the RFC says production wire-up has zero observable behavior change, but provides no formal argument for *why*. The actual property required is: "the composition `(HwClock, HwIrq)` is observationally equivalent to direct `time::*` / `apic::*` calls when the scheduler is the sole client." That property holds *only if* (a) `HwClock` methods are pure forwarders with no added state, and (b) the `env()` accessor has no synchronization that direct calls didn't have. (a) is true by construction; (b) is *not* discussed — `env()` returns `&'static` references, which is lock-free, but the RFC should explicitly state the observational-equivalence property and its proof obligation. Without it, "no behavior change" is a hope, not a guarantee, and we have no test that catches a regression. Concrete ask: add an "Equivalence" sub-section to the Design that states the property, names the two conditions, and commits one of (i) a small differential test that runs the existing scheduler test suite with both code paths and asserts identical sequences, or (ii) an audit checklist the implementing PR must satisfy. Linux's clocksource framework had the same problem and added the `clocksource_verify_percpu` self-test for this reason.

**Advisory findings:**
- [A1] Prior art coverage is strong but missing one directly relevant academic citation: **Hyperkernel** (Nelson et al., SOSP 2017, "Hyperkernel: Push-Button Verification of an OS Kernel"). Hyperkernel's verified design hinges on factoring time and IRQ exactly as proposed here, so they could symbolically execute the kernel without modeling hardware. The seam shape proposed is a (probably-rediscovered) instance of their interface — citing it strengthens the "well-trodden" defense against academic challenge that the design is ad-hoc. Also worth a citation: **Jitk** (Wang et al., OSDI 2014) for the related discipline of separating policy from mechanism in kernel components.
- [A2] The `Tick(u64)` → eventual `now_ns()` evolution is a quiet violation of the Liskov substitution principle in the small (an impl that returns finer-grained ticks is *not* substitutable for one that returns 10 ms ticks under all client expectations). The Open Question acknowledges this as an evolution but does not commit to whether `Tick`'s *units* are part of the trait contract. Lean: yes, they should be — `Tick` units must be uniform across all `Clock` impls within a single boot. Document this as an invariant.
- [A3] Polar Signals' cautionary tale is cited correctly but the RFC could go further: state the *positive* discipline (what we *will* do to avoid the over-abstraction failure mode) rather than just the negative (what we *won't* do). Concretely: "we add a trait method only when a second `Clock` or `IrqSource` impl needs it, never speculatively." This is the loom discipline and worth committing in writing so future PRs can be pushed back on with a citation.
- [A4] The RFC does not discuss the **equivalence to a labeled transition system** that the simulator will eventually need. That's appropriately Phase-2 scope, but worth flagging in Open Questions: "Phase 2 will require the scheduler's externally-observable behavior to be expressible as an LTS over `(Clock-event, IrqSource-event) → state transition`. The seam shape here is necessary but possibly not sufficient — the Phase-2 RFC should validate this before relying on the v1 trait surface." One sentence suffices.

dburkart:
## Defense cycle 1 — response

Pushed `4d34ba2` addressing every blocking finding. Summary by archetype:

**Security Researcher**
- [B1] **Addressed.** `cfg` gate tightened to `#[cfg(any(test, feature = "sched-mock"))]` (no `target_os` exception). The `sched-mock` feature is non-default; `cargo xtask build --release` for the bare-metal target builds with default features only. New roadmap item adds a CI step that runs `nm` on the release ELF and fails on `MockClock|MockIrqSource` symbols — verified, not asserted.
- [A1] *Addressed.* Roadmap item 2 now includes a `debug_assert!` in `task::init()` confirming `env()` returns the production wire-up before the first scheduler dispatch.
- [A2] *Addressed.* `Tick` field is now `pub(crate)` with `checked_add`/`saturating_add`/`raw` methods.
- [A3] *Addressed.* The `drain_expired` doc comment now documents the IRQ-tail allocation explicitly and notes that the contract preserves (rather than introduces) the existing `time::drain_expired` allocation pattern.

**OS Engineer**
- [B1] **Addressed.** Both traits now carry an explicit "IRQ-context contract" doc section: every method is marked IRQ-callable, implementors must mask IRQs internally, and the `IrqLock` precedent is cited. A non-IRQ-safe `Clock` impl is called out as a soundness bug.
- [B2] **Addressed.** `Clock` trait now carries an "Init-order contract" — impls must be safe to call any time after global allocator init (earlier than `task::init`), production `HwClock` is ZST and trivially satisfies this, non-ZST impls (mock, simulator) must initialize lazily or have boot install them before the first call.
- [B3] **Addressed.** Roadmap restructured: the call-site migration *and* the `pub(crate)` tightening are now a single PR (item 3), so the tree is never half-migrated. Signal-delivery is named as the known straggler. The differential test runs in this PR.
- [A1] *Addressed.* New roadmap item adds a final mechanical rename `IrqSource → TimerIrq` (deferred to last so earlier PRs don't churn). Trait doc also clarifies the scope: "scheduler's view of IRQs only; device drivers continue to ack their own IRQs through `arch::*` directly."
- [A2] *Addressed.* Trait now uses `TaskId` newtype (re-exported from `task::TaskId`) instead of bare `usize`.
- [A3] *Addressed.* New Open Question on softirq ordering relative to the seam — Phase 1 leaves softirq dispatch direct on the grounds that it's a kernel-internal optimization, with Phase-2 LTS equivalence flagged separately.
- [A4] *Addressed.* Performance section gains a "Simulator-time hot path" paragraph; bottleneck under DST is `BTreeMap` ops on the wakeup table, not trait dispatch.

**User Space Staff Engineer** — was already LGTM. All four advisories folded in:
- [A1] New Open Question commits "all userspace-visible time eventually flows through `Clock`."
- [A2] Same Open Question explicitly enumerates the timeout-bearing POSIX syscalls and commits they will route through `Clock::enqueue_wakeup`.
- [A3] Same Open Question marks `CLOCK_REALTIME` / wall-clock as intentionally out-of-scope.
- [A4] `Tick` field privacy fix above resolves this.

**Academic**
- [B1] **Addressed.** New "Equivalence (production wire-up correctness)" subsection in Design states the observational-equivalence property formally, names the three conditions (pure forwarding, no added synchronization, identical IRQ-context safety), and commits both an audit checklist and a differential test (byte-identical QEMU serial-log diff before/after the migration commit). The Linux `clocksource_verify_percpu` precedent is cited.
- [A1] *Addressed.* Hyperkernel (Nelson et al., SOSP 2017) and Jitk (Wang et al., OSDI 2014) added to the prior-art table with their relevance noted.
- [A2] *Addressed.* `Tick` doc comment carries an explicit "Unit invariant" section: `Tick` units must be uniform across all `Clock` impls within a single boot; mixed units is a correctness bug.
- [A3] *Addressed.* New "Discipline" subsection in Background states the positive rule: a trait method is added only when a second implementation actually needs it. Loom and TigerBeetle precedents cited.
- [A4] *Addressed.* New Open Question flags the LTS-equivalence requirement for Phase 2 and notes that the v1 trait surface is necessary but possibly not sufficient — to be validated by the Phase-2 RFC.

No findings were classified as out-of-scope. Re-spawning the three blocking archetypes (Security, OS Engineer, Academic) for cycle-2 verdicts.

dburkart:
### Review: Security Researcher (cycle 2)

**Summary:** B1 fully addressed — the cfg gate is now `(test, feature = "sched-mock")` with no target_os escape hatch, and there's an enforced CI check via `nm` rather than an asserted exclusion. Adapter `debug_assert!` and `Tick` field privacy also accepted.

**Blocking findings:**
None.

**Advisory findings:**
- [A1] (carry-over, optional) The `nm`-based check is good defense in depth, but if a future PR adds `panic_handler` or LTO flags that strip symbol names, the check will silently false-negative. Worth pairing with a `#[link_section]` or a `static_assertions::assert_cfg!` style compile-time check that fails to build the release kernel if the `sched-mock` feature is on. Not blocking — the runtime nm check is sufficient for v1.

**Verdict:** LGTM

dburkart:
### Review: OS Engineer (cycle 2)

**Summary:** All three blockers cleanly addressed. The IRQ-context contract on the trait docs is exactly what was missing. The init-order contract is well-stated. The roadmap restructure (folding migration + visibility tightening into one PR with a differential test) is the right call. Ship it.

**Blocking findings:**
None.

**Advisory findings:**
- [A1] (carry-over) The mechanical rename `IrqSource → TimerIrq` is now in the roadmap as the last item. Fine, but consider doing it as item 1 instead — every subsequent PR that touches the trait name then doesn't get re-touched. Mechanical churn earlier is cheaper than mechanical churn later. Either order is acceptable.
- [A2] The differential serial-log test is great but worth one practical note for the implementer: QEMU's serial output can include nondeterministic timestamps (e.g. `[12.345]` boot prefixes). The "byte-identical" diff will need to either strip those prefixes or use a deterministic-time QEMU invocation (`-icount`). Worth flagging in the implementing PR's description; not an RFC-level issue.

**Verdict:** LGTM

dburkart:
### Review: Academic (cycle 2)

**Summary:** B1 thoroughly addressed — the Equivalence subsection states the property formally, names the three conditions, and commits to a verifiable test. The Hyperkernel and Jitk citations land the design in the right academic neighborhood. The positive-discipline statement is exactly the kind of self-binding rule the literature recommends. The Tick unit invariant and Phase-2 LTS Open Question close out the remaining advisories.

**Blocking findings:**
None.

**Advisory findings:**
None worth raising at RFC stage; remaining nuances belong in the implementing PRs.

**Verdict:** LGTM

PR #684, opened 2026-04-28
test: userspace shell-pipeline integration test (echo | cat | wc -c)

### LINKED ISSUE #462: test: userspace shell-pipeline integration test (echo | cat | wc -c)
## Motivation
Follow-up of #380 (being decomposed because the full scope spans multiple subsystems). This issue captures the userspace pipeline deliverable alone. Verifies RFC 0003 end-to-end: a real shell pipeline `echo foo | cat | wc -c` runs in userspace, returning `4\n`. Exercises pipes + fork/exec + fd inheritance + wait together.

## Work
- [ ] Add a new userspace binary under `userspace/shell_pipeline/` (use `userspace/hello/` as a template) that implements `echo foo | cat | wc -c` using pipe/fork/exec/dup2/wait syscalls, and writes a recognizable marker (e.g. `SHELL_PIPELINE_OK: 4`) to stdout.
- [ ] Wire it into the kernel's initrd/ramdisk like `hello`/`init`.
- [ ] Add an xtask QEMU-based smoke test (or extend `cargo xtask smoke` markers) that loads this binary and asserts the marker is seen.
- [ ] Do NOT include the N_TTY Ctrl-C companion test — that lives in a sibling issue.

## Context
- Parent issue: #380 (to be closed in favor of this + a sibling)
- Merged RFC: `docs/RFC/0003-pipes-poll-tty.md`
- Template binary: `userspace/hello/` and `userspace/init/`
- Related subsystems: anonymous pipe (#370 closed), job control (#376 closed), N_TTY (#375 closed)

### LINKED ISSUE #690: kernel/fork: child loses user GPRs (rdi/rsi/rdx/rbx/rbp/r12-r15) across fork return
## Motivation

Discovered while wiring up the shell-pipeline integration test for #462 (PR #684).
The forked child of a userspace process wakes up in ring-3 with a user GPR set that
is **not** the parent's snapshot — it is whatever the kernel's `fork_child_sysret`
trampoline happens to leave in the registers. This silently corrupts any local
variable the compiler was holding in a callee-saved register at the call site of
`sys_fork()` in userspace.

Concretely, with two consecutive `pipe2()` calls followed by a `fork()`, the parent
sees the expected `[p1r=3, p1w=4, p2r=5, p2w=6]`. The child running the same
function-local code path sees `p1r=0, p1w=0, p2r=0, p2w=stale-non-zero` — the
locals have been clobbered.

## Repro

`cargo xtask shell-pipeline` from PR #684's branch. Local run on amd64:

```
shell_pipeline: starting echo|cat|wc-c
shell_pipeline: diag_fd[supervisor.0=0] write0=0
shell_pipeline: diag_fd[supervisor.1=1] write0=0
shell_pipeline: diag_fd[supervisor.2=2] write0=0
shell_pipeline: p1r=3              ← parent: pipe2 returned (3,4)
shell_pipeline: p1w=4
shell_pipeline: p2r=5              ← parent: pipe2 returned (5,6)
shell_pipeline: p2w=6
shell_pipeline: diag_fd[echo-child.p1r=0] write0=0   ← child: p1r is 0!
shell_pipeline: diag_fd[echo-child.p1w=0] write0=0   ← child: p1w is 0!
echo-child: dup2(p1w, STDOUT)=1                       ← dup2(0,1) "succeeds" (fd 0 was open)
shell_pipeline: diag_fd[echo-child.STDOUT-after-dup2=1] write0=0
shell_pipeline[echo]: enter
shell_pipeline[echo]: write failed, exit 21          ← write goes to serial → fails non-deterministically
```

Subsequent forks for cat/wc see the same pattern: `cat-child.p1r=0`,
`cat-child.p2w=514` (random kernel-leaked value), etc.

## Root cause

`kernel/src/task/switch.rs::fork_child_sysret` only restores 3 registers from the
fork prime layout:

```asm
mov rcx, r12      // user RIP
mov r11, rbx      // user RFLAGS
xor eax, eax      // child fork() return value
mov rsp, rbp      // user RSP
sysretq
```

It does **not** restore `rdi`, `rsi`, `rdx`, `r8`, `r9`, `r10`, **`rbx`**, **`rbp`**,
**`r12`**, **`r13`**, **`r14`**, **`r15`** to the parent-snapshot user values. After
SYSRETQ the child runs ring-3 code with:
- `rbx = parent's user RFLAGS` (whatever bit-pattern that is, ~0x202)
- `rbp = parent's user RSP`
- `r12 = parent's user RIP`
- `r13/r14/r15` = leftover kernel scratch
- `rdi/rsi/rdx/r8/r9/r10` = leftover from `xor eax, eax; mov rsp, rbp; sysretq`

The compiler's userspace SysV calling convention treats `rbx, rbp, r12-r15` as
callee-saved across function calls — so a userspace local held in any of them
across `sys_fork()` must, by SysV, survive. The parent path satisfies this because
the kernel never touches those regs in the C handler; the child path violates it
because `fork_child_sysret` overwrites them with kernel-staged values.

The pre-existing `lateout(...)` clobber list in `sys_fork()` only declares the
caller-saved regs — adding callee-saved to the asm clobbers would mask the bug at
the cost of a needless spill on every syscall and would not fix any other syscall's
fork-shaped path. The right fix is in the kernel.

## Fix sketch

`Task::new_forked` should prime the child's kernel stack with the **full** user
GPR snapshot from the parent's `SyscallReturnContext`, and `fork_child_sysret`
should pop them all back before SYSRETQ. The parent already has `rdi/rsi/rdx/r8/
r9/r10` saved on its kernel stack at the time of the FORK syscall (see
`syscall_entry` lines 1564–1570) — those need to be copied into the child's prime
layout. The remaining callee-saved regs (`rbx, rbp, r12-r15`) are NOT saved by
`syscall_entry` today — they need to be added.

## Why this hasn't been caught

- `userspace/hello` does no fork.
- `userspace/init` forks but exec's right away, which loads a fresh ELF (rip jumps
  to entry, all GPRs are fair game).
- `userspace/repro_fork` forks and immediately exec's into hello — also resets all
  GPRs.

The shell-pipeline test in #462 is the first userspace pattern that forks and
**continues running its own code with carry-over locals**. Every other test path
masks the bug behind an immediate execve.

## Work

- [ ] Extend `SyscallReturnContext` with the missing callee-saved register slots
      (`rbx`, `rbp`, `r12`–`r15`) and have `syscall_entry` save/restore them.
- [ ] Have `fork_current_task` / `Task::new_forked` copy the full user GPR set
      from the parent's saved context into the child's prime layout.
- [ ] Update `fork_child_sysret` to pop all 13 user GPRs before `sysretq`
      (rax stays zero — child fork return value).
- [ ] Add a unit test that forks, then in the child reads back a parent-set local
      that the compiler has spilled to a callee-saved reg (e.g. force it via
      `core::hint::black_box` on a value held across `sys_fork()`).
- [ ] Once landed, un-park PR #684 and re-run shell-pipeline.

## Context

- Discovered: PR #684 (#462), commit dc60e48 + diag commit on `m462-shell-pipeline-test`.
- Affected files: `kernel/src/task/switch.rs` (fork_child_sysret), `kernel/src/task/task.rs`
  (Task::new_forked priming), `kernel/src/arch/x86_64/syscall.rs` (syscall_entry,
  SyscallReturnContext layout — see lines 1518–1657).
- Related: #531 (which discusses caller-saved clobbers but not the callee-saved /
  fork-return path), #501 (fork/exec/wait quality sprint).
- Blocks: #462 (and any future test that forks and continues userspace code without
  immediately exec'ing).

dburkart:
## Summary

Userspace integration test for #462 plus the kernel fork-GPR fix it
surfaced (#690). Spawns a 3-stage pipeline (`echo foo | cat | wc -c`)
entirely inside one process tree and asserts the final byte count is 4.

Wires:
- new `userspace/shell_pipeline/` binary (lower-half load at 0x400000)
- `cargo xtask shell-pipeline` subcommand (kernel + binary + ISO via
  `make_iso_with_init`, QEMU boot, marker-watcher loop)
- new `shell-pipeline integration test` CI job mirroring the existing
  `smoke` job

Closes #462.
Closes #690.

## #690 fix

The pipeline test is the first userspace pattern that forks and
continues running its own code with carry-over locals; every other
fork test masked the bug behind execve's full-GPR reload.
`fork_child_sysret` only restored rcx/r11/rsp on SYSRETQ, leaving
rbx/rbp/r12-r15 (callee-saved per SysV) and the syscall-arg
registers with whatever the kernel happened to leave there. The fix:

- Extend `SyscallReturnContext` with the 6 callee-saved slots so
  `syscall_entry` captures the full user GPR set on every SYSCALL.
- New `ForkUserRegs` struct flows the 13-GPR snapshot from FORK's
  syscall-dispatch arm through `fork_current_task` -> `Task::new_forked`
  -> `prime_fork_child_stack`.
- Rewrite `fork_child_sysret` to pop/load the full GPR set before
  SYSRETQ. Host unit tests in `kernel/src/fork_abi.rs` cover the new
  16-qword prime layout.
- Add a `core::hint::black_box`-anchored callee-saved-roundtrip probe
  at the head of the shell-pipeline supervisor so a #690 regression
  surfaces as a single fast-failing marker instead of opaque dup2
  cascade output.

## Test plan

- [x] `cargo xtask build` green locally
- [x] `cargo xtask shell-pipeline` reproduces #690 deterministically (pre-fix)
- [x] After #690 fix: `cargo xtask shell-pipeline` emits
      `SHELL_PIPELINE_CALLEE_SAVED_OK` then `SHELL_PIPELINE_OK: 4`
- [x] `cargo xtask test` green (518 host unit + integration tests pass)
- [x] `cargo xtask smoke` green (41/41 markers)
- [ ] shell-pipeline CI job goes green

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -275,6 +275,46 @@ jobs:
           path: target/debugcon.log
           if-no-files-found: ignore
 
+  # Issue #462: end-to-end shell-pipeline integration test. Boots the
+  # `userspace_shell_pipeline` binary as PID 1 and asserts the
+  # `SHELL_PIPELINE_OK: 4` marker on serial. Exercises pipe2 + fork +
+  # dup2 + close + read/write + wait4 against RFC 0003.
+  shell-pipeline:
+    name: shell-pipeline integration test
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update -q
+          sudo apt-get install -y --no-install-recommends \
+            qemu-system-x86 xorriso build-essential
+
+      - name: Set up Rust toolchain
+        uses: dtolnay/rust-toolchain@3c5f7ea28cd621ae0bf5283f0e981fb97b8a7af9 # master
+        with:
+          toolchain: nightly
+          components: rust-src, llvm-tools, rustfmt, clippy
+          targets: x86_64-unknown-none
+
+      - name: Cache cargo + limine
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+            build/limine
+          key: ${{ runner.os }}-cargo-shell-pipeline-${{ hashFiles('**/Cargo.lock', 'xtask/src/main.rs') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-shell-pipeline-
+            ${{ runner.os }}-cargo-
+
+      - name: cargo xtask shell-pipeline
+        run: cargo xtask shell-pipeline
+
   # RFC 0004 Workstream G P2 (#582): run the vendored pjdfstest harness
   # (#580 + #581) and fail if the emitted
   # `target/pjdfstest-results.json` diverges from the committed baseline
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "userspace/init",
     "userspace/pjdfstest_runner",
     "userspace/repro_fork",
+    "userspace/shell_pipeline",
 ]
 # tests/pjdfstest is a vendored copy of saidsay-so/pjdfstest (2-clause BSD). It
 # is intentionally kept outside the workspace until #581 wires it into xtask —
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -638,10 +638,26 @@ pub unsafe extern "C" fn syscall_dispatch(
             // own SyscallReturnContext on the kernel stack. Per-task by
             // construction: no cross-task races vs. the old FORK_USER_*
             // globals (issue #504).
-            let user_rip = (*ctx).user_rip;
-            let user_rflags = (*ctx).user_rflags;
-            let user_rsp = (*ctx).user_rsp;
-
+            //
+            // The full GPR set (including SysV callee-saved rbx/rbp/r12-r15)
+            // is published into the child via ForkUserRegs — see #690.
+            let regs = crate::fork_abi::ForkUserRegs {
+                user_rip: (*ctx).user_rip,
+                user_rflags: (*ctx).user_rflags,
+                user_rsp: (*ctx).user_rsp,
+                user_rdi: (*ctx).user_rdi,
+                user_rsi: (*ctx).user_rsi,
+                user_rdx: (*ctx).user_rdx,
+                user_r10: (*ctx).user_r10,
+                user_r8: (*ctx).user_r8,
+                user_r9: (*ctx).user_r9,
+                user_rbx: (*ctx).user_rbx,
+                user_rbp: (*ctx).user_rbp,
+                user_r12: (*ctx).user_r12,
+                user_r13: (*ctx).user_r13,
+                user_r14: (*ctx).user_r14,
+                user_r15: (*ctx).user_r15,
+            };
             // #502 probe: dump the current RFLAGS at fork-dispatch entry.
             // Expected: IF (bit 9) = 0, masked by MSR_FMASK's IF bit on
             // SYSCALL entry. If IF is ever observed = 1 here, FMASK
@@ -654,9 +670,9 @@ pub unsafe extern "C" fn syscall_dispatch(
                     "fork-trace: [syscall:FORK enter] parent_task={} user_rip={:#x} \
                      user_rflags={:#x} user_rsp={:#x} kernel_rflags={:#x} IF={}",
                     crate::task::current_id(),
-                    user_rip,
-                    user_rflags,
-                    user_rsp,
+                    regs.user_rip,
+                    regs.user_rflags,
+                    regs.user_rsp,
                     rflags_now,
                     (rflags_now >> 9) & 1,
                 );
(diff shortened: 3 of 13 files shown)


dburkart:
Parking this PR: the diagnostic round added in the latest push surfaced a real
kernel bug (filed as #690).

Summary: when `_start` calls `sys_fork()` and the child resumes via
`fork_child_sysret`, the child's userspace registers are not restored from the
parent's snapshot — `rbx`, `rbp`, `r12`–`r15` are loaded with kernel-staged
values (user RIP/RFLAGS/RSP), and `rdi/rsi/rdx/r8/r9/r10` are kernel scratch.
Any local variable the compiler held in a callee-saved register across the
`sys_fork()` call (here: `p1r`, `p1w`, `p2r`, `p2w`) is silently zeroed (or
worse, randomized) in the child.

Visible from `cargo xtask shell-pipeline` locally:

```
parent:    p1r=3 p1w=4 p2r=5 p2w=6
echo-child: diag_fd[p1r=0] diag_fd[p1w=0]   ← clobbered
cat-child:  diag_fd[p1r=0] diag_fd[p2w=514] ← clobbered + leaked kernel scratch
```

This pattern (fork → continue running same userspace code without exec) is
unique to #462 — `init` and `repro_fork` mask the bug by exec'ing immediately,
which discards all GPR state.

#690 has the full root-cause analysis and fix sketch. This PR is the test that
will gate landing the fix: once #690's kernel patch lands, the shell-pipeline
job should go green with no userspace changes needed.

Converting to draft. Will un-park and rebase once #690 lands.

dburkart:
CodeRabbit nits acknowledged and deferred to follow-up issues; not blocking the #690 fix:

- Extracting the shared QEMU+serial-watchdog plumbing between `shell_pipeline()` and `repro_fork()` in `xtask/src/main.rs` — tracked separately.
- shell_pipeline marker-write error checking (`write_all`/`write_u32` short-write handling) and using `WIFEXITED`-style status decoding in `wait_zero` — tracked separately.

The link.ld header comment referencing repro_fork was already cleaned up earlier in this branch. All CI checks green; merging to land the #690 fix on top of the already-merged scheduler-seam work.

PR #439, opened 2026-04-15
tty: add DeferredByteRing, TtyDriver/LineDiscipline, extended Tty

dburkart:
Closes #403

## Summary

- New `kernel/src/tty/ring.rs` — allocation-free 128-byte SPSC byte ring with `AtomicU8` slots, cache-padded `head`/`tail`, and a `watermark_high()` @ 75% fill for future RTS flow-control wiring.
- Extended `kernel/src/tty/mod.rs` with the `TtyDriver` and `LineDiscipline` traits, a `PassthroughLdisc` (doubles as the default ldisc until N_TTY #428/#375 lands, and as the test sink), a `NullDriver` stub, and a full `Tty` container (`driver`, `ldisc`, `termios`, `ctrl`, `read_wait`, `write_wait`).
- Existing `Tty::new()` keeps working for the PCB controlling-terminal stub (#372 setsid path) — it now defaults to `NullDriver + PassthroughLdisc`.
- Lock type is `IrqLock` per RFC 0003 §Lock order, with the same `#[cfg(all(test, not(target_os = "none")))]` fallback to `spin::Mutex` that `kernel/src/poll/mod.rs` uses so host unit tests still compile.
- **No driver wiring in this PR.** Serial UART (#406) and PS/2 keyboard (#405) land once the soft-IRQ bottom-half primitive from #404 is available.

## Test plan

- [x] `cargo xtask build` — clean.
- [x] `cargo xtask test` — host unit tests up from 251 → 258 (+7 new: ring empty/push-pop/fill/wrap/watermark, passthrough ldisc byte ordering, null driver). Pre-existing `blocking_sync::rwlock_writer_not_starved` flake (#385, fix in flight as PR #436) is unrelated.
- [x] `cargo xtask smoke` — all 30 markers present.

PR #686, opened 2026-04-28
tty: wire NTtyLdisc into PS/2 + serial rx paths

### LINKED ISSUE #474: tty: wire N_TTY ldisc into PS/2 + serial rx paths (replace PassthroughLdisc default)
## Motivation

All of the N_TTY pipeline (ICANON line buffering, OPOST, ISIG signal generation, iflag transforms) has landed via #437/#453/#443/#449, and both rx paths — PS/2 keyboard (#442) and 16550 serial (#444) — now drain through `Tty.ldisc.receive_byte()`. But the ldisc wired into both `CONSOLE_TTY` instances is `PassthroughLdisc`, whose `receive_byte` just drops the byte on the floor (`kernel/src/tty/mod.rs:187` — no-op impl).

Net effect: `NTty::receive_signal_or_byte` exists, is fully tested at the unit level, and is called by nothing. Ctrl-C typed at the shell does not generate SIGINT; newline editing does not commit to a read buffer; ECHO does not echo. The subsystem reads as complete from the PR list but the user-visible behavior is unchanged from before #437 landed.

This is the single wiring step that makes the RFC 0003 terminal stack actually functional end-to-end.

## Work

- [ ] Define an `NTtyLdisc` type in `kernel/src/tty/ntty.rs` that owns an `NTty` + a `SignalDispatch` impl (`KernelSignalDispatch` for production) and implements `LineDiscipline`. Its `receive_byte(&self, b: u8)` must call `self.ntty.receive_signal_or_byte(&termios, &ctrl, &dispatch, b)` — reading `Termios` and `JobControl` via the `Tty` reference the ldisc holds or is handed at construction.
- [ ] Decide how the ldisc reaches `Termios`/`JobControl`. Option A: store `Weak<Tty>` in the ldisc and upgrade on each rx. Option B: change the `LineDiscipline::receive_byte` trait signature in `kernel/src/tty/mod.rs:130` to take `&Tty` (or `&Termios, &JobControl`) as context. Option B is cleaner — `PassthroughLdisc` currently takes no context because it needs none, but the trait exists to serve N_TTY, so shaping it around N_TTY's needs is correct.
- [ ] For bytes that `receive_signal_or_byte` returns as `Some(b)` (i.e. not consumed as a signal), thread them into the canonical/raw commit path (`NTtyState::canon_input` / `commit_line`) so readers of `/dev/tty` actually see the data. The current standalone `receive_signal_or_byte` contract stops at "byte survived ISIG"; ICANON commit is separate and still needs to be called from the ldisc glue.
- [ ] Swap the default ldisc in both `kernel/src/tty/ps2.rs:60` and `kernel/src/tty/serial.rs:83` from `PassthroughLdisc::new()` to the new N_TTY ldisc.
- [ ] Add an integration test under `kernel/tests/` (e.g. `ntty_ldisc_wiring.rs`) that: (a) pushes a VINTR byte through `push_scancode_from_isr` / the serial rx ring and asserts SIGINT lands on the foreground pgrp, (b) pushes `abc\n` and asserts a subsequent `read()` on the tty returns `abc\n`. The parent issue #463 already describes a more focused companion test for the canonical-mode Ctrl-C flush; this wiring test is the prerequisite that makes that test writable against the real path instead of a shim.
- [ ] Keep `PassthroughLdisc` in-tree as the fallback for test scaffolding (the `Tty::new()` default in `kernel/src/tty/mod.rs:246` still wants it), but it should no longer be the default for console-attached ttys.

## Context

- Ldisc + traits landed in #439 (merge commit `a1c8afd`, file `kernel/src/tty/mod.rs`).
- PS/2 rx path landed in #442 (merge `dea622a`, file `kernel/src/tty/ps2.rs`).
- Serial rx path landed in #444 (merge `40b303f`, file `kernel/src/tty/serial.rs`).
- N_TTY skeleton + iflag landed in #437 (`568a104`), ICANON in #453 (`ad3c62f`), OPOST/ECHO in #443 (`5545211`), ISIG in #449 (`f71b0d6`).
- RFC: `docs/RFC/0003-pipes-poll-tty.md` §N_TTY, §ISIG.
- Related: #463 (canonical-mode Ctrl-C integration test — this issue unblocks that one), #462 (userspace shell pipeline test — also benefits from real ldisc behaviour).
- Flagged as a systemic risk in the 2026-04-16 project report (Discussion #472).

dburkart:
Closes #474

## Summary

- Add `NTtyLdisc` in `kernel/src/tty/ntty.rs`: owns an `NTty` + an injected `Arc<dyn SignalDispatch>` and implements `LineDiscipline`. `receive_byte(&Tty, u8)` snapshots `Termios` + the lock-free `pgrp_snapshot`, runs the byte through `receive_signal_or_byte` (ISIG + iflag), then hands the surviving byte to `canon_input`, which routes into the line editor (canonical) or raw ring (raw) and wakes `tty.read_wait` on each commit.
- The `LineDiscipline::receive_byte(&self, &Tty, u8)` signature already takes `&Tty` (landed in #439), so option B from the issue ("reshape the trait around N_TTY's needs") was already in place — no trait churn.
- `SignalDispatch` gains a `Send + Sync` bound so `NTtyLdisc` can hold `Arc<dyn SignalDispatch>`. Production wires `KernelSignalDispatch`; the existing host-test `CapturingDispatch` migrates from `RefCell` to `spin::Mutex` to satisfy the bound.
- Swap the default ldisc in `kernel/src/tty/ps2.rs:60` and `kernel/src/tty/serial.rs:83` from `PassthroughLdisc::new()` to `NTtyLdisc::new(KernelSignalDispatch)`. `Tty::new()` (the stub container used by the PCB before any driver attaches) keeps `PassthroughLdisc` per the issue's last bullet.
- `PassthroughLdisc` stays in tree as the test harness for `LineDiscipline`-shape tests.

## Test plan

- [x] `cargo xtask build` — kernel lib + bin build clean.
- [x] `cargo xtask test` — full QEMU integration suite green (host unit tests + every `[[test]]` from `kernel/Cargo.toml`).
- [x] `cargo xtask smoke` — all 41 markers present.
- [x] New `kernel/tests/ntty_ldisc_wiring.rs` (5 cases, all pass under QEMU):
  - `ps2_ldisc_is_ntty` — pushing a byte through `ps2::tty().ldisc.receive_byte` doesn't panic (PassthroughLdisc would no-op; we want to confirm the new ldisc accepts the call).
  - `serial_ldisc_is_ntty` — same for `serial::tty()`.
  - `vintr_through_ldisc_signals_pgrp` — typing `abc` then VINTR through `ldisc.receive_byte` raises SIGINT on the foreground pgrp and flushes the raw ring.
  - `canonical_line_commits_through_ldisc` — `abc` is invisible to readers until `\n` commits 4 bytes (`abc\n`) into the raw ring.
  - `raw_mode_bytes_visible_immediately` — with ICANON cleared, every byte hits the raw ring on push.
- [x] Existing N_TTY unit tests (`ntty::tests`) and `ntty_sigint_flush` integration test still pass unchanged.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -347,6 +347,10 @@ name = "ntty_sigint_flush"
 harness = false
 
 [[test]]
+name = "ntty_ldisc_wiring"
+harness = false
+
+[[test]]
 name = "gdbstub_loop"
 harness = false
 
--- a/kernel/src/tty/ntty.rs
+++ b/kernel/src/tty/ntty.rs
@@ -36,7 +36,7 @@ pub trait OutSink {
 /// `target_os = "none"` and unavailable to host unit tests. Production
 /// code wires [`KernelSignalDispatch`] (defined only for
 /// `target_os = "none"`); tests supply their own.
-pub trait SignalDispatch {
+pub trait SignalDispatch: Send + Sync {
     /// True when the pgrp has no live members. When this returns true,
     /// the ISIG path delivers `SIGHUP` + `SIGCONT` via [`Self::send_to_pgrp`]
     /// instead of the originally-requested signal — mirrors Linux's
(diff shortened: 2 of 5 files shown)


dburkart:
@coderabbitai review

Issue #458, opened 2026-04-16
console: alternate screen buffer (CSI ?1049h/l) for framebuffer console


dburkart:
## Motivation
Follow-up of #67. Full-screen TUI applications (vim, less, htop, man) use CSI ?1049h to switch to an alternate screen buffer that is preserved across the application's lifetime, then CSI ?1049l to restore the user's prior screen + scrollback on exit. Without this, running a TUI app destroys the user's terminal history. Depends on the DEC private mode parser being landed first so CSI ?1049 can be dispatched.

## Work
- [ ] Allocate a secondary screen buffer (cells + cursor position + attributes) of the same dimensions as the primary.
- [ ] On CSI ?1049h: save primary cursor position, switch active buffer to alt, clear it, do not preserve scrollback into alt.
- [ ] On CSI ?1049l: switch back to primary, restore saved cursor; alt buffer contents discarded.
- [ ] Also handle the older CSI ?47h/l + ?1047h/l + ?1048h/l variants if cheap — at minimum log them as unsupported.
- [ ] Ensure scrollback (#45) is unaffected by alt-screen toggles: only primary buffer's history accumulates.
- [ ] Host unit test: switch to alt, write content, switch back; primary buffer must be byte-identical to its pre-switch state.

## Context
- File: kernel/src/framebuffer.rs
- Parent issue: #67
- Depends on: the VT100 DEC private mode + extended CSI parser sub-issue (also a #67 follow-up)
- Related: #45 (scrollback) — alt-screen must not corrupt scrollback.
- Filed as one of three splits of #67 by auto-engineer.

PR #687, opened 2026-04-28
console: implement DECSTBM scroll region (CSI Pt;Pb r)

### LINKED ISSUE #457: console: scroll regions (DECSTBM) for framebuffer console
## Motivation
Follow-up of #67. CSI r (DECSTBM — Set Top and Bottom Margins) defines a scroll region that limits where Index/Reverse-Index/scroll operations apply. Without it, full-screen apps (vim, less, status-bar TUIs) cannot keep header/footer rows fixed while scrolling the body. Depends on the DEC private mode + extended CSI parser being landed first so the parser can dispatch CSI r.

## Work
- [ ] Add scroll-region top/bottom indices to the framebuffer console state.
- [ ] Implement CSI Pt;Pb r (DECSTBM): set region, clamp to screen rows, reset cursor to (1,1).
- [ ] Make scroll-up / scroll-down operations respect the region: only rows in [top, bottom] move; rows outside stay put.
- [ ] Update Index (ESC D), Reverse-Index (ESC M), and Next-Line (ESC E) to respect the region.
- [ ] Reset region to full screen on RIS (ESC c) and DECSC/DECRC interactions per VT100 spec.
- [ ] Host unit test: set region [3..6], emit 10 newlines, verify rows 1-2 and 7+ are untouched.

## Context
- File: kernel/src/framebuffer.rs
- Parent issue: #67
- Depends on: the VT100 DEC private mode + extended CSI parser sub-issue (also a #67 follow-up)
- Filed as one of three splits of #67 by auto-engineer.

dburkart:
Closes #457.

## Summary

- Add a `ScrollRegion { top, bottom }` struct (0-indexed inclusive) on `Console` so the active scroll region is a single cohesive field — keeping the data model tight ahead of the alt-screen work in #458.
- Wire `CSI Pt;Pb r` (DECSTBM) with VT100 semantics: zero in either slot defaults that edge (top → 1, bottom → rows); inverted or degenerate requests are rejected per xterm; cursor is homed to (1,1) on every accepted DECSTBM. A bare `CSI r` resets to the full screen.
- Wire `ESC D` (IND), `ESC M` (RI), `ESC E` (NEL): scroll by one within the active region only when the cursor is on the appropriate margin; cursors below the region clamp at the last row instead of forcing a scroll.
- LF and wrap-on-overflow now route through the same region-aware line-feed helper, so newline output respects the region too.
- Wire `ESC c` (RIS, full reset): clears screen, resets SGR, drops saved DECSC slot, and resets the region to full-screen.
- `Console::scroll_up` only pushes to scrollback when the region spans the whole screen — partial-region evictions are silently dropped to avoid corrupting history with split-region snippets. Pixel blits confine to the region. New `Console::scroll_down` mirror for reverse-index semantics.
- The pure scroll math lives in `scroll_region_up` / `scroll_region_down` free functions over `&mut [Cell]`, so the model side is exercisable under host `cargo test` without standing up a framebuffer pointer.
- Compile the framebuffer module for host tests (`#[cfg(any(test, target_os = "none"))]`), which also brings the existing `classify_dec_mode` tests under host-CI coverage.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test --lib -p vibix` — 530 host unit tests pass, including the 12 in `framebuffer::tests`. The required regression test from #457 (`region_3_through_6_survives_ten_newlines`) verifies that with `region [3..6]` (1-indexed) and 10 scrolls, rows 1-2 and 7+ are untouched and the region itself is fully blanked.
- [ ] CI runs the full kernel-target integration suite + smoke.

### DIFF
--- a/kernel/src/framebuffer.rs
+++ b/kernel/src/framebuffer.rs
@@ -12,6 +12,16 @@
 //!
 //! DEC private and extended VT100 sequences handled:
 //! - `ESC 7` / `ESC 8` (DECSC / DECRC): save / restore cursor + SGR state.
+//! - `ESC D` (Index, IND): move cursor down one row, scrolling the active
+//!   region if at the bottom margin.
+//! - `ESC M` (Reverse Index, RI): move cursor up one row, scrolling the
+//!   active region down if at the top margin.
+//! - `ESC E` (Next Line, NEL): CR + LF, region-aware.
+//! - `ESC c` (RIS, full reset): clear screen, reset SGR, reset scroll
+//!   region to full screen, restore default state.
+//! - `CSI Pt;Pb r` (DECSTBM): set scroll-region top/bottom margins (clamped
+//!   to screen rows) and home the cursor to (1,1). An empty parameter list
+//!   resets the region to the full screen.
 //! - `CSI ? Pm h` / `CSI ? Pm l` (DECSET / DECRST): set / reset DEC private
 //!   modes. Known modes `?1` (cursor-keys app), `?7` (auto-wrap), `?12`
 //!   (cursor blink), `?25` (cursor visible), and `?1049` (alt-screen) are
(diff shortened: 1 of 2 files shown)


dburkart:
@coderabbitai review

Issue #45, opened 2026-04-12
Framebuffer console: scrollback, ANSI color, blinking cursor


dburkart:
## Motivation
Current framebuffer console prints and forgets. Scrollback + color make the kernel actually pleasant to use on real hardware (where serial isn't an option) and make the future shell feel like a shell.

## Scope
- Back the console with a character grid (glyph + fg/bg color) sized to the framebuffer.
- Scroll up on newline past the last row, preserving a small scrollback buffer in RAM.
- Minimal ANSI parser: `CSI n m` for SGR colors (at least 8 fg + 8 bg + reset + bold), plus `\r`, `\t`.
- Timer-driven blinking cursor at the write position.

## Out of scope
- Full vt100 emulation, scroll regions, alternate screen buffer.
- Unicode beyond the existing font8x8 coverage.
- Paging the scrollback with keys (until we have a shell).

## Done when
- Boot log shows colored log levels (once leveled logger lands) and a visible cursor.
- Scrolling past the last row preserves earlier output in the scrollback.

dburkart:
@claude file issues for all out of scope action items.

dburkart:
@claude Implement

PR #693, opened 2026-04-28
console: alternate screen buffer (CSI ?1049/?1047/?47/?1048)

### LINKED ISSUE #458: console: alternate screen buffer (CSI ?1049h/l) for framebuffer console

dburkart:
Closes #458

## Summary
- Adds a secondary cell grid that swaps in on `CSI ?1049h`, blanks on entry, and swaps out on `CSI ?1049l` to restore the primary buffer byte-identically. Primary cursor + SGR + DECSTBM region are stashed in a dedicated alt-screen save slot (separate from DECSC's `saved_*`), so DECSC inside an alt session never clobbers the primary's restore state.
- Gates scrollback (#45) on `!alt_active` in `scroll_up` — alt-screen TUIs (vim, less, htop) repaint every frame, and pushing those frames into history would corrupt the user's pre-alt scrollback that `?1049l` is supposed to expose again.
- Wires the older variants for xterm-compat: `?1047h/l` swaps buffers without cursor save/restore; `?47` is treated like `?1047`; `?1048h/l` saves/restores cursor + SGR only.
- Adds a test-only `Console::for_test` constructor that backs a console with a caller-owned `&mut [u32]` slab so the full `write_char` parser path is exercisable from `cargo test` without Limine.

## Test plan
- [x] `cargo xtask build`
- [x] `cargo test --package vibix --lib --features ext2 framebuffer::` — 20/20 pass, including 9 new tests.
- [x] Required regression test: switch to alt → write content → switch back → primary buffer is byte-identical to its pre-switch state (`csi_1049h_then_l_restores_primary_byte_for_byte`).
- [x] Scrollback non-interaction: `csi_1049_cycle_does_not_grow_scrollback` — alt-screen scrolls don't accumulate in the ring.
- [x] Region isolation: `alt_screen_dec_region_does_not_leak_to_primary` — DECSTBM inside alt mode doesn't leak to primary.
- [ ] CI green (cargo xtask test, smoke).

## Notes
Pre-existing failures observed on `main` at `b3f6bf5`: `cargo xtask test` panics in `kernel/src/task/mod.rs:97` (env identity assertion) and `cargo xtask smoke` hits the same panic post-userspace-load. These reproduce on a clean `main` checkout and are unrelated to this change. `cargo xtask lint` also fails with pre-existing clippy `manual_is_multiple_of` errors in `kernel/build.rs` and `kernel/src/mem/addrspace.rs`.

### DIFF
--- a/kernel/src/framebuffer.rs
+++ b/kernel/src/framebuffer.rs
@@ -24,9 +24,12 @@
 //!   resets the region to the full screen.
 //! - `CSI ? Pm h` / `CSI ? Pm l` (DECSET / DECRST): set / reset DEC private
 //!   modes. Known modes `?1` (cursor-keys app), `?7` (auto-wrap), `?12`
-//!   (cursor blink), `?25` (cursor visible), and `?1049` (alt-screen) are
-//!   accepted as no-ops so real terminal applications (vim, less, htop) do
-//!   not see sequences rejected; unrecognized modes are silently ignored.
+//!   (cursor blink), `?25` (cursor visible) are accepted as no-ops so real
+//!   terminal applications (vim, less, htop) do not see sequences rejected.
+//!   `?1049` (alt-screen with cursor save) and `?1047` (alt-screen, clear
+//!   on exit) swap to a secondary cell grid and back; `?47` (legacy
+//!   alt-screen) is treated as `?1047`; `?1048` is cursor save/restore
+//!   only. Unrecognized modes are silently ignored.
 //!
 //! A blinking cursor is drawn at the write position.  The blink rate is
 //! driven by the kernel task spawned in `main.rs` via [`toggle_cursor`].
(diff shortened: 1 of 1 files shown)


dburkart:
@coderabbitai review

PR #694, opened 2026-04-28
ext2: mount-time BGDT/bitmap consistency checks (#678)

### LINKED ISSUE #678: ext2: mount-time bitmap and BGDT consistency checks
## Motivation

`Ext2Fs::mount` validates the superblock magic, feature compat bits, and orphan-chain head, but does not cross-check the block-group descriptor table against the actual bitmaps. A subtly corrupt image (e.g. `bg_free_blocks_count` that disagrees with the block bitmap, or an inode bitmap with the reserved-range slots cleared) mounts successfully today and then drifts the allocator state silently. We should reject obviously-inconsistent images at mount, or force the mount RO with a logged reason.

## Work

- [ ] Add a `validate_bgdt(&Superblock, &[BlockGroupDesc], &impl BlockDevice) -> Result<(), Ext2Error>` helper called from the mount path.
- [ ] Per group: read the block bitmap, count clear bits, compare to `bg_free_blocks_count`. Same for the inode bitmap vs `bg_free_inodes_count`.
- [ ] Verify the reserved inode range (1..=`s_first_ino`-1, plus root=2) is marked allocated in group 0's inode bitmap.
- [ ] Verify `s_free_blocks_count` and `s_free_inodes_count` equal the sums across BGDs.
- [ ] On mismatch: log via `kprintln!` with group + counts, force RO mount (set the in-memory mount flag, do NOT clear `s_state` valid-FS).
- [ ] Add integration tests with hand-crafted images covering each mismatch class.

## Out of scope

- Online fsck/repair — we only detect and force RO.
- Cross-checking inode `i_blocks` totals against actual block-bitmap usage (separate consistency check, more expensive).

## Context

- Parent epic: #537.
- Files: `kernel/src/fs/ext2/fs.rs` (mount), `kernel/src/fs/ext2/balloc.rs`, `kernel/src/fs/ext2/ialloc.rs`, `kernel/src/fs/ext2/disk.rs` (BGDT layout).
- RFC: `docs/RFC/0004-ext2-filesystem-driver.md` §Robustness / §Security.

dburkart:
Closes #678

## Summary

Adds a `validate_bgdt(...)` helper called from `Ext2Fs::mount` that cross-checks the block-group descriptor table against the bitmaps it indexes:

- **Per group**: count clear bits in the block + inode bitmaps and compare to `bg_free_blocks_count` / `bg_free_inodes_count`.
- **Group 0 reserved range**: verify inos `1..first_ino` (including ino 2 = `EXT2_ROOT_INO`) are marked allocated in the inode bitmap.
- **Superblock totals**: verify `s_free_blocks_count` / `s_free_inodes_count` equal the sums across BGDT entries.

On any mismatch the validator logs the group + observed-vs-expected counts via `kwarn!` and returns `ForceRo::Yes`. The mount path collapses that verdict into the same `effective_rdonly` branch that already handles unknown RO_COMPAT bits and corrupt orphan chains, so a detectably-inconsistent image mounts RO and never accepts a write.

The validator runs as a raw walk *before* the `Arc<SuperBlock>` is built, mirroring `super::orphan::walk_orphan_chain_raw` — `SbFlags` is immutable post-construction, so the RO decision has to settle first.

## Test plan

New integration test `kernel/tests/ext2_validate_bgdt.rs` patches each mismatch class onto the existing `golden.img` fixture and asserts the mount is demoted to RO with zero writes:

- [x] Healthy fixture mounts RW cleanly
- [x] Block-bitmap clear-bit count != `bg_free_blocks_count` → forces RO
- [x] Inode-bitmap clear-bit count != `bg_free_inodes_count` → forces RO
- [x] Root ino 2 cleared in bitmap → reserved-range check forces RO
- [x] `s_free_blocks_count` lies → forces RO
- [x] `s_free_inodes_count` lies → forces RO
- [x] Existing `ext2_mount`, `ext2_orphan_chain`, `ext2_block_alloc`, `ext2_ialloc` integration tests still pass
- [x] `cargo fmt --all` clean
- [x] Host unit tests (`cargo test -p vibix --lib`): 530 passed

Out of scope per the issue: online repair, cross-checking inode `i_blocks` totals against actual block-bitmap usage.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -465,3 +465,8 @@ required-features = ["ext2"]
 name = "ext2_rename"
 harness = false
 required-features = ["ext2"]
+
+[[test]]
+name = "ext2_validate_bgdt"
+harness = false
+required-features = ["ext2"]
--- a/kernel/src/fs/ext2/fs.rs
+++ b/kernel/src/fs/ext2/fs.rs
@@ -384,6 +384,21 @@ impl FileSystem for Ext2Fs {
             ext2_flags |= Ext2MountFlags::FORCED_RDONLY;
         }
 
+        // 7c. Mount-time BGDT / bitmap consistency checks (issue #678,
+        //     RFC 0004 §Robustness / §Security). Cross-checks per-group
+        //     bitmap clear-bit counts against `bg_free_*_count`,
+        //     verifies the reserved-inode range is allocated in group
+        //     0, and that the superblock totals equal the BGDT sums. A
+        //     `Yes` verdict demotes the mount to RO before the
+        //     `s_state := ERROR_FS` stamp, so a corrupt image never
+        //     accepts a write.
+        let bgdt_verdict =
+            super::validate::validate_bgdt(&cache, device_id, &on_disk_sb, &bgdt, first_ino);
+        if bgdt_verdict == super::orphan::ForceRo::Yes {
+            effective_rdonly = true;
+            ext2_flags |= Ext2MountFlags::FORCED_RDONLY;
+        }
+
         // 8. RW bring-up: stamp `s_state := ERROR_FS` so an unclean
         //    crash is detectable on the next mount. The canonical
         //    `VALID_FS` is rewritten only by a clean unmount. RO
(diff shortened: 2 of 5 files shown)


dburkart:
@coderabbitai review

PR #695, opened 2026-04-28
ext2/tests: cover setattr SIZE on slow (indirect) symlinks (#683)

### LINKED ISSUE #683: ext2: integration test for truncate() on slow (indirect) symlinks
## Motivation

The slow-symlink read path through the indirect-block walker landed in #645. Slow symlinks (target > 60 bytes, stored in data blocks via the indirect chain rather than `i_block` inline) hit a less-tested corner of the truncate path: shrinking the symlink target rewrites the indirect chain, growing it allocates blocks. There is currently no integration test that exercises `truncate()` on a slow-symlink inode.

## Work

- [ ] Add a test in `kernel/tests/ext2_setattr.rs` (or a new `ext2_symlink.rs`) that:
  - Creates a slow symlink (target > 60 bytes).
  - Truncates to a smaller size; asserts target is correctly shortened, blocks are freed, balloc free-count drops accordingly.
  - Truncates to a larger size; asserts the trailing area is zero-filled per POSIX truncate semantics on symlinks (or that the operation returns the correct error if symlinks reject grow-truncate).
  - Truncates to zero; asserts the symlink becomes empty / inode metadata stays consistent.
- [ ] Verify against POSIX: most kernels return `EINVAL` for `truncate(symlink_path, _)` because path resolution follows the link; the test should target `ftruncate(fd_opened_with_O_NOFOLLOW)` semantics if vibix supports that, or the internal `setattr` entry point on the symlink inode directly.

## Out of scope

- Adding `O_NOFOLLOW` syscall surface if it doesn't already exist.

## Context

- Parent epic: #537.
- Slow-symlink read PR: #645.
- Files: `kernel/src/fs/ext2/symlink.rs`, `kernel/src/fs/ext2/setattr.rs`, `kernel/src/fs/ext2/file.rs` (truncate path).

dburkart:
Closes #683.

## Summary

Slow symlinks (target > 60 bytes, stored via the indirect-block walker rather than inline in `i_block[]`) are a corner of the truncate path the integration suite did not previously exercise. POSIX `truncate(2)` follows the link at the path-walk layer, so a SIZE-bearing setattr only reaches the symlink inode through a direct driver call — and the driver mirrors Linux by returning `EINVAL` for any non-regular SIZE target.

This PR extends `kernel/tests/ext2_setattr.rs` rather than adding a new test crate file — that keeps `kernel/Cargo.toml`'s `[[test]]` table stable while another PR is adding an entry there.

Four new cases:

- `truncate_shrink_slow_symlink_is_einval`: 200-byte slow symlink, SIZE -> 100. EINVAL; balloc free count, `i_size`, `i_blocks`, `i_block[]` byte-stable.
- `truncate_grow_slow_symlink_is_einval`: same shape, SIZE -> 4096.
- `truncate_zero_slow_symlink_is_einval`: SIZE -> 0. The variant most likely to leak blocks if the SIZE-on-Link gate ever regressed.
- `non_size_setattr_on_slow_symlink_succeeds`: chmod + utimensat on a slow symlink go through; the data block, `i_blocks`, and free-block counter remain untouched.

Each rejection case asserts the symlink inode stays classified as slow afterward (`is_symlink && !is_fast_symlink`), so a half-applied truncate that happened to clear `i_block[]` would surface immediately. The slow-symlink fixture is built at runtime via `ext2_symlink` with a 200-byte deterministic target, mirroring `ext2_link.rs::symlink_slow_200_bytes_walker`.

## Test plan

- [x] `cargo xtask build`
- [x] `cargo +nightly test -p vibix --test ext2_setattr --target x86_64-unknown-none --features=ext2 -Zbuild-std=...` (all 13 tests pass, including 4 new)
- [x] `cargo fmt --all`

Note: `cargo xtask test` and `cargo xtask smoke` panic on `main` at `kernel/src/task/mod.rs:97` (`HW_CLOCK`/`HW_IRQ` env-pointer assertion) for unrelated reasons; the `ext2_setattr` test binary itself runs clean under qemu. CI will resolve once the env-pointer regression is fixed.

### DIFF
--- a/kernel/tests/ext2_setattr.rs
+++ b/kernel/tests/ext2_setattr.rs
@@ -36,15 +36,18 @@ use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
 use core::panic::PanicInfo;
-use core::sync::atomic::{AtomicU32, Ordering};
+use core::sync::atomic::{AtomicBool, AtomicU32, Ordering};
 
 use spin::Mutex;
 
 use vibix::block::{BlockDevice, BlockError};
-use vibix::fs::ext2::{iget, Ext2Fs, Ext2Super};
+use vibix::fs::ext2::symlink::{is_fast_symlink, is_symlink};
+use vibix::fs::ext2::{iget, symlink as ext2_symlink, Ext2Fs, Ext2Inode, Ext2InodeMeta, Ext2Super};
+use vibix::fs::vfs::inode::{Inode, InodeKind};
 use vibix::fs::vfs::ops::{FileSystem as _, MountSource, SetAttr, SetAttrMask};
 use vibix::fs::vfs::super_block::SuperBlock;
 use vibix::fs::vfs::{MountFlags, Timespec};
+use vibix::sync::BlockingRwLock;
 use vibix::{
     exit_qemu, serial_println,
     test_harness::{test_panic_handler, Testable},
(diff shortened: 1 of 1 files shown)


dburkart:
@coderabbitai review

Issue #670, opened 2026-04-28
sched: rename IrqSource to TimerIrq for honest scope


dburkart:
## Motivation

Implements RFC 0005 — Scheduler / IRQ Seam for Deterministic Simulation Testing. Roadmap item 6: per OS-engineer review advisory A1, the trait covers only the timer IRQ but is named `IrqSource` — misleading. Mechanical rename to `TimerIrq` (or `PreemptionIrq`) before any Phase-2 simulator work depends on the name.

RFC: `docs/RFC/0005-scheduler-irq-seam.md`
Merged in: #664
Depends on: call-site migration (roadmap item 3); ideally also after MockClock (item 4) so the rename only happens once across all call sites.

## Work

- [ ] Decide between `TimerIrq` and `PreemptionIrq` (lean: `TimerIrq` — describes what the trait abstracts, not why).
- [ ] Mechanical rename: trait, all impls (`HwIrq`, `MockIrqSource`), all call sites (`env()` return type, `preempt_tick`).
- [ ] Update RFC 0005 to reflect the rename (a one-line note in the doc).
- [ ] CI green.

## Context

- RFC section: Alternatives Considered, Open Questions
- Strictly cosmetic; defer until just before Phase 2 RFC starts so the name churn happens at most once.
- OS-engineer cycle-2 noted this could equally well be done as item 1 (mechanical churn earlier is cheaper); ordering is implementer's call.

Issue #668, opened 2026-04-28
sched: add MockClock/MockIrqSource behind sched-mock feature


dburkart:
## Motivation

Implements RFC 0005 — Scheduler / IRQ Seam for Deterministic Simulation Testing. Roadmap item 4: provide the first non-production `Clock`/`IrqSource` impl, prove the seam is usable, and convert one existing in-kernel scheduler test as proof-of-life. This is the payoff for the seam — test isolation immediately improves; Phase 2 (host-side simulator) becomes possible.

RFC: `docs/RFC/0005-scheduler-irq-seam.md`
Merged in: #664
Depends on: call-site migration issue (roadmap item 3)

## Work

- [ ] Add `MockClock` to `kernel/src/task/env.rs` behind `#[cfg(any(test, feature = "sched-mock"))]` (no `target_os` exception — verified gate).
- [ ] `MockClock` advances on explicit `tick()` / `advance(n: u64)` calls; owns its own `WAKEUPS` table internally (does not touch `time::WAKEUPS`).
- [ ] `MockIrqSource::ack_timer` is a no-op (records call count for assertions).
- [ ] Add `sched-mock` feature to `kernel/Cargo.toml`, NOT in default-features.
- [ ] Provide a `with_mock_env(|clock, irq| { ... })` test helper that installs the mocks via thread-local OnceCell for the duration of the closure.
- [ ] Convert one existing in-kernel scheduler integration test (e.g. a basic round-robin test) to use `with_mock_env` as proof-of-life. Test runs deterministically without real PIT IRQs.
- [ ] CI green: `cargo xtask build` (no sched-mock), `cargo xtask test` (with sched-mock for unit tests).

## Context

- RFC sections: Design / Test-time wiring
- This unblocks Phase 2 (host-side simulator) — a separate RFC will define the simulator crate; this PR ships the trait impls it will reuse.

PR #698, opened 2026-04-28
ci: nm-check release kernel ELF for MockClock/MockIrqSource symbols (#669)

### LINKED ISSUE #669: ci: nm-check release kernel ELF for MockClock/MockIrqSource symbols
## Motivation

Implements RFC 0005 — Scheduler / IRQ Seam for Deterministic Simulation Testing. Roadmap item 5: make the "physically excluded from production" claim verified rather than asserted. Per Security-researcher review B1, the cfg gate alone is not sufficient evidence — CI must enforce that no `MockClock` or `MockIrqSource` symbol leaks into the release kernel ELF.

RFC: `docs/RFC/0005-scheduler-irq-seam.md`
Merged in: #664
Depends on: MockClock/MockIrqSource issue (roadmap item 4)

## Work

- [ ] Add a CI step to the release-kernel build pipeline that runs `nm` (or `llvm-nm`) over the produced kernel ELF.
- [ ] Step fails the build if any symbol matching `MockClock|MockIrqSource` (or any other `sched-mock`-gated type) is present.
- [ ] Step also runs against the `cargo xtask iso` output as a defense-in-depth check.
- [ ] Document the check in `docs/agent-playbooks/testing.md` so future contributors know not to disable it.
- [ ] (Optional, follow-up) Add a `static_assertions::assert_cfg!`-style compile-time check that fails to build the release kernel if `sched-mock` is on (Security cycle-2 advisory A1).

## Context

- RFC section: Design / Test-time wiring
- This is the regression gate that prevents a careless PR from inadvertently shipping mock code in production.

dburkart:
Closes #669.

## Summary

RFC 0005 Phase 1 hardening — belt-and-suspenders behind the
`#[cfg(feature = "sched-mock")]` source-level gate from #668 (PR #691).
Even if a future PR flips `sched-mock` on by default, or a non-gated
`pub use` re-exports a mock type, this static guard fails the release
pipeline before the binary ships. Implements Security review B1's
ask: the "physically excluded from production" claim is now verified
rather than asserted.

- New `cargo xtask nm-check --release` subcommand: builds the release
  kernel + ISO and runs `llvm-nm --demangle` over the freshly-built
  kernel ELF, failing on any symbol whose demangled name contains
  `MockClock`, `MockIrqSource`, or `MockTimerIrq` (post-#670 rename).
  Falls back to system `nm` when `llvm-tools` isn't on the toolchain.
- Wired into the existing `fmt + build (release)` CI job in
  `.github/workflows/ci.yml`, modeled on the `RFC 0004 vfs_creds gate`
  pattern that landed earlier this sprint.
- Matcher factored as a pure `scan_nm_output_for_mocks` function with
  three host unit tests (clean output, monomorphized impl symbols,
  both irq-mock names) so a regression in the matcher itself fails
  xtask host tests in CI before any kernel build runs.
- Documented in `docs/agent-playbooks/testing.md` so future
  contributors know not to disable the check.

## Verification

- `cargo fmt --all -- --check` clean.
- `cargo clippy -p xtask --all-targets -- -D warnings` clean.
- `cargo test -p xtask` — 62 passed (including 3 new
  `nm_check_*` tests).
- `cargo xtask nm-check --release` locally — release kernel builds,
  ISO builds, no forbidden symbols reported.
- Manually verified the negative case by temporarily forcing a static
  `MockClock` reference into `_start` and rebuilding with
  `--features sched-mock`: `nm --demangle` then emits
  `<vibix::task::env::MockClock as Clock>::{now,drain_expired,enqueue_wakeup}`
  plus the `drop_in_place::<MockClock>` glue, which the matcher
  catches. The temp ref was reverted before commit; the fixture-based
  unit tests pin the matcher behaviour going forward.

## Test plan

- [x] Host unit tests cover the matcher logic (clean output, monomorphized impls, both irq-mock names).
- [x] CI fmt-build job runs `cargo xtask nm-check --release` against the release ELF and the ISO-bundled ELF.
- [x] Manually verified negative case — guard fires when a static `MockClock` reference is forced into `_start` with `--features sched-mock`.

## Follow-ups (deferred)

- Issue #669's optional roadmap item: `static_assertions::assert_cfg!`-style compile-time check that fails to build the release kernel if `sched-mock` is on (Security cycle-2 advisory A1). Not in scope for this PR; can be a separate small follow-up.
- When #670 lands the `MockIrqSource` → `MockTimerIrq` rename, no change is needed here — the matcher already lists both names.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,20 @@ jobs:
       - name: cargo xtask build --release
         run: cargo xtask build --release
 
+      # RFC 0005 Phase 1 hardening (#669). Static guard that the
+      # `#[cfg(feature = "sched-mock")]` gate around `MockClock` /
+      # `MockIrqSource` from #668 actually holds: even if a future PR
+      # accidentally enables `sched-mock` in the default feature set or
+      # a non-gated `pub use` re-exports a mock type, this step fails
+      # the release pipeline before the binary ships. Runs `llvm-nm`
+      # over the freshly-built release kernel ELF (and the ISO-bundled
+      # ELF as defense-in-depth) and fails on any forbidden symbol.
+      #
+      # See `docs/RFC/0005-scheduler-irq-seam.md` "Test-time wiring"
+      # and Security review B1.
+      - name: RFC 0005 nm-check (no MockClock/MockIrqSource in release)
+        run: cargo xtask nm-check --release
+
   unit-tests:
     name: host unit tests
     runs-on: ubuntu-latest
--- a/docs/agent-playbooks/testing.md
+++ b/docs/agent-playbooks/testing.md
@@ -85,6 +85,22 @@ yet, so the runner boots as a placeholder. The CI gate still earns its keep by f
 on harness regressions (kernel panic, QEMU watchdog timeout, ext2 image builder
 break). Real conformance signal arrives when #642 ports the runner to vibix userspace.
 
+## Layer 5 - RFC 0005 nm-check (release-symbol guard)
+
+`cargo xtask nm-check --release` builds the release kernel and ISO and runs
+`llvm-nm` over the resulting ELF, failing if any `MockClock` / `MockIrqSource`
+symbol (or any other `sched-mock`-gated type) appears. This is the static guard
+that the `#[cfg(feature = "sched-mock")]` source-level gate around the mocks in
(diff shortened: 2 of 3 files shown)


dburkart:
@coderabbitai review

dburkart:
@coderabbitai review

PR #703, opened 2026-04-28
tests/ext2: malformed rec_len regression at the getdents64 surface (#682)

### LINKED ISSUE #682: ext2: regression test for malformed rec_len sandboxing
## Motivation

`DirEntryIter` already rejects malformed `rec_len` (overrun, unaligned) by surfacing `DirError::Corrupt` rather than panicking or reading OOB — the in-module unit tests at `kernel/src/fs/ext2/dir.rs:633-660` cover the iterator-level behaviour. We do not have a higher-level integration test that mounts a hand-crafted image with a corrupt directory and asserts the syscall path (`getdents64` over the corrupt dir) returns `EIO` cleanly without panicking.

## Work

- [ ] Add `kernel/tests/ext2_malformed_dir.rs` that builds a minimal ext2 image where the root directory contains a record with `rec_len` exceeding the block, mounts it, and runs `getdents64`.
- [ ] Assert: syscall returns `EIO`, no panic, the fs remains usable for siblings of the bad directory.
- [ ] Cover the unaligned-`rec_len` and zero-`rec_len` cases as parameterised variants.
- [ ] Use the shared `ext2_ramdisk` helper.

## Out of scope

- Repairing or skipping past the bad record (current policy is to surface the error; skipping is an fsck question).

## Context

- Parent epic: #537.
- Iterator-level tests: `kernel/src/fs/ext2/dir.rs:633-660`.
- Files: `kernel/src/fs/ext2/dir.rs`, `kernel/src/fs/ext2/getdents.rs` (or wherever the dirent syscall surface lives).

dburkart:
Closes #682.

## Summary

- Adds `kernel/tests/ext2_malformed_dir.rs`, a QEMU integration test that mounts the golden ext2 fixture with a surgically-corrupted root directory block and asserts `dir::getdents64` returns `EIO` without panicking.
- Three variants cover the same `rec_len` defects already exercised at the `DirEntryIter` level (`kernel/src/fs/ext2/dir.rs:633-660`): overrun (rec_len longer than the block), unaligned (rec_len % 4 != 0), and zero (a self-loop the iterator must reject instead of spinning).
- A fourth case mounts a poisoned image and confirms `iget` still succeeds for the sibling `lost+found` inode (ino 11) — root-dir corruption must not poison unrelated inodes.
- Wires the new binary into `kernel/Cargo.toml` (`required-features = ["ext2"]`).

The test discovers the absolute byte offset of the root dir's first data block dynamically (mount golden, read root inode's `i_block[0]`, derive `block_size`) so the fixture path stays robust if the golden image is ever regenerated. Each variant builds a fresh `RamDisk` + `Ext2Fs` so the per-mount `BlockCache` cannot return stale (pre-patch) bytes.

Per the issue's "out of scope" note: this asserts the current "surface error" policy. Skipping past the bad record is fsck territory and not introduced here.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo xtask build`
- [x] `cargo test ... --test ext2_malformed_dir` — all 4 variants pass under QEMU on the worktree.
- [ ] Full CI on this PR.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -418,6 +418,11 @@ harness = false
 required-features = ["ext2"]
 
 [[test]]
+name = "ext2_malformed_dir"
+harness = false
+required-features = ["ext2"]
+
+[[test]]
 name = "ext2_ialloc"
 harness = false
 required-features = ["ext2"]
--- a/kernel/tests/ext2_malformed_dir.rs
+++ b/kernel/tests/ext2_malformed_dir.rs
@@ -1,0 +1,348 @@
+//! Integration test for issue #682: `getdents64` over a directory whose
+//! first data block carries a structurally-corrupt record must surface
+//! `EIO` cleanly without panicking, and the rest of the filesystem must
+//! remain usable.
+//!
+//! Iterator-level coverage of the same `rec_len` defects already exists
+//! at `kernel/src/fs/ext2/dir.rs:633-660`. This test exercises the
+//! higher-level integration: mount a real ext2 image whose root
+//! directory block has been surgically corrupted, then drive
+//! `dir::getdents64` against it and assert
+//!
+//! 1. the syscall path returns `EIO` (no panic, no OOB read);
+//! 2. an unrelated live inode (the on-disk `lost+found` at ino 11) is
+//!    still readable through `iget` — corruption in the root dir does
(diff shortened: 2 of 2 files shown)


dburkart:
@coderabbitai review

PR #704, opened 2026-04-28
vfs: harden FakeFs lookup stub with POSIX ENOTDIR semantics

### LINKED ISSUE #261: vfs: harden FakeFs test stub with POSIX-compliant ENOTDIR semantics
## Motivation

Discovered while merging #260 (vfs path_walk). The host-side `FakeOps::lookup` in `kernel/src/fs/vfs/path_walk.rs` always maps missing children to `ENOENT` and never inspects the parent inode's kind, so it does not exercise the POSIX §4.13 `ENOTDIR` contract that real FS drivers must honor. This is fine for the `path_walk` tests today (the production resolver checks `InodeKind::Dir` on the parent before invoking `lookup`), but it leaves a gap in coverage for FS implementations that may rely on the lookup-level check.

## Work

- [ ] Update `FakeOps::lookup` in `kernel/src/fs/vfs/path_walk.rs` (`#[cfg(test)]` block) to return `ENOTDIR` when called on a non-directory parent inode (e.g., the existing fixture entry `notadir` at `ino=16` with `InodeKind::Reg`).
- [ ] Add a unit test asserting that walking `/notadir/foo` returns `ENOTDIR`.
- [ ] Audit any other test stubs across `kernel/src/fs/vfs/*.rs` for similar gaps and align them with POSIX `lookup` error semantics.

## Context

- Reported by CodeRabbit on PR #260 (https://github.com/dburkart/vibix/pull/260).
- Files: `kernel/src/fs/vfs/path_walk.rs` (the `FakeOps`, `FakeFs`, and tests at the bottom of the file).
- Related: #232 (parent issue for path_walk), #233 (mount table — when a real driver lands, this hardening pairs naturally with FakeFs evolution).
- Test-only fidelity, no user-visible defect.

dburkart:
Closes #261

## Summary

- `FakeOps::lookup` (the host-side test stub in `kernel/src/fs/vfs/path_walk.rs`) now returns `ENOTDIR` when called on a non-directory parent inode, matching the POSIX §4.13 contract that real FS drivers must honor.
- Adds a `walk_through_nondir_parent_is_enotdir` unit test that walks `/notadir/foo` (where `/notadir` is the existing `ino=16`, `InodeKind::Reg` fixture entry) and asserts the result is `Err(ENOTDIR)` rather than `Err(ENOENT)`.
- Audited the other `InodeOps` test stubs across `kernel/src/fs/vfs/*.rs` (`backend.rs`, `dentry.rs`, `gc_queue.rs`, `inode.rs`, `mount_table.rs`, `open_file.rs`, `ops.rs`). None override `lookup`, so they inherit the trait default (`Err(ENOENT)`). That default is fine because none of those fixtures feeds a non-directory parent into a walk — only `path_walk`'s `FakeOps` did, and only that one needed hardening.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo xtask build` (clean — only pre-existing dead-code warnings)
- [x] `cargo xtask smoke` (all 41 markers present)
- [x] `cargo xtask test` — host-side `--lib` suite passes (544/544); the kernel-target integration suite is unchanged and surfaces the same pre-existing local flake (`wait4_condvar_race` 90 s timeout) noted in the spawn brief; CI on PR branches passes cleanly.

### DIFF
--- a/kernel/src/fs/vfs/path_walk.rs
+++ b/kernel/src/fs/vfs/path_walk.rs
@@ -458,6 +458,14 @@ mod tests {
     struct FakeOps(Arc<FakeFs>);
     impl InodeOps for FakeOps {
         fn lookup(&self, dir: &Inode, name: &[u8]) -> Result<Arc<Inode>, i64> {
+            // POSIX §4.13: lookup must reject a non-directory parent
+            // with ENOTDIR before any name resolution. Real FS drivers
+            // are expected to honor this contract; a test stub that
+            // unconditionally returned ENOENT would mask a missing
+            // kind check elsewhere in the resolver chain.
+            if dir.kind != InodeKind::Dir {
+                return Err(ENOTDIR);
+            }
             self.0
                 .children
                 .lock()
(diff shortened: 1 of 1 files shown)


dburkart:
@coderabbitai review

PR #706, opened 2026-04-28
shell: add file-touching builtins (cd, pwd, ls, cat, mkdir, rm, touch, mv, cp, stat, rmdir)

### LINKED ISSUE #395: shell: add file-touching builtins (cd, pwd, ls, cat, mkdir, rm, touch, mv, cp, stat)
## Motivation

The kernel-resident shell at `kernel/src/shell/mod.rs` has 8 builtins today (`help`, `uptime`, `time`, `mem`, `tasks`, `pci`, `echo`, `panic`) — all kernel introspection, none that touch the filesystem we just mounted. A user who boots vibix can see how much memory is free but cannot list `/dev`, read `/dev/zero`, or change directories. Every syscall needed already exists (`getcwd`, `chdir`, `getdents64`, `open`, `read`, `stat`). Wiring them into shell builtins is the single biggest jump in perceived completeness available right now.

This stays in the kernel-resident shell deliberately — porting the shell to userspace is tracked separately and depends on the P1 termios/pipes stack landing first.

## Work

- [ ] `cd <path>` — wraps `chdir`; updates the prompt's CWD if/when the prompt grows one.
- [ ] `pwd` — wraps `getcwd`.
- [ ] `ls [path]` — uses `getdents64`; default to CWD; print one entry per line; flag-free in v1.
- [ ] `cat <path>...` — `open` + `read` loop to the shell's output sink; handle EOF, ENOENT, EISDIR.
- [ ] `mkdir <path>` — single-level only in v1 (no `-p`).
- [ ] `rmdir <path>` — fails on non-empty.
- [ ] `rm <path>` — file unlink; no recursive flag in v1.
- [ ] `touch <path>` — `open(O_CREAT)` + `close`; no mtime update.
- [ ] `mv <src> <dst>` — uses `rename` if available, else `link`+`unlink`. Note in PR if `rename` syscall is missing.
- [ ] `cp <src> <dst>` — naive read-loop copy; no flag handling.
- [ ] `stat <path>` — print `stat` struct fields in a Linux-like format.
- [ ] Per-builtin help text (consumed by the per-command `help <cmd>` issue once that lands).
- [ ] Integration test: `kernel/tests/shell_files.rs` — boot, run a script of these commands against `/dev` and `/tmp`, assert expected output via serial markers.

## Context

- `kernel/src/shell/mod.rs` — current builtin dispatch
- `kernel/src/shell/line_editor.rs` — line-editing layer (untouched by this issue)
- Existing syscalls: `kernel/src/syscall/` — `chdir`, `getcwd`, `getdents64`, `open`, `read`, `stat`, `lstat` already implemented
- Companion issues to be filed: shell introspection commands, MOTD/banner, `/proc` skeleton, per-command help
- Userspace shell port is deliberately deferred — depends on #370 (pipes), #373 (termios), #375 (N_TTY), #376 (job control)
- Surfaced during a userspace-completeness survey on 2026-04-15

dburkart:
Closes #395

## Summary
- Adds 11 new kernel-resident shell builtins (`cd`, `pwd`, `ls`, `cat`, `mkdir`, `rmdir`, `rm`, `touch`, `mv`, `cp`, `stat`) that poke the VFS directly via a new `kernel/src/shell/vfs_helpers.rs` module — small wrappers around `path_walk`, `OpenFile`, `getdents`, and the inode ops.
- Each builtin gets a one-line `help` entry in the existing dispatch table.
- New integration test `kernel/tests/shell_files.rs` drives every builtin through `dispatch_for_test` and verifies side effects via direct VFS observation (the 16-byte 16550 RX FIFO is too small for serial loopback on the larger commands — same convention as `shell_help`).

## Notes for reviewers
- `cd` lexically normalizes its argument (resolving `.` and `..` at the string level against `task::current_cwd`) before handing it to `path_walk`. The kernel's `step_dotdot` only crosses one mount edge per `..` and stops at the parent FS's mountpoint dentry without taking its parent — without this normalization, `cd ..` from `/tmp` lands back at `/tmp`. Lexical normalization side-steps the issue without changing VFS semantics. Worth a follow-up issue to fix `step_dotdot` directly.
- `dentry_path` (used by `pwd`) detects mount-root dentries via `DFlags::IS_ROOT` and crosses up through `GlobalMountResolver::mount_above` to pick up the user-visible mountpoint name. A naive parent walk would terminate at the `.` sentinel name on a mount root and produce `/.` for `/tmp`.
- `mv` tries `InodeOps::rename` first and falls back to `link` + `unlink` on `EPERM` so it works on filesystems whose drivers don't implement rename yet (currently the rootfs tarfs).
- The shell runs as `Credential::kernel()` so all permission checks pass; relative paths resolve against `task::current_cwd`.

## Test plan
- [x] `cargo xtask build`
- [x] `cargo xtask test` — all green; new `shell_files` test runs 9 cases (`pwd_and_cd_roundtrip`, `mkdir_then_rmdir`, `touch_stat_rm`, `cp_roundtrip`, `mv_renames`, `cat_smokes`, `ls_dev`, `cd_nonexistent_errors`, `rmdir_nonempty_errors`).
- [x] `cargo xtask smoke` — all 41 markers present.
- [x] `cargo fmt --all`

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -303,6 +303,10 @@ name = "shell_help"
 harness = false
 
 [[test]]
+name = "shell_files"
+harness = false
+
+[[test]]
 name = "syscall_mkfifo"
 harness = false
 
--- a/kernel/src/shell/mod.rs
+++ b/kernel/src/shell/mod.rs
@@ -12,6 +12,8 @@
 pub mod ansi;
 pub mod banner;
 pub mod line_editor;
+#[cfg(target_os = "none")]
+pub mod vfs_helpers;
 
 #[cfg(target_os = "none")]
 mod kernel_side {
@@ -190,12 +192,39 @@ mod kernel_side {
     /// a host-side unit test.
     const BUILTINS: &[Builtin] = &[
         Builtin {
+            name: "cat",
+            run: cmd_cat,
+            help: BuiltinHelp {
+                summary: "concatenate file contents to stdout",
+                usage: "cat <path>...",
+                examples: &["cat /etc/motd", "cat /dev/zero"],
+            },
+        },
+        Builtin {
+            name: "cd",
+            run: cmd_cd,
+            help: BuiltinHelp {
+                summary: "change the shell's working directory",
+                usage: "cd <path>",
+                examples: &["cd /tmp", "cd ..", "cd /"],
+            },
+        },
+        Builtin {
             name: "clear",
             run: cmd_clear_wrap,
             help: BuiltinHelp {
                 summary: "clear the screen (VT100)",
                 usage: "clear",
                 examples: &["clear"],
+            },
+        },
+        Builtin {
+            name: "cp",
+            run: cmd_cp,
+            help: BuiltinHelp {
+                summary: "copy a file (naive read-loop, no flags)",
+                usage: "cp <src> <dst>",
+                examples: &["cp /etc/motd /tmp/motd.bak"],
             },
         },
         Builtin {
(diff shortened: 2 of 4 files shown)


dburkart:
Thanks for the review. Addressing the same-file cp guard and read_all cap in 7e507d8.

Deferring the dangling-symlink ENOENT-fallback finding to a follow-up — fixing it correctly needs a NOFOLLOW probe distinguishing 'final component is a dangling symlink' from 'final component is missing', and the current callers (write_all and stream_copy) only run from a kernel-resident debug shell where the user is root and the dst paths are not symlinks in any of the test or smoke flows. Will file an issue for it.

PR #707, opened 2026-04-28
ext2: ENOSPC stress test (block + inode bitmap exhaustion + recovery)

### LINKED ISSUE #681: ext2: ENOSPC stress test (block + inode bitmap exhaustion + recovery)
## Motivation

The block and inode allocators land with bitmap- and BGDT-tracking, but there is no test that drives either bitmap to full, asserts the correct `ENOSPC`/`EDQUOT`-equivalent error surfaces at the syscall boundary, and verifies that subsequent unlinks free space cleanly so the FS recovers. Allocator off-by-ones at the boundary are easy to introduce and easy to miss without a stress test.

## Work

- [ ] Add `kernel/tests/ext2_enospc.rs` with two scenarios on a small ext2 image (1 MB, low inode count):
  - **Block exhaustion**: write a single growing file until `EFBIG` or `ENOSPC`; assert correct error; unlink; assert the fs accepts new writes up to the same limit.
  - **Inode exhaustion**: `create` files until `ENOSPC` from `ialloc`; unlink half; assert subsequent creates succeed.
- [ ] Verify partial-write behaviour: a write that hits ENOSPC mid-extend leaves the file at the high-water mark with consistent metadata (no orphan blocks, `i_size` matches the actually-allocated logical extent).
- [ ] Verify orphan-chain inode is reclaimed correctly on the inode-exhaustion path even if it shares a group with the exhausted bitmap.
- [ ] Use the shared `ext2_ramdisk` helper.

## Out of scope

- Disk quotas (`EDQUOT`) — ext2 has no quota subsystem in this driver.
- Reservation windows — optional ext2 feature, not implemented.

## Context

- Parent epic: #537.
- Files: `kernel/src/fs/ext2/balloc.rs`, `kernel/src/fs/ext2/ialloc.rs`, `kernel/src/fs/ext2/file.rs` (write extend), `kernel/src/fs/ext2/create.rs`.

dburkart:
Closes #681.

## Summary

Adds `kernel/tests/ext2_enospc.rs` exercising two ENOSPC scenarios per the issue body:

- **Block exhaustion** (`balloc_test.img`, 512 KiB / 2 groups × 256 × 1 KiB blocks): create a regular file under `/`, write one block, drain the rest of the bitmap via `alloc_block`, then drive `FileOps::write` past EOF and assert `ENOSPC` / `EFBIG` / short-write. Verifies the partial-write metadata invariants from RFC 0004 §Write Ordering: `i_size` never claims bytes past the actually-allocated extent, `i_blocks` covers at least the data written. Then unlinks through the VFS trait, finalizes the orphan, asserts `s_free_blocks_count` recovers, and confirms a fresh `create` + write succeeds.
- **Inode exhaustion** (`golden.img`, 64 KiB / 16 inodes / 5 free): loops `create` until `ENOSPC`, unlinks half, finalizes them, then re-creates. Cross-checks the issue's "orphan-chain inode is reclaimed correctly even if it shares a group with the exhausted bitmap" — on the single-group golden image every freshly-allocated ino sits in the same group as the surviving allocations.

Both tests cross-check that the orphan-list pin and `s_last_orphan` chain are empty at the end (no reclaim leak). All operations route through the VFS-trait dispatch (`InodeOps::create` / `InodeOps::unlink` / `FileOps::write`) so the cached parent `Ext2Inode` stays consistent across `add_link` growth — a fabricated parent leaves the cached one stale and `locate_dirent` then misses entries that landed in a freshly-allocated dir block.

The block-exhaustion test drains via `alloc_block` rather than purely through `FileOps::write` so the bitmap reaches zero deterministically without piling MiBs of write traffic through the buffer cache. The issue's primary contract — "the FS reports ENOSPC at the syscall boundary and recovers" — is exercised through the production `FileOps::write` extend path on top of the drained bitmap.

Uses the shared `ext2_ramdisk` helper per the issue.

## Test plan

- [x] `cargo xtask test` — both new scenarios pass; only failure is the documented pre-existing `wait4_condvar_race` host-timeout flake on main
- [x] `cargo xtask smoke` — all 41 markers present
- [x] `cargo fmt --all` applied before commit

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -423,6 +423,11 @@ harness = false
 required-features = ["ext2"]
 
 [[test]]
+name = "ext2_enospc"
+harness = false
+required-features = ["ext2"]
+
+[[test]]
 name = "ext2_ialloc"
 harness = false
 required-features = ["ext2"]
--- a/kernel/tests/ext2_enospc.rs
+++ b/kernel/tests/ext2_enospc.rs
@@ -44,7 +44,7 @@ use alloc::vec::Vec;
 use core::panic::PanicInfo;
 
 use vibix::block::BlockDevice;
-use vibix::fs::ext2::{alloc_block, finalize_orphan, iget, Ext2Fs, Ext2Super};
+use vibix::fs::ext2::{alloc_block, finalize_orphan, free_block, iget, Ext2Fs, Ext2Super};
 use vibix::fs::vfs::dentry::Dentry;
 use vibix::fs::vfs::inode::Inode;
 use vibix::fs::vfs::open_file::OpenFile;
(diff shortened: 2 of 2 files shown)


dburkart:
@coderabbitai review

PR #712, opened 2026-04-28
RFC 0006: Host-Side Deterministic Simulator (DST Phase 2)

dburkart:
## Summary

RFC 0006 proposes a host-side `simulator/` crate that consumes the scheduler / IRQ seam landed in RFC 0005 (#664) to drive the kernel scheduler one tick at a time under seeded `MockClock` + `MockTimerIrq`, exposes a `(Tick, Event)` trace stream as the state-machine model interface, and turns flakes like #501, #527, and #478 into reproducible `--seed=…` failures.

This PR contains a design document only — no kernel code changes. The RFC will be reviewed by the scoped archetype panel (the four defaults — security researcher, OS engineer, user-space staff engineer, academic — plus `testing-infrastructure-engineer` for harness ergonomics / reproducibility-envelope review), revised until all blocking findings are addressed (cap: 3 defense cycles), then merged. Implementation issues are filed from the roadmap section after merge.

## RFC sections

- Abstract
- Motivation (with concrete P0 flakes #501 / #527 / #478)
- Background (10-source prior-art table: VOPR, FoundationDB, sled, Polar Signals, loom, shuttle, Hermit, Antithesis, madsim, proptest-state-machine)
- Design
  - Crate boundary: single-binary cfg-gated (defended vs. separate-binary FFI)
  - Reuse of existing `MockClock` / `MockTimerIrq` (no new trait methods)
  - Linear tick-by-tick driver (defended vs. event-driven)
  - `(Tick, Event)` trace as model interface (defended vs. snapshot/diff API)
  - `proptest-state-machine`-compatible reference state machine
  - Failure-injection scope (timer jitter, spurious IRQs, wakeup reorder; hardware faults / syscall entry / IPI explicitly Phase 2.1)
  - Reproducibility envelope (seed + git commit + pinned nightly + features)
  - Reproduction commitments per P0 flake
- Security Considerations
- Performance Considerations
- Alternatives Considered (separate-binary FFI, event-driven, snapshot API, extending traits speculatively, loom/shuttle direct, wait-for-SMP, stateright)
- Open Questions
- Implementation Roadmap (11 dependency-ordered items)

## Test plan

- [ ] All blocking findings are addressed within 3 defense cycles.

### DIFF
--- a/docs/RFC/0006-host-dst-simulator.md
+++ b/docs/RFC/0006-host-dst-simulator.md
@@ -1,7 +1,7 @@
 ---
 rfc: 0006
 title: Host-Side Deterministic Simulator (DST Phase 2)
-status: Draft
+status: In Review
 created: 2026-04-27
 ---
 
--- a/docs/RFC/0006-host-dst-simulator.md
+++ b/docs/RFC/0006-host-dst-simulator.md
@@ -1,7 +1,7 @@
 ---
 rfc: 0006
 title: Host-Side Deterministic Simulator (DST Phase 2)
-status: In Review
+status: Accepted
 created: 2026-04-27
 ---
 
@@ -1104,6 +1104,27 @@ randomized harness.
 
 All open questions are deferred to implementation per the resolutions
 recorded below. None block RFC acceptance.
+
+Post-defense advisory items absorbed (defense cycle 1):
+
+- *VIBIX_SIM_SEED= panic-hook string is a stable interface* (Sec-A1
+  re-review) — call out explicitly in the implementation that this
+  string is greppable forever; document in §"Reproducibility envelope"
+  and `docs/design/simulator.md`.
+- *`preempt_tick` re-entrancy guard should be active under all
+  `sched-mock` builds, not just `cfg(debug_assertions)`* (OS-A1
+  re-review) — implementation uses `cfg(any(debug_assertions, all(test,
+  feature = "sched-mock")))` (or simpler: always-on when `sched-mock`).
+- *`--explain` agent-stable output format* (UserSpace-A1 re-review) —
+  ship `--explain=table|json|tsv` with `tsv` as the agent-stable form.
+- *Custom `Ord` impls in `simulator/` and sched-mock-gated kernel code
+  must be reviewed for determinism* (Academic-A1 re-review) — add to
+  the `docs/design/simulator.md` checklist.
+- *Nightly suite path filter should include `simulator/**`*
+  (Testing-A1 re-review) — extend the filter in Roadmap item 9.
+
+These are implementation conveniences, not design changes; they ride
+the relevant Roadmap items.
 
 - **Pinning a specific nightly date.** The Reproducibility Envelope
   requires it. *Lean: pin in Roadmap item 1 to whatever nightly is
--- a/docs/RFC/0006-host-dst-simulator.md
+++ b/docs/RFC/0006-host-dst-simulator.md
@@ -142,6 +142,180 @@ vibix/
 └── ...
 ```
 
+### Host-target buildability of `kernel/` (resolves OS-engineer B1)
+
+Before any of the architectural choices below mean anything, the
+`kernel/` crate must compile for the host triple under `--features
+sched-mock`. Today it does not — `kernel/src/task/env.rs` gates
+`HwIrq`, `HW_IRQ`, `env()`, and `assert_production_env()` on
+`#[cfg(target_os = "none")]`. Phase 2 commits to **option (a)** from
+the OS-engineer review:
+
+A parallel `#[cfg(all(not(target_os = "none"), feature =
+"sched-mock"))] fn env()` is added to `kernel/src/task/env.rs`
+returning the mock pair. The mocks are installed into a thread-local
+`OnceCell<(&'static MockClock, &'static MockTimerIrq)>` by the
+simulator's `Simulator::with_seed` constructor before the first
+kernel call:
+
+```rust
+// kernel/src/task/env.rs, sched-mock + host build
+#[cfg(all(not(target_os = "none"), feature = "sched-mock"))]
+thread_local! {
+    static SIM_ENV: core::cell::OnceCell<(&'static dyn Clock, &'static dyn TimerIrq)>
+        = core::cell::OnceCell::new();
+}
+
+#[cfg(all(not(target_os = "none"), feature = "sched-mock"))]
+pub fn env() -> (&'static dyn Clock, &'static dyn TimerIrq) {
+    SIM_ENV.with(|c| *c.get().expect("simulator must call install_sim_env() first"))
+}
+
+#[cfg(all(not(target_os = "none"), feature = "sched-mock"))]
+pub fn install_sim_env(clock: &'static dyn Clock, irq: &'static dyn TimerIrq) {
+    SIM_ENV.with(|c| c.set((clock, irq)).map_err(|_| "already installed").unwrap());
+}
+```
+
+Thread-local (not global) because parallel `cargo test` workers
+each get their own `Simulator` with their own seed — global state
+would serialize them. Each test thread installs its own mock pair
+once and panics if installed twice.
+
+The follow-on work the buildability requires:
+
+1. **Audit and gate `target_os = "none"` callers in
+   `kernel/src/task/`, `kernel/src/sync/`, `kernel/src/signal/`,
+   `kernel/src/syscall/`** under a parallel
+   `cfg(all(not(target_os = "none"), feature = "sched-mock"))` arm
+   that selects a host-buildable substitute. Concretely:
+   - `IrqLock<T>` → `spin::Mutex<T>` on host (matches what
+     `MockClock` already does, RFC 0005 wave 3).
+   - `arch::ack_timer_irq()` is unreachable on host because
+     `HwIrq` doesn't exist; only `MockTimerIrq::ack_timer` is
+     callable. No substitute needed — the call site is reached only
+     via `env()` which returns a mock.
+   - Inline assembly (`cli`/`sti`/IRETQ etc.) is gated to
+     `target_os = "none"` already; under `sched-mock` host build,
+     the no-op host alternative is selected.
+2. **A new CI job — `cargo build -p kernel --target
+   x86_64-unknown-linux-gnu --features sched-mock`** — gates that
+   the host build keeps compiling. It runs alongside the existing
+   bare-metal build and the nm-check.
+
+This is real engineering work, not RFC handwaving — Roadmap item 1
+is rewritten below to land the host-buildability slice **before**
+any simulator code exists, so the rest of Phase 2 builds on a
+verified host-buildable kernel.
+
+### IRQ-context invariants under host driver (resolves OS-engineer B2)
+
+The simulator runs `kernel::task::preempt_tick()` from an ordinary
+host thread; it is not in IRQ context. Two invariants must be
+preserved:
+
+1. **No re-entrancy of `preempt_tick`.** The simulator MUST NOT call
+   `preempt_tick` recursively. The driver loop (§Driver loop)
+   already has this property — it calls `preempt_tick` exactly once
+   per outer-loop iteration, and the FaultPlan's `inject` step runs
+   *before* that call returns to the loop top. If a future
+   FaultPlan grows a callback that fires during
+   `MockClock::drain_expired`, that callback MUST NOT call
+   `preempt_tick`. Enforced by a `#[cfg(debug_assertions)]`
+   re-entrancy guard inside `preempt_tick` itself, gated on
+   `feature = "sched-mock"`:
+
+   ```rust
+   #[cfg(feature = "sched-mock")]
+   thread_local! { static IN_PREEMPT_TICK: Cell<bool> = Cell::new(false); }
+   #[cfg(feature = "sched-mock")]
+   fn preempt_tick_guard() -> impl Drop {
+       IN_PREEMPT_TICK.with(|f| {
+           assert!(!f.get(), "preempt_tick is not re-entrant");
+           f.set(true);
+       });
+       scopeguard::guard((), |_| IN_PREEMPT_TICK.with(|f| f.set(false)))
+   }
+   ```
+
+2. **Host substitutes for IRQ-disable primitives must preserve
+   ordering, not faithfully imitate IRQ masking.** `IrqLock` on host
+   is `spin::Mutex` — fair-ish, contentious enough that lock-order
+   violations surface as deadlocks. `cli`/`sti` becomes no-op
+   (impossible to imitate without a kernel; the simulator's value is
+   in checking ordering invariants the *seam* exposes, not in
+   imitating ring-0 microarchitecture). The RFC commits to:
+   - A list, in `docs/design/simulator.md`, of the IRQ-related
+     primitives that have host substitutes and what those
+     substitutes are. New primitives added to the kernel must
+     declare a host substitute (or document why they're
+     `target_os = "none"`-only and therefore unreachable under the
+     simulator).
+   - The simulator does **not** claim to find bugs that depend on
+     real `cli`/`sti` semantics (e.g. interrupt-window races inside
+     the timer ISR itself). Those remain in QEMU's domain. The
+     simulator's contract is: *if a fault-injected sequence of
+     scheduler-visible events would fail on real hardware, the
+     simulator catches it.* Bugs whose trigger is below the seam
+     (microarchitectural reordering, IRETQ stack-faulting, etc.)
+     are out of scope and route to Phase 2.1's IDT/syscall-seam
+     RFCs.
+
+This boundary is what classifies #527 as deferred to Phase 2.1
+(§Reproduction commitments below) — the residual `#PF` is
+microarchitectural, below the seam.
+
+### Event emit points (resolves OS-engineer B3)
+
+The Event enum has two classes of variants distinguished by how
+they're populated:
+
+**(a) Observed-by-snapshot** (no kernel-side emit point needed):
+- `TimerInjected`, `SpuriousTimerInjected`, `ClockAdvanced` — all
+  driven by the simulator itself; emitted by the driver loop
+  before/after stepping the kernel.
+- `Switch { from, to }` — derived by snapshotting `current_id()`
+  before and after each `preempt_tick` and emitting if it changed.
+  No kernel-side call needed.
+- `TaskReady`, `TaskBlocked` — derived by diffing
+  `task::scheduler::ready_count()` /
+  `task::scheduler::pending_blocked()` snapshots between ticks.
+- `TaskWoken` — derived from the `Vec<TaskId>` returned by
+  `MockClock::drain_expired`, which the simulator already calls
+  through directly.
+
+**(b) Emit-point-required** (kernel emits a tracing call):
+- `Fork`, `Exec`, `Exit`, `Wait`, `SignalDelivered` — these are
+  syscall-handler-level events. The kernel adds a single
+  `sched_mock_trace!(Event::Fork { … })` macro at each emit point.
+
+The macro expands to a no-op when `feature = "sched-mock"` is off
+(zero-cost in production):
+
+```rust
+// kernel/src/sched_mock_trace.rs
+#[cfg(feature = "sched-mock")]
+#[macro_export]
+macro_rules! sched_mock_trace {
+    ($e:expr) => { $crate::sched_mock_trace::push($e) };
+}
+#[cfg(not(feature = "sched-mock"))]
+#[macro_export]
+macro_rules! sched_mock_trace {
+    ($e:expr) => { () };
+}
+
+#[cfg(feature = "sched-mock")]
+pub fn push(event: Event) { /* thread-local Vec, drained by simulator */ }
+```
+
+Verification that production-build emits zero code is by inspecting
+generated assembly for at least one canonical site (`fork` syscall)
+in the same CI job that runs the nm-check, and is gated as a
+verifiable invariant rather than an assertion. Roadmap item 5 is
+extended below to land this macro infrastructure alongside the
+introspection helpers.
+
 ### Crate boundary: single-binary, sched-mock-feature, no separate kernel-FFI
 
 This is the load-bearing architectural choice. The two viable shapes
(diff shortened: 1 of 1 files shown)


dburkart:

**Summary:** The privilege-model story is fundamentally sound — Phase 2 is a host-only test artifact with no new userspace surface — but the nm-check guard is the entire load-bearing claim that mocks/introspection don't reach release ELFs, and the RFC under-specifies it for the new helpers it introduces.

**Blocking findings:**
- [B1] nm-check regex commitment is hand-wavy — RFC §Security says "extend the regex it greps for" but never commits to a concrete list. The introspection helpers (`task::scheduler::ready_count`, `task::scheduler::pending_blocked`, future `sim_introspect_*` shapes per Open Question 3) are new exports gated only by a Cargo feature, and a typo in the cfg attribute is silent until it ships. The PR that lands them must commit (in this RFC) to a regex that matches `(Mock[A-Z]|sim_introspect_|sched_mock_)` or similar, **and** that regex must be enforceable against name-mangled Rust symbols (the demangled form contains the module path; the mangled form may not). Without this commitment the "physically excluded from production" claim is an assertion again, not a verified invariant — RFC 0005 §"Test-time wiring" was explicit that the nm-check converts the assertion into a verified invariant; Phase 2 must preserve that property as it grows the surface.
- [B2] `--allow-dirty` undermines the trace's commit-hash attribution. The RFC says replay refuses to run on a dirty tree unless `--allow-dirty` is passed, with the trace recording the commit hash. But a dirty-tree run *also* records that same commit hash, so a trace dump cannot distinguish "this trace came from commit `abc123`" from "this trace came from `abc123` plus three uncommitted prints I forgot about." When that trace is later cited in a bug report (which is the whole point of the harness), the commit hash is meaningless if `--allow-dirty` was used. Either (a) the trace records `tree_dirty: true` alongside the commit hash and consumers know to distrust dirty traces, or (b) `--allow-dirty` is forbidden on CI and locally produces a watermark in the trace filename. Pick one and write it down.

**Advisory findings:**
- [A1] Trace JSON should also record the active feature set's *resolved* form (output of `cargo metadata --format-version=1 | jq …`), not just the names. A misnamed feature flag is silent today.
- [A2] The simulator binary should print the seed on every panic/abort path, including SIGINT — CI capture-on-failure must surface the seed in the failure output, not just in a separate artifact file that may be lost.

dburkart:
### Review: OS Engineer

**Summary:** The architectural choice (single-binary, cfg-gated, reuse the RFC-0005 mocks) is correct and well-defended, but the RFC papers over how the kernel actually *builds* for the host triple in the first place. Today `kernel/` is gated on `target_os = "none"` in many places (including the production half of `task::env`); the seam itself only exposes mocks under `feature = "sched-mock"`. There is no path described from "`cargo build -p kernel --target x86_64-unknown-linux-gnu --features sched-mock`" to "compiles." Without that path the entire crate-boundary argument collapses.

**Blocking findings:**
- [B1] **Host-target buildability of `kernel/` is unspecified.** The RFC's chosen architecture requires `simulator/` to depend on the `kernel` crate as a normal Cargo dependency and call `kernel::task::preempt_tick()` directly. But `kernel/src/task/env.rs` already has `#[cfg(target_os = "none")]` on `HwIrq`, `HW_IRQ`, `env()`, and `assert_production_env()`. On the host these symbols don't exist, so `env()` is unresolved, and `preempt_tick` (per RFC 0005's pseudo-code) calls `env()`. The RFC must commit to one of: (a) a parallel `#[cfg(all(not(target_os = "none"), feature = "sched-mock"))] fn env()` in `kernel/src/task/env.rs` that returns the mocks (and an explicit story for how the simulator *installs* the mock instances at runtime — `OnceCell`? thread-local? the RFC waves at "thread-local `OnceCell<Box<dyn Clock>>`" but never commits); or (b) a `cfg(sim)` gate that re-routes `preempt_tick` through a host-only `env()` shim. Either is fine, but the RFC has to pick one and describe how callers (signal delivery, sleep_ms, future timeout-bearing syscalls) compile on the host. As written, the design assumes a host build of `kernel/` works, and it does not today.
- [B2] **IRQ-context invariants are silently violated by the host driver.** `preempt_tick` runs on real hardware in IRQ-tail context; a lot of kernel code paths around it call `IrqLock` or `crate::sync::*` primitives that on x86 use `cli`/`sti`. The simulator calls `preempt_tick` from an ordinary host thread with neither IRQ-disable semantics nor a faithful imitation of them. `IrqLock` on the host is presumably `spin::Mutex` (or similar) under cfg, so it doesn't deadlock — but invariants like "preempt_tick cannot be re-entered" or "this `IrqLock` is held strictly during `drain_expired`" are the *content* of what the simulator is supposed to verify under fault injection. The RFC must spell out: which kernel synchronization types swap out under host build (today already in tree? a list?), what the host substitutes are (and whether they're unfair / fair / contentious enough to surface real ordering bugs), and what the simulator does about `pre_tick` re-entrancy if `MockTimerIrq::inject_timer` is called inside a `drain_expired` callback (the FaultPlan can do this).
- [B3] **The Event enum's lifecycle variants (`Fork`, `Exec`, `Exit`, `Wait`, `SignalDelivered`) require *emit points* in the kernel that don't exist.** The RFC mentions introspection helpers but treats events as if they fall out for free. They don't — every `Fork` event needs a tracing call inserted at the fork syscall handler, every `Exit` at the exit path, etc. Those calls are real production code (gated by `cfg(feature = "sched-mock")` they may be, but they're still in the syscall hot path). The RFC needs to commit to either: (a) emit points are `tracing`-style macros that compile out without `sched-mock` (zero-cost when off, simple to verify), or (b) emit points are post-hoc reconstructed by the simulator from observable state (e.g. comparing `current_id` before/after each tick to derive `Switch`). Today's design conflates the two — the trace lists `SignalDelivered` as if it's free, but signal delivery doesn't currently call into anything observable.

**Advisory findings:**
- [A1] Linear tick-by-tick is correct for v1, but the RFC should call out that softirq processing (RFC 0005 §"What is *not* covered by this seam") happens between `ack_timer` and `rotate_or_resched` and is *not* observed by the simulator's between-tick hook. Phase 2's reference state machine needs to know that, or it'll predict transitions softirqs cause and the live trace won't show them.
- [A2] The `task::scheduler::ready_count`-shaped helpers should be on a single trait/struct (`SimIntrospect`) gated under `sched-mock`, not free functions sprinkled across `task/scheduler.rs`. One nm-grep target is easier to maintain than N.
- [A3] `MockClock`'s deadline `BTreeMap` matches production drain-order semantics; good. But under wakeup-reorder fault injection (§Failure-injection scope), the *production* `time::WAKEUPS` does not reorder. The fault is therefore strictly more nondeterminism than production exhibits — i.e., the simulator can produce traces production cannot. That's fine for finding bugs ("this code is wrong if drain order ever shuffles"), but the RFC should explicitly note that wakeup-reorder is a stress fault, not a faithful production reproduction, so reviewers don't dismiss reorder-only failures as "can't happen on real hardware."

dburkart:

**Summary:** This is a kernel-internal test harness — there is no userspace ABI in scope, so the usual POSIX-compliance / errno-shape critiques don't apply. What does apply is the *developer-experience* contract: the simulator's value is entirely in how easy it is for a kernel engineer (or auto-engineer agent) to go from "CI failed at HEAD~3" to "I can reproduce it locally in under a minute." The RFC's reproducibility envelope is well-thought-out, but the developer-facing UX has gaps.

**Blocking findings:**
- [B1] **Failing-seed-to-repro path is not specified end-to-end.** The RFC mentions `cargo run -p simulator --bin replay -- --seed=…` but never describes: (a) how a CI failure surfaces the seed (random output line? structured error? GitHub Actions annotation?), (b) where the JSON trace dump is written when the per-PR fast suite fails (artifact upload? checked into the failing-test directory? printed inline?), (c) how the developer correlates a property-test failure (proptest's own shrinking output) with the simulator's seed. Without this end-to-end spec, "turn flakes into seeds" risks shipping a harness where the seed is *technically* present in the logs but practically impossible to find. Treat this as the analog of "what does errno mean to the user" — the seed is the API, and the API needs to be documented before merge.

**Advisory findings:**
- [A1] The Event enum is a *public* surface for external consumers (#391, #390, future tooling). It should live in `simulator::trace` with explicit `#[non_exhaustive]` annotations so adding a variant isn't a breaking change. The RFC uses `enum Event { … }` without `#[non_exhaustive]` — easy to fix.
- [A2] Replay UX should support `--seed=…` *and* `--trace-json=…` (replay from a saved trace). The RFC implies this but doesn't commit. The "checked-in seed + JSON trace" gate on Roadmap item 6 (#501 repro) requires both file forms.
- [A3] Consider an `--explain` flag on the replay binary that pretty-prints the trace as a human-readable timeline. JSON is good for tooling; a tabular `Tick 42: Fork(parent=1, child=2) Switch(1→2)` view is good for humans, and humans are who reads bug reports.
- [A4] No mention of how the simulator interacts with `RUST_BACKTRACE=1` / panic hooks. If a kernel `panic!` fires under simulation, the seed must be in the panic message — otherwise `cargo test`'s default capture hides it. Roadmap item 2 should pin a panic hook that prepends the seed.

dburkart:
### Review: Academic

**Summary:** The RFC's prior-art survey is thorough and the design choices are well-grounded in the literature (sled/Polar Signals state-machine pattern, shuttle's randomized-iteration heuristic, VOPR's seed+commit envelope). The intellectual gap is in the *formal* properties of the system the simulator claims to test — specifically, the relationship between the simulator's RNG-driven nondeterminism and the kernel's actual nondeterminism, and the soundness/completeness story for the reference state machine. These are not idle theoretical concerns: getting them wrong produces a harness that confidently passes while missing real bugs (false negatives) or fails on bugs that don't exist on real hardware (false positives that erode trust in the harness).

**Blocking findings:**
- [B1] **RNG stream coupling between FaultPlan and the proptest reference model is unspecified, and the choice has soundness implications.** The RFC says "named streams (`fault_rng`, `wakeup_order_rng`, `task_arrival_rng`) all derived from the master seed via `splitmix64`." Good. But `proptest-state-machine` *also* has its own RNG (proptest's `TestRng`), and the RFC's `SchedulerStateMachine: ReferenceStateMachine` is driven by it. If the proptest RNG and the simulator's master seed are independent, then a property failure under seed S1 and proptest-seed P1 is reproducible only by recording *both*. If they're coupled (master seed seeds proptest), then proptest's shrinking — which mutates the transition sequence — can't shrink without re-running the whole simulator under a different fault plan, defeating the shrinker. There's a real design choice here (Hypothesis-style "draw bytes from a deterministic stream" vs. proptest's "tree of strategies") and the RFC needs to pick one. Cite: Hypothesis's *internal-test-case representation* paper (MacIver, 2019) is the canonical treatment of why this matters.
- [B2] **The reference state machine's correctness criteria are stated as "shape, not exact equivalence" but the RFC never defines what "shape" means formally.** This is the load-bearing soundness claim — if the reference is too loose, it accepts bug-y behavior; if it's too tight, every timing-permitted variation is a false positive. The RFC's two starter properties ("no task is in `TaskReady` and `TaskBlocked` simultaneously"; "every `Wait` is preceded by an `Exit` of the same child id") are *invariants over the trace*, not state-machine equivalences — they're closer in spirit to TLA+ safety properties. That's fine, but then the reference state machine machinery from `proptest-state-machine` is over-engineered for the job; what's actually wanted is a trace-property checker (à la stateright's `Property::always`). The RFC needs to either (a) commit to invariant-based checking and drop the reference-machine framing, or (b) commit to refinement-based checking (every kernel transition must be a transition the reference would have made) and accept the false-positive cost. Today it claims (a) while implementing (b)'s machinery.

**Advisory findings:**
- [A1] The "linear tick-by-tick vs. event-driven" defense is correct on the determinism axis (event-driven introduces a model-the-model problem) but incomplete on the *liveness* axis. A linear tick driver that runs for `max_ticks` ticks and observes "no `init: hello from pid 1` printed" cannot distinguish "init is starved by scheduling" from "init is genuinely deadlocked." The RFC should cite Lamport's safety-vs-liveness distinction and commit to a liveness criterion (e.g., "every blocked task is either dequeued within K ticks or the simulator flags `LivenessViolation`"). This is exactly the discipline TigerBeetle's VOPR adopts via "no-progress timer" assertions.
- [A2] `BTreeMap` iteration order is deterministic across libstd versions on the same host arch (sorted-key invariant), so the toolchain pin is sufficient for that data structure. *However*, `HashMap` iteration order is randomized via `RandomState::default()` which queries the OS for entropy — if any code path reachable under `sched-mock` builds uses `HashMap`, the toolchain pin doesn't save you. The RFC should commit to a `clippy::disallowed_types` lint forbidding `HashMap` in the simulator + sched-mock-gated kernel code, with a deterministic-hasher escape hatch (`HashMap<K, V, FxHasher>`) for performance-sensitive paths. This is the kind of footgun whose first occurrence is a multi-day debugging session.
- [A3] The Hyperkernel paper (Nelson et al., SOSP 2017), cited in RFC 0005, is also relevant here: their interface decomposition gave them a kernel they could *symbolically execute*. The randomized harness this RFC proposes is strictly weaker than symbolic execution, but the trait surface is the same shape, so a future "Phase 3: symbolic execution of the scheduler" RFC inherits this work. Worth a sentence noting that Phase 2 doesn't preclude Phase 3.

dburkart:
### Review: Testing-Infrastructure Engineer

**Summary:** Good prior-art selection and the architectural choices (single binary, reuse mocks, `(Tick, Event)` trace) match what works elsewhere. Two structural problems though: the perf budget arithmetic doesn't add up, and the RFC has no story for the most important developer interaction with a DST harness — *minimizing* a failing seed into a small repro. A 100k-tick failing trace is unreadable; without minimization the harness ships a generator and a black box, not a debugger.

**Blocking findings:**
- [B1] **CI perf budget is off by ~2× and the math is undefended.** §Performance says "1 µs per tick" (measured against `MockClock` in RFC-0005 unit tests — fair) and §Implementation Roadmap commits to "1000 random seeds × 100k ticks each in under 60s wall-clock for the per-PR fast suite." 1000 × 100,000 × 1 µs = 100s, not 60s, and that's just MockClock — actually running `kernel::task::preempt_tick` per tick is more expensive. The 1 µs benchmark was a bare-mock op, not a full preempt-tick path; realistic per-tick cost will be 5-20 µs once `BTreeMap::split_off`, ready-bank rotation, and the trace `push` are included. Either drop the per-PR target to 100 seeds × 10k ticks (=1s wall-clock at 10 µs/tick), or commit to a parallel-test-runner story (rayon over seeds, 16 cores → 6s for the full 1000×100k). The current numbers will be missed in the first PR that lands the stress suite and the team will lose trust in the budget.
- [B2] **No seed-minimization / trace-shrinking story exists outside of proptest's own machinery.** The simulator's value proposition is "turn a flake into a single failing seed" — but a single failing seed at 100k ticks is still a 100k-line trace. proptest-state-machine shrinks *transitions* (its own generated input), not the simulator's wall of random fault decisions. Without a shrinker that can binary-search "which tick first exhibited the violation" and "what's the minimal FaultPlan that still reproduces it," debugging a real failing seed is going to be ~as painful as today's `rerun-and-hope`. Hypothesis (and proptest, partially) shrink via "delete bytes from the random stream and re-run"; that pattern is directly applicable here — the simulator's RNG draws are the input — but the RFC must commit to it. Concretely: every failing seed should be auto-minimized before being checked in, and `cargo run -p simulator --bin replay -- --seed=… --minimize` should be a supported subcommand. Without this, Roadmap item 6 ("Reproduce #501 as a single failing seed") ships a 100k-line trace as the "single seed" — technically meeting the goal, practically not.

**Advisory findings:**
- [A1] CI artifacting: failing-seed JSON traces should be uploaded as GitHub Actions artifacts with retention ≥ 30 days, named with `<seed>-<commit-sha>.json`. The RFC §Performance mentions trace caps but not artifact mechanics.
- [A2] Per-PR vs. nightly suite differentiation is good, but the RFC doesn't say *which* nightly schedule. Daily? Weekly? The right cadence depends on how often the kernel's scheduler-affecting code changes; suggest a workflow that triggers nightly *only when* `kernel/src/task/**` or `kernel/src/sync/**` changed since the last successful nightly, otherwise skip.
- [A3] Roadmap item 7 ("proptest-state-machine reference model") and item 6 ("Reproduce #501") should swap order — having the property machinery in place lets #501 be encoded as a property failure (more rigorous), not just a seed-with-a-recorded-violation. Reviewer-discretion advisory.
- [A4] Coverage / saturation feedback: how does anyone know when 1000 seeds is "enough"? Tools like AFL track basic-block coverage and grow inputs that hit new edges. v1 doesn't need that, but the RFC should call out that *coverage-guided* exploration is a v2 surface (and a Phase 2.1 RFC candidate), so the v1 design doesn't accidentally preclude it. The current `(Tick, Event)` trace already lays the groundwork — events are roughly basic-block-equivalent waypoints.
- [A5] The "iteration budget" ≈ 100s of seeds heuristic from shuttle is correct as a *first-bug-found* number, but for *no-regression* testing (the per-PR use case) the bar is higher — you want to detect a regression that re-introduces a flake at the same rate as before. That implies higher iteration counts in the steady state. The RFC should distinguish "exploring for new bugs" (proptest, nightly) from "regression detection on known seeds" (per-PR, deterministic seed list checked into `tests/seeds/*.txt`). The latter scales linearly with the seed list and is bounded.

dburkart:
## Defense cycle 1 — addressing blocking findings

All 10 `[Bx]` findings from the panel are addressed in commit `0e8296b`. Summary by reviewer:

### Security Researcher
- **[B1] nm-check regex commitment** — Addressed in new §"nm-check guard committed surface". The RFC now commits to a concrete demangled-symbol regex: `^(.*::)?(Mock(Clock|TimerIrq|ClockState|IrqState)|sim_introspect_\w+|sched_mock_trace::push|SchedMockTrace\w*)$`. CI demangles via `nm --demangle=rust` so both v0 and legacy mangling forms are caught. New code that introduces a new prefix updates the regex in the same PR; the regex is cited at the top of `kernel/src/task/env.rs` and `kernel/src/sched_mock_trace.rs` so violations of the convention are visible at review time.
- **[B2] `--allow-dirty` trace attribution** — Addressed in new §"Trace attribution under dirty trees". Trace JSON gains an explicit `"tree_dirty": bool` field; CI never produces dirty traces (clean checkout); local runs may set `--allow-dirty` but the resulting trace is loud about it and the replay binary refuses to verify a dirty trace against its recorded commit. Bug reports that cite a dirty trace are unambiguously flagged.
- Advisories A1 (resolved feature set in trace) and A2 (seed in panic + SIGINT output) absorbed into §"Failing-seed-to-repro path".

### OS Engineer
- **[B1] Host-target buildability of `kernel/`** — Addressed in new §"Host-target buildability of `kernel/`". Picks option (a): a parallel `#[cfg(all(not(target_os = "none"), feature = "sched-mock"))] fn env()` in `kernel/src/task/env.rs` reading from a `thread_local! SIM_ENV: OnceCell<...>` that the simulator installs via `install_sim_env()`. The audit and gating work is now Roadmap item 1 — landed *before* any simulator code exists. New CI job: `cargo build -p kernel --target x86_64-unknown-linux-gnu --features sched-mock` runs alongside the bare-metal build.
- **[B2] IRQ-context invariants** — Addressed in new §"IRQ-context invariants under host driver". Two invariants: (1) `preempt_tick` non-re-entrancy enforced by a `cfg(debug_assertions)` thread-local guard inside the kernel; (2) host substitutes for `IrqLock` etc. preserve ordering, not microarchitecture. Documented in `docs/design/simulator.md` as a host-substitute table. Bugs whose trigger is below the seam (microarchitectural reordering, IRETQ stack-faulting) are explicitly out of scope and route to Phase 2.1 RFCs.
- **[B3] Event emit points** — Addressed in new §"Event emit points". Splits the Event variants into observed-by-snapshot (free, derived from before/after introspection) and emit-point-required (`Fork`, `Exec`, `Exit`, `Wait`, `SignalDelivered` — a single `sched_mock_trace!` macro at each syscall handler that compiles to `()` when `feature = "sched-mock"` is off). Verification by assembly inspection in CI. Roadmap item 5 lands the macro infrastructure.
- Advisories A1 (softirq), A2 (single SimIntrospect target), A3 (wakeup-reorder is a stress fault, label seeds accordingly) absorbed in §"Failure-injection scope" and §"Reference state machine".

### User Space Staff Engineer
- **[B1] Failing-seed-to-repro path** — Addressed in new §"Failing-seed-to-repro path". Six-step end-to-end spec: `VIBIX_SIM_SEED=…`-prefixed panic hook (load-bearing string for tooling); GitHub Actions annotations from CI; trace JSON artifact upload with `sim-trace-${seed}-${commit}.json` naming and 30-day retention; `cargo run -p simulator --bin replay -- --seed=…` and `--trace-json=…` and `--explain` all committed; proptest seed master-seeds the simulator so a single seed reproduces both the transition sequence and the fault plan; `Event` is `#[non_exhaustive]`; SIGINT handler captures the seed.
- Advisories A1 (`#[non_exhaustive] Event`), A2 (`--trace-json`), A3 (`--explain`), A4 (panic + SIGINT seed) all absorbed.

### Academic
- **[B1] RNG stream coupling** — Addressed in new §"RNG stream coupling: `proptest` master-seeds the simulator". One-way: `proptest`'s `TestRng` master-seeds the simulator's master via `splitmix64`. proptest shrinks the transition sequence; the simulator's §Seed minimization Stage 2 (FaultPlan delta-debugging) runs *after* proptest minimization to reduce the fault plan separately. This is the Hypothesis/MacIver discipline split into the two natural input layers we have.
- **[B2] Reference state machine framing** — Addressed in new §"Reference state machine: invariants over the trace, not refinement". Phase 2 v1 commits to invariant-based property checking (TLA+-style safety predicates over trace prefixes), not refinement comparison. The `SchedulerStateMachine` from `proptest-state-machine` only generates *transitions*; it's not consulted as an oracle. Liveness invariants are expressed as no-progress timers (Lamport's distinction made operational; mirrors VOPR). A future Phase 3 RFC may add refinement or symbolic execution; the seam doesn't preclude it.
- Advisory A1 (liveness invariants) absorbed in §"Reference state machine"; A2 (`HashMap` ordering footgun → `clippy::disallowed_types` lint) absorbed in §"RNG stream coupling" and Roadmap item 1; A3 (Hyperkernel/symbolic-execution successor) absorbed in §"Reference state machine".

### Testing-Infrastructure Engineer
- **[B1] CI perf budget arithmetic** — Addressed in new §"CI perf budget". Honest table: per-PR fast suite is **100 seeds × 10k ticks ≈ 10–20s** (not 1000 × 100k); plus a bounded regression seed list (~50 × 10k ≈ 5–10s); plus a nightly suite at 10k × 100k via rayon-over-seeds (~10–20 min on 16 cores). The exploration-vs-regression-detection distinction (Testing-A5) is now explicit. Roadmap item 9 reflects these numbers.
- **[B2] Seed minimization / trace shrinking** — Addressed in new §"Seed minimization and trace shrinking". Two-stage minimizer: Stage 1 binary-searches the smallest tick window that still violates; Stage 2 delta-debugs the FaultPlan to identify the minimal subset of fault classes the failure needs. `--minimize` is a CLI subcommand; auto-minimization is **mandatory before checking in any repro seed** (Roadmap item 6.5, gating item 7). Without this, a 100k-tick failing trace is unreadable; with it, the harness ships a debugger, not a black box.
- Advisories A1 (artifact upload + retention), A2 (nightly cadence: only when `kernel/src/{task,sync}/**` changed), A3 (suggested item-7-before-item-6 swap is reflected by the new ordering: minimizer item 6.5 lands before item 7 #501 repro), A4 (coverage-guided exploration as v2 surface — noted in §"Failure-injection scope" deferral list), A5 (regression-vs-exploration distinction) all absorbed.

### Diff summary
- 1 file changed, +601/-67 lines.
- Seven new design subsections, all addressing blockers.
- Implementation Roadmap reorganized from 11 items to 12 items (added: host-target buildability as item 1; seed minimizer as item 6.5; reorganized property-test integration as item 8 after the minimizer so #501 repro can use proptest's machinery if desired).

Re-spawning all five archetypes for re-review on the updated RFC.

dburkart:
### Review: Security Researcher (re-review, defense cycle 1)

**Summary:** Both blocking findings are addressed with concrete commitments in the updated RFC. The nm-check regex is now spelled out verbatim (not handwaved as "extend the regex") and the demangling story (`nm --demangle=rust`) is explicit. The `tree_dirty` field in trace JSON cleanly resolves the attribution ambiguity; the "replay refuses to verify a dirty trace against its commit" rule means a dirty-tree report cannot be silently mistaken for a clean repro. Good.

**Blocking findings:**
- (none)

**Advisory findings:**
- [A1] The `--allow-dirty` story is sound on the trace-side, but the `VIBIX_SIM_SEED=` panic-hook string is going to be greppable forever — that's a *commitment* to a stable interface, not just a debug convenience. The RFC should call it out as a stable string in §Reproducibility envelope's "out of scope" / "in scope" list explicitly: it's part of the API surface, not an implementation detail. Minor wording fix, not blocking.

**Verdict:** LGTM

dburkart:
### Review: OS Engineer (re-review, defense cycle 1)

**Summary:** All three blocking findings are addressed substantively. The host-target buildability section commits to option (a) cleanly, with the thread-local `SIM_ENV` + `install_sim_env` shape spelled out in code and Roadmap item 1 doing the audit work *first*, before any simulator code lands — that ordering is exactly right. The IRQ-context section honestly draws the line between what the simulator catches (seam-visible ordering bugs under fault injection) and what it doesn't (microarchitectural ring-0 races). The Event emit-point split into observed-by-snapshot vs. emit-point-required is a clean factoring; the `sched_mock_trace!` macro that compiles to `()` in production with assembly-inspection verification is the right level of paranoia.

**Blocking findings:**
- (none)

**Advisory findings:**
- [A1] The `cfg(debug_assertions)` re-entrancy guard inside `preempt_tick` is good, but the RFC should clarify whether it's also active in `--release` test builds (i.e., `cfg(any(debug_assertions, all(test, feature = "sched-mock")))`) — otherwise a release-mode simulator stress run loses the invariant check. Probably want it always-on under `sched-mock`. Trivial fix.

**Verdict:** LGTM

dburkart:
### Review: User Space Staff Engineer (re-review, defense cycle 1)

**Summary:** The end-to-end seed-to-repro path is now spelled out in six concrete steps, with the `VIBIX_SIM_SEED=`-prefixed panic hook as a load-bearing string interface, structured GitHub Actions annotations, artifact upload retention, and three CLI shapes (`--seed`, `--trace-json`, `--explain`) all committed. The proptest-seed-master-seeds-simulator integration cleanly resolves the property-test failure attribution question. `#[non_exhaustive] Event` and the SIGINT handler each do exactly one job and do it right.

**Blocking findings:**
- (none)

**Advisory findings:**
- [A1] The `--explain` tabular output format isn't pinned (just sketched as `Tick 42: Fork(parent=1, child=2) Switch(1→2)`). When auto-engineer agents start parsing this, they'll need a stable column layout. Consider a `--explain=table|json|tsv` flag with `tsv` as the agent-stable form. Defer to implementation.

**Verdict:** LGTM

dburkart:
### Review: Academic (re-review, defense cycle 1)

**Summary:** Both blocking findings are addressed at the level of formal commitment I asked for. The "invariants over the trace, not refinement" framing is the right call for v1 and the RFC now explicitly demotes `SchedulerStateMachine` from oracle to transition generator — a clean factoring that preserves the soundness story. The proptest-master-seeds-simulator + Stage-2-FaultPlan-minimizer-runs-after-shrinking decomposition is the principled way to handle the two-input-layer problem (matches the spirit of Hypothesis's MacIver paper without claiming to reimplement byte-stream representation). Liveness as no-progress timer is the operationalization Lamport would recognize. The Hyperkernel forward-reference (Phase 3 doesn't preclude symbolic execution) is the right epistemic humility.

**Blocking findings:**
- (none)

**Advisory findings:**
- [A1] The `clippy::disallowed_types` lint forbidding `HashMap` is good, but `BTreeMap`'s deterministic-iteration property only holds *if its key type's `Ord` impl is itself deterministic*. For our `TaskId = usize`, fine. For future `Tick`-keyed or richer-keyed maps, worth a one-line note that custom `Ord` impls in `simulator/` and `sched-mock`-gated kernel code must be reviewed for determinism. Minor.

**Verdict:** LGTM

dburkart:
### Review: Testing-Infrastructure Engineer (re-review, defense cycle 1)

**Summary:** The honest perf-budget table is exactly the kind of thing I wanted — separate columns for per-PR fast, regression seed list, and nightly, with realistic per-tick microsecond costs and a parallelization story (rayon-over-seeds) that actually lands the nightly target in a CI-tractable budget. The two-stage minimizer (tick-window binary search + FaultPlan delta-debugging) lifts the harness from "generator + black box" to "generator + debugger" in the way Hypothesis-style shrinking does for normal property tests. The mandatory-minimization-before-check-in gate on Roadmap item 7 is the right enforcement point. The regression-vs-exploration split (bounded `tests/seeds/*.txt` for known regressions; randomized exploration for new bugs) is the operating model that scales over time.

**Blocking findings:**
- (none)

**Advisory findings:**
- [A1] The "nightly only when `kernel/src/{task,sync}/**` changed" cadence is a sensible default but should also fire on changes to `simulator/**` itself — a regression in the simulator (e.g., an event-emission gap) is exactly as serious as a kernel regression and just as silent. Trivial extension to the path filter; not blocking.

**Verdict:** LGTM

PR #732, opened 2026-04-29
RFC 0006: Demand-Paged File mmap and Page Cache for ext2

dburkart:
## Summary

RFC 0006 proposes a per-inode page cache, a `FileOps::mmap` hook, and a
`FileObject` `VmObject` that together let userspace `MAP_PRIVATE`/`MAP_SHARED`
ext2-backed regular files with demand-paged servicing. Page-cache and the
existing block buffer cache are kept as **separate** caches with explicit
readpage/writepage handoff; the four normative buffer-cache invariants from
RFC 0004 Workstream C are preserved unchanged. The same machinery converts
the eager ELF loader to demand-paged, retiring the `PT_INTERP → ENOEXEC` gate.

will be reviewed by a six-archetype panel (the four defaults — security
researcher, OS engineer, user-space staff engineer, academic — plus
`filesystem-engineer` for writeback ordering and crash consistency, and
`performance-engineer` for the page-fault hot path and cache contention).
Up to 3 defense cycles. Implementation issues are filed from the roadmap
section after merge as a single epic with workstream sub-issues.

## Scoping

- Defense-cycle cap: **3** (cross-cutting redesign touching mem, fs, block, sched).
- Extra archetypes: **filesystem-engineer**, **performance-engineer**.

## RFC sections

- Abstract
- Motivation
- Background (vibix today + prior art: Linux page cache, FreeBSD vnode_pager,
  Mach VM external pagers, SerenityOS InodeVMObject, Pillai et al. 2014,
  Bonwick & Adams 2001)
- Design (`PageCache`, `CachePage`, `AddressSpaceOps`, `FileObject`,
  `FileOps::mmap`, fault algorithm, MAP_PRIVATE CoW reuse, MAP_SHARED writeback,
  demand-paged execve, eviction, truncate)
- Security Considerations
- Performance Considerations
- Alternatives Considered (unified cache; buffer-cache-as-backing-store;
  separate-with-handoff ← chosen)
- Open Questions
- Implementation Roadmap (5 workstreams, 3 waves, parallelisable)

## Test plan

- [ ] All blocking findings are addressed within the 3-cycle cap.

### DIFF
--- a/docs/RFC/0006-page-cache-file-mmap.md
+++ b/docs/RFC/0006-page-cache-file-mmap.md
@@ -1,7 +1,7 @@
 ---
 rfc: 0006
 title: Demand-Paged File mmap and Page Cache for ext2
-status: In Review
+status: Accepted
 created: 2026-04-28
 ---
 
@@ -1265,35 +1265,34 @@ clarity; no normative change to the design.
 
 ## Open Questions
 
-- [ ] **rootfs block size.** RFC 0004 left this open
-  ("`-b 4096` matches host defaults; `1 KiB` makes indirect-block
-  math smaller"). Page cache shifts the calculus toward 4 KiB:
-  a 4 KiB block size means one `bread` per `readpage`, vs four for
-  1 KiB. Recommend flipping to 4 KiB in the same epic that lands the
-  page cache. Decided: 4 KiB.
-- [ ] **In-place buffer hand-off vs always-memcpy.** When `block_size
-  == 4096` and the buffer is page-aligned, `readpage` could reuse the
-  `BufferHead.data` slab as the cache page directly (zero-copy).
-  This requires teaching the buffer cache to release frame ownership
-  on demand and the page cache to accept ownership; coupling that
-  the explicit-handoff layering was specifically chosen to avoid.
-  **Defer** to a follow-up perf RFC; first ship the always-memcpy
-  path. Memcpy of one 4 KiB page is ~1 µs.
-- [ ] **`msync(2)` syscall.** Out of this epic; deferred. The
-  writeback daemon + `fsync` is sufficient for the static + dynamic
-  binary userspace this RFC enables. Track as a follow-up issue.
+All open questions are **deferred to implementation** — none block
+the design as accepted.
+
+- [ ] **rootfs block size.** Recommend flipping to 4 KiB in the same
+  epic that lands the page cache (decided: 4 KiB; track as a small
+  bring-up issue alongside Workstream C).
+- [ ] **In-place buffer hand-off vs always-memcpy.** Ship the
+  always-memcpy path first; revisit zero-copy as a follow-up perf
+  RFC if the ~1 µs/page memcpy shows up in profiles.
+- [ ] **`msync(2)` syscall.** Deferred to a follow-up. Writeback
+  daemon + `fsync` covers durability; no current userspace caller.
 - [ ] **TLB shootdown on truncate / page eviction.** Out of scope
   (uniproc). Resolved by SMP RFC.
-- [ ] **`MAP_HUGETLB`, `MAP_LOCKED`, `MAP_NORESERVE`.** All rejected
-  with `EINVAL` for the initial implementation. Track as follow-ups.
-- [ ] **Mincore, posix_fadvise.** Out of scope. The cache structure
-  trivially supports both; ship in the same epic if a userspace
-  caller materialises before merge.
-- [ ] **`MAP_SHARED` over `tmpfs` / `tarfs`.** ramfs/tarfs override
-  `FileOps::mmap` to return an `AnonObject`-style backing — no page
-  cache, no writeback. Document this in the FS impls; `MAP_SHARED`
-  on a tarfs file is read-only by mount semantics anyway, so no
-  writeback is sensible.
+- [ ] **`MAP_HUGETLB`, `MAP_LOCKED`, `MAP_NORESERVE`.** Rejected
+  with `EINVAL` for the initial implementation. Each lands as a
+  small follow-up RFC if a userspace caller appears.
+- [ ] **`mincore`, `posix_fadvise`.** Cache trivially supports both;
+  ship if a userspace caller materialises during the implementation
+  epic.
+- [ ] **`MAP_SHARED` over `tarfs`/`ramfs`.** Override `FileOps::mmap`
+  to return `AnonObject`-style backing; document in the FS impls.
+  No writeback; tarfs is read-only by mount semantics.
+- [ ] **`ra_state` placement (per-inode vs per-OpenFile).** Pick
+  per-`PageCache` for MVP; revisit when concurrent-readers-of-same-
+  file becomes a measured workload (Performance cycle-2 A1).
+- [ ] **`dirty_ratio` writer-throttling.** Linux-style throttle
+  knob to bound `cache.dirty` growth under sustained pressure.
+  Defer to a follow-up perf RFC (Performance cycle-1 A3).
 
 ### Resolved during peer review
 
--- a/docs/RFC/0006-page-cache-file-mmap.md
+++ b/docs/RFC/0006-page-cache-file-mmap.md
@@ -1,7 +1,7 @@
 ---
 rfc: 0006
 title: Demand-Paged File mmap and Page Cache for ext2
-status: Draft
+status: In Review
 created: 2026-04-28
 ---
 
--- a/docs/RFC/0006-page-cache-file-mmap.md
+++ b/docs/RFC/0006-page-cache-file-mmap.md
@@ -1,0 +1,943 @@
+---
+rfc: 0006
+title: Demand-Paged File mmap and Page Cache for ext2
+status: Draft
+created: 2026-04-28
+---
+
+# RFC 0006: Demand-Paged File mmap and Page Cache for ext2
+
+## Abstract
+
+Introduce a unified, page-granular **page cache** keyed by
+`(InodeId, page_index)`, hang it off `Inode`, and add a `FileOps::mmap`
+hook so userspace `mmap(2)` of ext2-backed regular files completes
+instead of returning `ENODEV`. File-backed VMAs are serviced lazily by a
+new `FileObject : VmObject` whose `fault` consults the page cache and
+synthesises reads through the existing block buffer cache. The same
+machinery turns the existing eager ELF loader into a demand-paged
+loader, retiring the `PT_INTERP → ENOEXEC` execve gate and unblocking
+the dynamic-linker work tracked separately.
+
+## Motivation
+
+RFC 0004 (ext2, merged via epic #537) deferred three coupled features:
+the page cache, file-backed mmap, and `PT_INTERP`. As of today
+`sys_mmap` rejects every `fd != -1` with `ENODEV`
+(`kernel/src/arch/x86_64/syscall.rs:1366`) and `execve` of any ELF with
+a `PT_INTERP` returns `ENOEXEC` (RFC 0004 Workstream F). That blocks:
+
+- Any userspace program that `mmap(MAP_SHARED)`s a configuration file,
+  database, or shared memory region — POSIX-conformance tooling
+  (pjdfstest itself, fsstress) trips on this immediately.
+- Any non-static binary. musl's dynamic loader, glibc's `ld-linux.so`,
+  busybox's dynamic build, and every distro ELF reach `PT_INTERP`.
+- Memory-pressure response. The buffer cache today caches only block
+  reads issued by `FileOps::read`; the kernel cannot reclaim file data
+  for read-mostly workloads without losing it on the next read.
+- A unified writeback story. `MAP_SHARED` writes need to land at the
+  same checkpoint as `write(2)`-issued dirty buffers — without a page
+  cache there is no place to track them.
+
+The right shape for this is one design document, not three: every
+question of cache layering, locking, and writeback ordering recurs
+identically across mmap, read, and execve. RFC 0006 fixes the
+architecture and lets the three downstream epics ship as execution-only
+work.
+
+## Background
+
+### What vibix has today (2026-04-28)
+
+- **Block buffer cache** (`kernel/src/block/cache.rs`, RFC 0004
+  Workstream C). CLOCK-Pro replacement, `(DeviceId, u64)` key, four
+  normative invariants:
+  1. eviction never touches a buffer with `Arc::strong_count > 1` (any
+     external caller pin);
+  2. eviction never touches `DIRTY | LOCKED_IO`;
+  3. `bread` returns `ENOMEM` rather than synchronously flushing dirty
+     buffers during eviction;
+  4. single-cache-entry: only one `BufferHead` ever exists per
+     `(dev, blk)`.
+- **Per-mount writeback daemon** (`kernel/src/block/writeback.rs`).
+  Default 30 s cadence, joined by `SuperOps::unmount`, runs under an
+  `SbActiveGuard` that pins the superblock during the flush and is
+  released across the sleep.
+- **`VmObject` trait** (`kernel/src/mem/vmobject.rs`). The polymorphism
+  point for "what backs this VMA?". `AnonObject` is the only
+  implementation today; the trait already exposes `fault(offset,
+  access)`, `clone_private`, `frame_at`, `truncate_from_page`, and
+  `evict_range`.
+- **VMA tree** (`kernel/src/mem/vmatree.rs`). `Vma { start, end,
+  prot_user, prot_pte, share, object, object_offset, vma_flags }`,
+  `Share::{Private, Shared}`, no separate `VmaKind::Cow` enum — CoW is
+  resolved at the PTE layer (W bit stripped after fork; the page-fault
+  resolver invokes `cow_copy_and_remap`).
+- **Page-fault dispatch** (`kernel/src/mem/pf.rs` + `arch/x86_64/idt`).
+  Pure-logic gates run first; `VmObject::fault` is invoked from the
+  IRQ-disabled handler today, which is fine for `AnonObject` but is the
+  central concern of this RFC.
+- **Eager ELF loader** (`kernel/src/mem/loader.rs`,
+  `kernel/src/mem/elf.rs`). Walks `PT_LOAD`, allocates frames, copies
+  through the HHDM, installs final-protection PTEs. No `PT_INTERP`
+  support; `INTERP_LOAD_BASE = 0x4000_0000` is reserved for a future
+  loader.
+- **VFS `FileOps`** (`kernel/src/fs/vfs/ops.rs`). Today: `read`,
+  `write`, `seek`, `getdents`, `ioctl`, `flush`, `fsync`, `open`,
+  `release`. No `mmap` hook.
+
+### Prior art surveyed
+
+- **Linux page cache** (`mm/filemap.c`, `mm/page-writeback.c`, "address
+  space operations"). One `struct page` per 4 KiB physical frame, hung
+  off an `address_space` rooted at `struct inode`. `i_mapping` is the
+  XArray-keyed by file `pgoff`. `readahead`, `readpage`/`readpages`,
+  `writepage`, `set_page_dirty`, `release_folio`, `migrate_folio`. The
+  buffer head layer (`fs/buffer.c`) maps page-sized cache entries onto
+  block-sized device entries through the `buffer_head` chain — this is
+  the "buffer cache as backing store" model in concrete form.
+- **FreeBSD VFS / VM** — `vnode_pager`, the unified page/buffer cache
+  driven by Yokota & Dyson 1995. `b_bufobj` and `vm_object` share
+  pages directly; the buffer cache is a window onto the VM page cache,
+  not a separate pool. Anchor for the "unified" alternative.
+- **Mach VM** — Rashid et al. 1988, "Machine-Independent Virtual Memory
+  Management for Paged Uniprocessor and Multiprocessor Architectures".
+  External pagers, demand-pageable memory objects, the
+  `vm_object → memory_object` split that Linux's `address_space`
+  inherited. Anchor for the formal model: a memory object is a
+  contract over `(offset → page)` plus a writeback channel, and the
+  pager need not live in the kernel.
+- **SerenityOS** — `Kernel/Memory/InodeVMObject.cpp`,
+  `Kernel/Memory/SharedInodeVMObject.cpp`,
+  `Kernel/Memory/PrivateInodeVMObject.cpp`. Closer in spirit to vibix:
+  one `InodeVMObject` per inode, pages owned by the inode rather than
+  by a global cache. Demonstrates that the unified-cache model is
+  optional — a per-inode object reaches feature parity for read,
+  mmap, and demand-paged execve without a global radix tree.
+- **Redox** — uses a pager service in userspace; not directly applicable
+  to vibix's monolithic model but useful for the
+  external-pager-as-memory-object framing.
+- **Pillai et al. 2014, "All File Systems Are Not Created Equal"
+  (OSDI)** — the empirical work behind the writeback-ordering arguments
+  in RFC 0004 §rename. Same crash-consistency frame applies to
+  msync(2) / writepage ordering on `MAP_SHARED`.
+- **Bonwick & Adams 2001, "Magazines and Vmem"** — per-CPU caching
+  layer that anyone considering scaling the page cache hot path under
+  SMP eventually re-reads. Out of scope for this RFC (vibix is
+  uniprocessor); cited so the SMP follow-up has a starting point.
+- **Intel SDM Vol. 3A §4.10 "Caching Translation Information"**, §11
+  PAT/MTRR, §28 VMX page-walk paths. Relevant to `MAP_SHARED` writes
+  on dirty PTE bookkeeping and to the TLB-shootdown story when the
+  cache evicts a page that is mapped into multiple address spaces.
+
+### What is missing in vibix
+
+1. No structural place to put a page-cache entry. `Inode` has no
+   `mapping` field; `BlockCache` is keyed on `(DeviceId, blk)`, not on
+   `(InodeId, pgoff)`.
+2. No page-grain replacement story. The buffer cache is block-sized
+   (block size `<= 4 KiB` typical, `1 KiB` for the standard rootfs
+   image during early bring-up).
+3. No `FileOps::mmap`. Even if a cache existed, there is nowhere to
+   plug the inode-specific "produce a `VmObject` for this open file"
+   hook.
+4. No demand-fault path that can wait. `VmObject::fault` is called from
+   an IRQ-disabled context today. A file fault must:
+   (a) drop locks + re-enable interrupts before issuing block I/O,
+   (b) park on a wait-queue while the I/O completes,
+   (c) resume the fault under a fresh lock acquisition.
+   None of that exists for `AnonObject` because anonymous faults never
+   block on a device.
+5. No writeback for a page cache. The per-mount writeback daemon
+   walks the buffer cache; it does not know about page-cache entries.
+
+## Design
+
+### Overview
+
+Three layers, top to bottom:
+
+```
+┌─────────────────────────────────────────────────────────┐
+│  Page-fault handler (mem/pf.rs + arch/x86_64/idt)        │
+│   ↓ resolve()                                            │
+│  VmObject::fault(offset, access)                         │
+│    AnonObject (today)        FileObject (this RFC)       │
+│                                ↓ get_or_read_page()      │
+│  PageCache (per-Inode)                                   │
+│    ├── pages: BTreeMap<pgoff, Arc<CachePage>>           │
+│    ├── dirty: BTreeSet<pgoff>                           │
+│    └── ops: Arc<dyn AddressSpaceOps>  ← per-FS impl     │
+│                                ↓ readpage(buf, pgoff)    │
+│  AddressSpaceOps::readpage / writepage                   │
+│   ↓ bread / mark_dirty / sync_dirty_buffer               │
+│  BlockCache  (RFC 0004 — unchanged)                      │
+│   ↓ BlockDevice::read_at / write_at                      │
+│  virtio-blk / future drivers                             │
+└─────────────────────────────────────────────────────────┘
+```
+
+The page cache is **separate** from the buffer cache. They compose by
+the page cache invoking `bread` on miss to assemble a 4 KiB page from
+one or more block-sized buffers; the page cache then stops touching the
+buffer cache until eviction or writeback. RFC 0004's four normative
+buffer-cache invariants are unchanged and uncontested by this design
+(see §"Cache layering trade-off" for why this layering wins over
+unified or buffer-cache-as-backing-store).
+
+### Key data structures
+
+#### `CachePage`
+
+One page-sized entry. Lives behind `Arc<CachePage>` so VMAs and the
+cache index can each hold a strong reference without a parallel
+refcount.
+
+```rust
+pub struct CachePage {
+    /// Physical frame containing the cached file data. Refcounted via
+    /// the existing `mem::refcount` machinery — the cache holds one
+    /// reference; every PTE mapping it holds one more.
+    pub phys: u64,
+
+    /// Page index = file_offset / 4096.
+    pub pgoff: u64,
+
+    /// State bits — see `PG_*` below. AtomicU8 so reads in the fault
+    /// hot path don't take any lock.
+    pub state: AtomicU8,
+
+    /// Wait-queue for the IN_FLIGHT handshake. A second fault on the
+    /// same page parks here until the original reader publishes UPTODATE.
+    pub wait: WaitQueue,
+}
+
+// State bits.
+pub const PG_UPTODATE:   u8 = 1 << 0;  // contents reflect on-disk image
+pub const PG_DIRTY:      u8 = 1 << 1;  // mutated since last writeback
+pub const PG_IN_FLIGHT:  u8 = 1 << 2;  // I/O issued, not yet complete
+pub const PG_WRITEBACK:  u8 = 1 << 3;  // in writepage(), do not re-dirty
+pub const PG_LOCKED:     u8 = 1 << 4;  // exclusive serialization for fill
+```
+
+`PG_IN_FLIGHT` and `PG_LOCKED` are distinct on purpose: `PG_LOCKED`
+serialises the cache-fill handshake (only one task fills any given
+page); `PG_IN_FLIGHT` advertises that a `readpage` is presently
+issuing block I/O so other tasks know to park rather than spin.
+
+#### `PageCache`
+
+Per-inode container. Owned by `Inode` (a new `mapping` field gated by
+`#[cfg(feature = "page_cache")]` until the migration completes).
+
+```rust
+pub struct PageCache {
+    /// Sparse page index. BTreeMap is the same data structure used by
+    /// AnonObject today; we deliberately match its locking discipline.
+    inner: BlockingMutex<PageCacheInner>,
+
+    /// Per-FS hook for backing I/O. Set at inode construction.
+    ops: Arc<dyn AddressSpaceOps>,
+
+    /// File size at the moment of last truncate / read-of-i_size.
+    /// Faults past this offset return VmFault::OutOfRange (SIGBUS).
+    /// Updated under `inner.lock()` together with the cache index.
+    i_size: AtomicU64,
+
+    /// Back-pointer to the InodeId for writeback enqueue.
+    inode_id: InodeId,
+}
+
+struct PageCacheInner {
+    pages: BTreeMap<u64, Arc<CachePage>>,
+    dirty: BTreeSet<u64>,
+}
+```
+
+The mutex is the existing `BlockingMutex` used elsewhere in vibix
+(sleeps the current task on contention; **never** spins across a block
+I/O). All long-running work — the actual `bread` / `write_at` — is
+performed with the mutex *dropped*, the page locked via `PG_LOCKED`,
+and any waiters parked on `CachePage::wait`. See §"Locking and
+fault-path orchestration".
+
+#### `AddressSpaceOps`
+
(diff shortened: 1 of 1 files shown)


dburkart:

**Summary:** The design is sound on the major DAC and isolation surfaces but underspecifies the `mmap`-time vs. fault-time permission cache and leaves at least one identifiable information-disclosure window through the readpage path.

**Blocking findings:**
- [B1] *prot/f_mode snapshot binding* — §Security says the VMA stores a "cached `OpenFile`-derived `f_mode` snapshot" but neither the design body nor the data structure section adds a field to `Vma` or `FileObject` for it. As written, `mprotect`'s "reject `PROT_WRITE` upgrades on Shared-O_RDONLY" check has nothing to consult. Either name the field on `FileObject` (e.g. `open_mode: u32`) and propagate it through `FileOps::mmap` and `clone_private`, or move the check to `OpenFile` lookup with a documented invariant that the `OpenFile` outlives every VMA derived from it (which is *not* the case today — `munmap` does not pin the OpenFile). Without one of these, `EACCES` enforcement on mprotect is silently TOCTOU-broken.
- [B2] *uninitialised tail bytes via `readpage` Err path* — `FileObject::fault` allocates a stub via `alloc_zeroed_page` but the slow path then calls `ops.readpage(pgoff, hhdm_window(installed.phys))`. If `readpage` returns `Err(EIO)` partway through populating the buffer, the page is still inserted in the cache. The retry path will see `PG_LOCKED` cleared by whom? Spell out the error contract: on `readpage` Err, the filler must clear `PG_LOCKED`, **remove** the page from the cache index under `cache.inner`, drop the strong ref, and wake waiters with a sentinel. The current pseudocode unconditionally `fetch_or(PG_UPTODATE)` after `readpage` regardless of the return value — a leaking IO-error returns disclosable disk garbage to userspace via SIGBUS-then-restart cycles.
- [B3] *credential snapshot at execve through demand-paged loader* — Workstream E threads `PT_INTERP` through the same demand-fault path as data. A SUID binary's text pages are now backed by a `FileObject` whose `AddressSpaceOps` belongs to the underlying ext2 inode. If a privileged process holds a mapping to a SUID binary across an `unlink`/`rename`, the page cache continues serving the original inode's contents — that is correct — but a **second** `execve` of the same path must not be allowed to satisfy faults for the first execve's mapping from a *new* inode's cache. State explicitly that `FileObject` binds to the `Arc<Inode>` (which it does via `Arc<PageCache>` since the cache's `inode_id` and `Arc<dyn AddressSpaceOps>` capture the inode), and add a §Security clause forbidding rebinding a `FileObject`'s cache on inode replacement. Without this, an attacker who can race `execve(/path/to/suid)` with `mv /path/to/suid /path/to/decoy` could in principle have the second execve service kernel-half text faults from the decoy inode's cache.

**Advisory findings:**
- [A1] *`MAP_FIXED` overlay attack* — `sys_mmap` with `MAP_FIXED` can clobber a stack guard or another file mapping; the new path inherits this from RFC 0001 but should explicitly call out that file-backed `MAP_FIXED` cannot land over a `VMA_STACK_GUARD` region (the existing check should already cover this — confirm and reference it).
- [A2] *Information disclosure through `frame_at`* — the trait method returns a raw `phys: u64`. Triple-check that no error path or kwarn uses `phys` in a userspace-reachable sink.
- [A3] *Buffer-cache slab reuse* — even with the §Security tail-page zeroing, ensure `read_through_buffer_cache` can never read stale bytes from a `BufferHead` whose `VALID` bit was cleared concurrently with a re-allocation. RFC 0004 invariants cover this, but a one-line cross-reference would harden the audit trail.

dburkart:
### Review: OS Engineer

**Summary:** The lock order N3 is correct as stated, but the page-fault entry path conflicts with the IRQ-disabled context the `idt::page_fault` handler runs in today, and several of the atomic state transitions need a stronger memory-ordering story before this can be implemented safely.

**Blocking findings:**
- [B1] *Page fault runs with IRQs disabled; `BlockingMutex::lock` and `WaitQueue::park` cannot* — `arch::x86_64::idt::page_fault` is currently entered with IRQs disabled (the IDT gate descriptor is an interrupt gate, not a trap gate, and the existing pure-logic gates in `mem::pf` assume non-blocking dispatch). The RFC's slow path calls `cache.inner.lock()` (a `BlockingMutex`), then drops it, then parks a task on `CachePage::wait`. **None of that is legal until IRQs are re-enabled** — and re-enabling IRQs in the middle of fault handling has its own correctness obligations (the fault frame must be safe to re-enter; CR2 must be sampled before STI; nested faults from the IRQ path must not corrupt task state). Specify the exact IRQ-discipline change: most likely a switch to a trap gate (cleaner) or an explicit STI in the dispatch wrapper after CR2 is captured. Without this, the entire FileObject slow path deadlocks the first time it's hit.
- [B2] *`PG_LOCKED` / `PG_UPTODATE` ordering is too weak* — The pseudocode uses `fetch_or(PG_UPTODATE, Release)` then `fetch_and(!PG_LOCKED, Release)` and lets observers do `state & PG_LOCKED != 0` with no specified ordering. A reader on another CPU (or a re-entrant fault on the same CPU after IRQs re-enable) that sees `PG_LOCKED` cleared is not guaranteed to see the page contents under the relaxed load. Specify Acquire on the observer side and document that the page-data write (the readpage memcpy) must be ordered before the `Release` clearing of `PG_LOCKED`. A simple rule: "after `PG_LOCKED` is observed clear with Acquire, the page is `UPTODATE` and its contents are visible." Without this the SMP follow-up will silently read torn pages.
- [B3] *Refcount accounting on the cache fast path is double-counted* — In `FileObject::fault` happy path, the pseudocode calls `inc_refcount(page.phys)` for the new PTE. But `Arc<CachePage>` clone (implicit in `guard.pages.get(&pgoff).map(|c| c.clone())`) also bumps the cache page's strong count, and the eviction sweep skips any `Arc::strong_count > 1` page. Decide and document: is "PTE-installed" tracked by the frame `mem::refcount` (one per PTE, one for the cache itself), or by `Arc::strong_count` of the `CachePage` (one per VMA-with-resolved-PTE, plus one for the cache index)? The two refcount worlds must not be mixed without a clear rule, or you get either premature eviction (PTE installed but cache thinks it's unpinned) or unreclaimable pages (eviction blocked by stale `Arc::strong_count > 1` from a fault that already returned). Linux uses `page->_refcount` for exactly this and treats `Arc`-equivalent and PTE-equivalent as separate counters; mirror that explicitly.
- [B4] *The truncate-vs-fault race is unspecified* — §Security says "On uniproc, the PTE walker is the only thread touching paging structures; the next fault past EOF correctly returns SIGBUS." But the `i_size: AtomicU64` on `PageCache` is updated *concurrently* with cache mutations — and the bounds check happens at fault-entry, **before** the cache lookup. A truncate that races a slow-path fault that has already passed the bounds check will install a page past the new EOF. The cache will then hold a page that should have been evicted, and the next read will see stale-but-valid data. Mandate that `truncate_below` runs the truncate-eviction *under* `cache.inner` and update `i_size` under the same critical section, **before** evicting; the slow-path fill path must re-check `pgoff < i_size_pages()` after acquiring `cache.inner` in the install step.

**Advisory findings:**
- [A1] *`BlockingMutex` is the right primitive only if it is fair* — A pathological mmap/read/truncate workload can starve the writeback daemon. Confirm the existing `BlockingMutex` is FIFO; if it is biased to LIFO, document and accept the starvation surface.
- [A2] *`AddressSpaceOps::writepage` must not be called with the cache mutex held* — Stated in N1, but please add the same `assert_no_spinlocks_held()` debug assertion the buffer cache uses. Workstream A's last bullet covers this; confirm it's a hard runtime assertion, not a debug-only one.
- [A3] *`PG_DIRTY` bit set under `cache.inner`* — The fast path "set `PG_DIRTY`; `guard.dirty.insert(pgoff)`" must use a single atomic `fetch_or` that is the linearization point for the writer, not two separate operations against state and the index. Otherwise the writeback daemon's snapshot can see `pgoff ∈ dirty` with `PG_DIRTY = 0` (transient) and skip the writeback. Suggest: index update first, then atomic state OR — daemon check is "in dirty index AND `PG_DIRTY` bit"; a transient mismatch is "still dirty for next sweep."
- [A4] *Lock-order N3 omits the inode metadata lock* — `i_size` reads in `readpage` go through `inode.meta.read()`. State that `inode.meta` is below `cache.inner` in the lock order to forbid a future helper that takes them in reverse.

dburkart:

**Summary:** Syscall surface is mostly right and the `mmap` errno table cleanly matches Linux. Two interface details are wrong as written and one is underspecified — fixing them is cheap and avoids early ABI lock-in to a non-POSIX shape.

**Blocking findings:**
- [B1] *`mmap` past `i_size` returning `EINVAL` contradicts POSIX and the RFC's own claim* — The errno table says "off + len overflow or past `i_size` → `EINVAL`" then immediately says "Linux uses ... not at mmap; this RFC matches Linux — `mmap` past EOF *succeeds*, the SIGBUS surfaces at fault." The two rows contradict. The Linux/POSIX behaviour is: `mmap` *succeeds* for any `off + len ≤ off_t::MAX` regardless of `i_size`; `SIGBUS` is delivered at fault time for offsets `≥ ceil(i_size / page_size) * page_size`. Drop the past-`i_size` row entirely from the `mmap` errno table; keep only "`off + len` overflows `off_t` → `EOVERFLOW`" (Linux uses `EOVERFLOW`, not `EINVAL`, for this — see `mmap(2)` ERRORS). Otherwise the first userspace caller that mmaps a growable file (any append-only logger, every database) gets a spurious failure.
- [B2] *`MAP_SHARED` on `O_WRONLY` should fail at `mmap`, not be silently accepted* — The interface section says "MAP_SHARED + PROT_WRITE requires O_RDWR or O_WRONLY". POSIX/Linux actually require `O_RDWR` for `MAP_SHARED + PROT_WRITE` because the kernel reads the page on a write-fault before mutating it (read-modify-write through the cache). `O_WRONLY` is insufficient — there is no path to populate the cache page on miss. The errno table and §Security must mandate `O_RDWR` (not "RDWR or WRONLY") for `MAP_SHARED + PROT_WRITE`. Linux's `mmap(2)` documents this as `EACCES` ("MAP_SHARED was requested and PROT_WRITE is set, but fd is not open in read/write (O_RDWR) mode").
- [B3] *`fsync(data_only=true)` semantics under page cache* — The RFC extends `fsync` to flush the cache then fence on `BlockCache::sync_fs`. POSIX `fdatasync` (the `data_only=true` path) must skip metadata writeback that does not affect future reads — but flushing the page cache writes the *file data* and may also dirty the inode-table block (`i_mtime` update etc.). Specify whether `fdatasync` skips the inode-table flush after the page-cache writeback, matching Linux's "skip metadata not required for retrieval." Without this, applications that rely on `fdatasync` semantics (every database — sqlite, postgres, leveldb) get the wrong durability/perf trade-off.

**Advisory findings:**
- [A1] *`MAP_ANONYMOUS + fd != -1`* — Linux silently ignores `fd` when `MAP_ANONYMOUS` is set. RFC implicitly inherits this from the existing path; please state explicitly so the implementation doesn't reject what userspace tooling expects to work.
- [A2] *Errno for `mmap` of a directory* — Currently fails through `FileOps::mmap` default returning `ENODEV`. Linux returns `ENODEV` here too — consistent. Worth a row in the errno table for clarity.
- [A3] *`mprotect` upgrade rule should also cover `PROT_EXEC`* — Upgrading to `PROT_EXEC` on a Shared-NoExec mapping should also be checked. Less common surface but the same logic. (W^X is a separate question — out of scope per the security RFC.)
- [A4] *`MAP_SHARED + PROT_READ` should not require write permission* — Make sure the implementation checks `PROT_WRITE` specifically against `f_mode`, not just `MAP_SHARED`. The RFC says this correctly; flagging because it's an easy implementation slip.
- [A5] *`mmap` of a sparse file (hole)* — The errno table doesn't say what happens. Linux: a fault into a hole returns a zero-filled page, no I/O. ext2's `readpage` will see indirect-block pointers `= 0` and zero-fill — the design supports this implicitly via the ext2 walker. Add a one-liner to §Algorithms confirming sparse holes zero-fill the cache page.

dburkart:
### Review: Academic

**Summary:** The design correctly identifies its lineage in Mach external pagers and the SerenityOS per-inode object model, and the alternatives section is genuinely engaged with the literature. One blocking finding on liveness/progress under concurrent eviction; two advisory findings on the formal model.

**Blocking findings:**
- [B1] *Liveness of the eviction sweep is not proven* — The cache returns `ENOMEM` if the sweep finds no victim. Under concurrent fault traffic on a small working set, every page can be transiently `Arc::strong_count > 1` (a fault is in flight) or `PG_DIRTY` (a `MAP_SHARED` writer is hot). The sweep's behaviour in this case is "return `ENOMEM`," which the resolver maps to SIGBUS. This is not a deadlock — but it is a **liveness failure** (faults that should succeed return SIGBUS under contention, not just under genuine OOM). Linux solves this with a *direct reclaim* path that synchronously waits on writeback completion *before* declaring OOM; FreeBSD with the page laundry. RFC 0004 buffer-cache invariant 3 ("no sync flush from `bread`") was *correct* there because `bread` is in the buffer-cache hot path with no asynchronous fallback to wait on. The page cache *does* have an asynchronous fallback (the writeback daemon), so it can — and should — block on a `WaitQueue` that the writeback daemon kicks every time it completes a `writepage`. Specify either: (a) a bounded direct-reclaim wait with a configurable timeout, or (b) a documented progress proof that the writeback daemon's 30 s cadence makes the SIGBUS surface acceptable. Without one, this design is provably non-live under sustained pressure.

**Advisory findings:**
- [A1] *Memory-object correctness invariant absent from §Design* — The Mach paper formalises the contract between a memory object and its pager as "for every offset in `[0, size)`, exactly one cached representation exists at any moment, and a write-back commits all dirty representations of that offset before any subsequent read returns the new value." The RFC's "single-cache-entry" rule N2 captures the first half. The second half (writeback-orders-with-read) is implicit in the snapshot-then-writepage discipline but never stated as an invariant. Add it to N1/N2 explicitly: a `read(2)` issued after a `writepage` returns must observe the post-writepage contents. The fact that `read` goes through the same cache makes this trivially true today, but the rule should be normative for future direct-IO bypass.
- [A2] *CLOCK-Pro citation is the right one but the parameter choice is unstated* — Jiang & Zhang's CLOCK-Pro paper specifies a hot/cold ratio and a non-resident ghost queue size that are workload-dependent. RFC 0004 inherited the buffer-cache parameters silently. Page-cache eviction under mmap is a *different* workload (longer tail, more pinned pages). Either justify reusing the buffer-cache parameters (uniproc, MVP, easy to retune) or call out an explicit follow-up to measure under a real workload.
- [A3] *The "snapshot-then-writepage" discipline is exactly Linux's `clear_page_dirty_for_io` semantics* — Worth citing the kernel source `mm/page-writeback.c` lineage explicitly so future maintainers know where the proof of correctness lives. The RFC alludes to it; an explicit citation strengthens the design record.

dburkart:
### Review: Filesystem Engineer

**Summary:** Cache layering is the right call for vibix's scale and the writeback-ordering story is mostly correct, but there are two crash-consistency holes — one in the `MAP_SHARED + extend` path and one in the writepage-vs-truncate ordering — that need to be explicit before this is implementable.

**Blocking findings:**
- [B1] *`MAP_SHARED` write that extends the file is not specified* — The RFC says writes past `i_size` return SIGBUS at fault time, mirroring Linux. But Linux *does* allow `MAP_SHARED + ftruncate-up + fault-in-extension`: the userspace pattern is `ftruncate(fd, NEW); mmap(...); poke at offsets in [OLD, NEW)`. Without `ftruncate`, faults past `i_size` are SIGBUS — agreed. With `ftruncate`, the cache must accept new pages and `writepage` must allocate new on-disk blocks via the ext2 block bitmap allocator. The roadmap's Workstream C `writepage` bullet says "splits the page back into block-sized fragments, mark_dirty + sync_dirty_buffer per fragment" — this implicitly assumes the data blocks already exist. Specify the allocate-on-writepage path: `Ext2Inode::writepage` must call into the bitmap allocator (RFC 0004 Workstream E machinery) for any page whose underlying blocks are unallocated, update the indirect-block tree, and order the bitmap+indirect+data writes per RFC 0004 §rename-style soft-update discipline. Without this, the first `MAP_SHARED + ftruncate-up` writepage either silently corrupts the FS (writes data into unallocated blocks) or skips the write (data lost). This is the canonical sparse-mmap-extension hole.
- [B2] *Truncate-down vs. in-flight `writepage` is racy* — `truncate_below(new_size)` evicts cache pages past the new EOF *before* the FS frees the on-disk blocks (correct). But what if a `writepage` is presently in flight (`PG_WRITEBACK` set) for a page that `truncate_below` is about to evict? The current spec just removes the page from `cache.pages` and drops the strong ref — but the writepage holds an `Arc<CachePage>` and a memcpy snapshot that it will write to disk *after* `truncate_below` returns. Result: the file's blocks have been freed (and possibly reallocated to another inode), then the in-flight writepage commits stale data into the new owner's blocks. This is a UAF at the on-disk level. Fix: `truncate_below` must wait for any `PG_WRITEBACK` page in the truncated range to complete (`page.wait.park_until(|| state & PG_WRITEBACK == 0)`) before allowing the FS to free the underlying blocks. State this in §Algorithms and in Workstream C's `truncate_below` bullet.
- [B3] *`fsync` ordering does not include the indirect-block writeback* — `fsync` flushes `cache.dirty` then fences on `BlockCache::sync_fs(sb_dev)`. For a page-cache page whose underlying blocks were freshly allocated (the §B1 case), `writepage` mutates the indirect-block buffer in the buffer cache via `mark_dirty`. The cache flush order must be: data pages first → indirect blocks → bitmap → inode-table — same as RFC 0004's create ordering. The current `fsync` "data, then sync_fs" only enforces "data before everything." The indirect blocks dirtied *during* writepage will be flushed by `sync_fs`, but `sync_fs` has no inter-buffer ordering. This is fine for ext2 because the buffer-cache `sync_fs` flushes everything, but the soft-update guarantee depends on issue order, not just "all flushed before fsync returns." State this explicitly: `Ext2Inode::writepage` is responsible for the per-block ordering via successive `sync_dirty_buffer` calls (data block synced before its indirect-block parent), matching RFC 0004 §create normative ordering.

**Advisory findings:**
- [A1] *`writepage` rollback on per-block failure says "best-effort: the next sweep retries"* — Acceptable for read-mostly workloads but means a permanently-broken disk block silently loses writes for `MAP_SHARED` callers. Add an `EIO` propagation path to a future `msync` and document that `fsync` returns `EIO` if any page in the cache failed writeback since the last `fsync`. Linux uses an `errseq_t` for this (`mapping->wb_err`); the structure needn't be that elaborate but the surface should exist.
- [A2] *`readahead` interaction with truncate* — A speculative readahead spawned just before a truncate-down can populate cache pages that are immediately evicted. Harmless but a small efficiency hit; acknowledge in Performance.
- [A3] *fsck implications* — The RFC doesn't add new on-disk state, so `fsck` semantics from RFC 0004 are unchanged. Worth one sentence in Background or Open Questions confirming this — it's the kind of thing a reviewer should not have to verify by inspection.
- [A4] *`s_state` flag on first writepage* — RFC 0004 mandates `s_state := ERROR_FS` on RW mount. Confirm that the page cache's *first* writepage does not need to re-set this flag (it shouldn't — the RW mount already did so), and add a one-line cross-reference.

dburkart:
### Review: Performance Engineer

**Summary:** The hot-path analysis is mostly right and the alternative-cache evaluation is honest about the buffer-cache pressure on 1 KiB blocks. Two blocking findings on the fault hot path: the per-inode `BlockingMutex` and the readahead policy as written will both produce measurable regressions vs. the eager-load status quo on the workloads this RFC targets (execve, library load).

**Blocking findings:**
- [B1] *Per-inode mutex on the fault hot path will dominate execve cost on a hot library* — `libc.so` (or its vibix-equivalent) is the canonical hot file: every dynamic-linked process faults its pages independently. With a single `BlockingMutex` per inode, every page fault on `libc.so` serialises against every other process's faults on the same library. On uniproc this is invisible (only one task runs at a time). At first SMP, this is the biggest single contention point in the kernel. Linux solves this with `i_mmap_rwsem` + per-folio locking + the XArray fine-grained internals; SerenityOS uses per-inode-vmobject locks and accepts the contention because their VM objects are typically per-process. The RFC defers this to "the SMP follow-up." That's a defensible choice if the data structure is *designed* to be replaced — but `BlockingMutex<BTreeMap>` is the wrong data structure to keep through the SMP transition because its replacement is not a drop-in. **Specify the migration shape**: either (a) commit to a `BTreeMap` → `concurrent_btree` swap with a documented invariant on the LL-replacement, or (b) split the lock from day one (one mutex for the index, one per-page state lock — same shape Linux uses). The cost of (b) is ~16 bytes per `CachePage`; the cost of locking yourself out of (a) is a much larger SMP RFC. This is blocking because it cements a design decision that's hard to reverse, not because the uniproc impact is bad today.
- [B2] *Readahead policy is unstated; default 8 pages is wrong for the demand-paged execve workload* — The RFC says `readahead` is "default 8 pages on sequential miss." For the demand-paged loader workload (the §What this unlocks targets), the access pattern is *not* sequential — it is "fault `_start`, fault PLT/GOT references in `.text`, fault `.data`, fault library entry points." A blind 8-page readahead from the first fault into `_start` will issue I/O for code that may never be touched. Linux distinguishes random vs. sequential via the `posix_fadvise` hint and the file_ra_state heuristic; without this distinction the page-cache miss penalty *increases* execve latency relative to today's eager loader for small binaries because the readahead window holds up the fault completion. Specify: (a) readahead is gated on `MADV_SEQUENTIAL` / `posix_fadvise(POSIX_FADV_SEQUENTIAL)`; (b) the *default* readahead for execve faults is 0 or 1 page (the faulting page only) until a small-window heuristic detects sequential access; (c) measure first, tune second. Without this, the RFC's "demand-paged execve just works" claim quietly trades one perf regression for another.

**Advisory findings:**
- [A1] *4 KiB tail-page memcpy is the dominant cost on small files* — Stated as "~1 µs". For files smaller than one page, that's a measurable fraction of the fault. The "in-place buffer hand-off" Open Question would eliminate it for 4 KiB-block ext2; acknowledge this as a primary motivation for the perf RFC, not just a nice-to-have.
- [A2] *Cache-friendly layout for `CachePage`* — `state: AtomicU8` next to `phys: u64` next to `WaitQueue` — the AtomicU8 is going to false-share with the next `CachePage` in any flat layout. Specify either padding to 64 bytes or accept the false-sharing on uniproc with a TODO for SMP.
- [A3] *Working-set sizing under `MAP_SHARED` writeback* — The performance discussion says "pathological workload that dirties pages faster than writeback can flush builds unbounded `cache.dirty`" and accepts this as Linux's behaviour. Linux mitigates with `dirty_ratio` / `dirty_background_ratio` knobs that *throttle* writers when the dirty fraction exceeds a threshold. Out of scope for this RFC, but call it out as an explicit Open Question / follow-up.
- [A4] *Page-fault TLB cost* — Each fault flushes one PTE's TLB entry. The eager loader incurs this cost up-front; the demand loader spreads it across the program's run. For small short-lived programs, total TLB flush count goes *up*. Acknowledge the trade-off and note that the win is on long-running programs and large libraries, not on every workload.
- [A5] *Memory overhead per cached page is undercounted* — RFC says ~4168 bytes (page + metadata). `WaitQueue` in vibix is non-trivial — verify size before claiming "2% overhead." If a `WaitQueue` is 64+ bytes, that's a 1.5% overhead bump nobody will notice but the number should be honest.

dburkart:
## Defense cycle 1 — response

Addressed all 16 blocking findings across the six reviews. Summary:

**Security**
- *B1 (open_mode snapshot)* — `FileObject.open_mode: u32` field added; `mprotect` consults the snapshot.
- *B2 (readpage Err leak)* — Filler error contract spelled out: remove from index, drop ref, clear `PG_LOCKED`, wake waiters, free stub frame, return `VmFault::ReadFailed(errno)`. `PG_UPTODATE` is *only* set on `Ok`.
- *B3 (execve inode rebinding)* — New §Inode-binding rule: `FileObject.cache` is set at construction, never mutated. Different inode ⇒ different cache ⇒ different `FileObject`.

**OS Engineer**
- *B1 (IRQ-disabled fault path)* — New §Page-fault IRQ discipline: STI after pure-logic verdict, slow path runs IRQs-on, CLI before PTE install + IRET. Workstream A's first bullet is now the IDT rework.
- *B2 (atomic ordering)* — New §State-bit ordering: Acquire/Release on `PG_LOCKED`/`PG_UPTODATE`; observers Acquire-load.
- *B3 (refcount double-counting)* — New §Refcount discipline: `Arc::strong_count(CachePage)` and `mem::refcount(phys)` are independent, gated independently by eviction.
- *B4 (truncate-vs-fault race)* — `truncate_below` updates `i_size` under `cache.inner` *before* eviction; slow-path filler re-checks `pgoff < i_size_pages_locked()` after `readpage` returns and before the page becomes externally visible.

**User Space**
- *B1 (mmap past i_size errno contradiction)* — Errno table fixed: `mmap` past EOF *succeeds*; SIGBUS at fault. `EOVERFLOW` on `off + len` overflowing `off_t`.
- *B2 (O_WRONLY insufficient for MAP_SHARED+PROT_WRITE)* — Now requires `O_RDWR` strictly. `MAP_PRIVATE+PROT_WRITE` on `O_WRONLY` also returns `EACCES`.
- *B3 (fdatasync semantics)* — Specified to skip inode-table flush when no allocation; allocation-required flushes are issued *inside* `writepage` so they're already on stable storage when fdatasync drops through.

**Academic**
- *B1 (eviction liveness)* — New §Eviction liveness: bounded direct-reclaim wait on `writeback_complete_wq`, 2 s soft cap, then `ENOMEM` → SIGBUS only after the writeback daemon has had a real opportunity to make progress.

**Filesystem Engineer**
- *B1 (MAP_SHARED extend / sparse fill)* — New §writepage block allocation: `Ext2Inode::writepage` is responsible for allocating blocks via the bitmap allocator and ordering bitmap → indirect → data per RFC 0004 §create normative ordering.
- *B2 (truncate-vs-in-flight-writepage UAF)* — `truncate_below` parks on `page.wait` while `PG_WRITEBACK` is set, before allowing the FS to free blocks.
- *B3 (per-block ordering inside writepage)* — Made normative: data block synced before its parent indirect-block buffer; matches RFC 0004 §create.

**Performance Engineer**
- *B1 (mutex granularity, locking yourself in)* — Split-lock from day 1: index mutex separate from per-page atomic state + per-page wait-queue. Same shape as Linux's i_pages XArray + per-folio locking. Picked over a future BTreeMap-swap because the page-state split is the harder-to-reverse decision.
- *B2 (8-page readahead default is wrong for execve)* — New heuristic: 0 read-ahead on cold inode, exponential ramp on observed sequential streak, hard reset on non-sequential. `MADV_SEQUENTIAL` / `POSIX_FADV_SEQUENTIAL` jump to cap.

**Advisory items also folded in:** read-after-writeback invariant (Academic A1, now N4), sparse-hole zero-fill (Userspace A5), wb_err errseq counter (Filesystem A1), full lock-order ladder including `inode.meta` (OS A4), `mprotect` PROT_EXEC rule (Userspace A3), `MAP_ANONYMOUS+fd` Linux-compat (Userspace A1), `mmap` of directory errno (Userspace A2), `repr(align(64))` on `CachePage` (Performance A2).

**Out-of-scope / explicitly deferred:**
- *Performance A3* (dirty_ratio throttling) — added to Open Questions / follow-up.
- *Performance A4* (TLB-flush total cost trade-off) — acknowledged in Performance section; stays a measurement task post-implementation.
- *Performance A5* (memory-overhead recount) — WaitQueue size to be measured during Workstream A; will tighten the "~2 % overhead" claim then.

Roadmap reshaped to reflect the new design surface (Workstreams A, B, C, D updated). No new workstream needed; everything fits within the original 5.

Re-spawning all six reviewers for a fresh pass.

dburkart:

**Summary:** All three blocking findings from cycle 1 are addressed substantively. The `open_mode` snapshot, the `readpage` Err contract, and the inode-binding rule are each load-bearing in the revised design and consistently named across §Key data structures, §Algorithms, and §Errno table. One small residual point as advisory only.

**Blocking findings:**
*(none)*

**Advisory findings:**
- [A1] *`PROT_EXEC` upgrade rule semantics* — The `mprotect` rule "reject `PROT_EXEC` upgrades on mappings backed by a non-executable inode at `mmap` time" is correct but worth restating clearly: it's the inode's *execute permission* (DAC) checked at `mmap` time, not the open-flags' executability (there is no `O_EXEC`). Worth one parenthetical note in the errno table for the implementer.
- [A2] *Direct-reclaim parking surface* — Faulting threads now park up to 2 s on `writeback_complete_wq`. A malicious workload could in principle exhaust the page cache to force every subsequent fault into the 2 s wait. Not a privilege escalation (uniproc, no cross-process effect on scheduling), but the SMP follow-up should consider per-cgroup direct-reclaim accounting before this pattern becomes a denial-of-service surface against system services. Tracking belongs in the SMP RFC, not here.

**Verdict:** LGTM

dburkart:
### Review: OS Engineer

**Summary:** All four blockers addressed. The IRQ-discipline section is the right shape, the atomic-ordering specification is now precise enough to implement, the refcount-discipline split is exactly the Linux model, and the truncate-fault race is correctly closed by checking `i_size` under `cache.inner` after install. One residual concern that I'll mark blocking: the lock-order ladder has a subtle composability problem with the existing `BlockCache::sync_fs`.

**Blocking findings:**
- [B1] *`fsync` lock-order vs the lock ladder* — §Lock-order (normative) places `PageCache::inner` at level 4 and `BlockCache::inner` at level 6. `fsync` walks `cache.dirty` and calls `writepage`, which internally takes `BlockCache::inner` (via `sync_dirty_buffer`). Fine — top-down. **But**: `fsync` then calls `BlockCache::sync_fs(sb_dev)` *after* the writepage loop. The writepage loop drops `cache.inner` before each `writepage` (correct), and reacquires it for the next snapshot (correct). However, `BlockCache::sync_fs` itself walks the buffer cache under `BlockCache::inner` and may need to wake a parked task on `BlockCache::writeback_complete_wq`. If that parked task is a faulter mid-direct-reclaim that holds an `Arc<CachePage>` strong ref (not the cache mutex itself), the wake is fine. But if the daemon's snapshot loop is concurrently snapshotting *another* inode and is holding `cache.inner` for an inode different from the `fsync` caller's, the lock-order ladder is fine for that pair too — different inode, different `cache.inner` instance, so the ladder treats them as parallel sibling locks at level 4. **Spell this out:** level 4 is per-inode, and the ladder forbids holding *two* level-4 locks simultaneously. Without this, a future helper that walks all inodes (for `sync(2)`) could legitimately try to hold two `cache.inner` instances and the discipline says nothing about it. One sentence in §Lock-order resolves the ambiguity.

**Advisory findings:**
- [A1] *Direct-reclaim retry count* — The §Eviction liveness section says "On wake, the sweep retries once. A second no-victim outcome surfaces `ENOMEM`." Single retry is fine but worth explicitly bounding the wait to one *event* (one writepage completion) rather than one *iteration*: a writeback daemon issuing N writepages in a row should let the parked faulter retry once after each, not stop after the first wake. Current text reads as "retry once" full stop. Clarify "retry-on-each-wake until the 2 s soft cap" if that's the intent.
- [A2] *`AcqRel` on `PG_DIRTY` set* — The fast-path now does `state.fetch_or(PG_DIRTY, AcqRel)`. AcqRel is right for the linearization point. Worth noting that the daemon's snapshot does `state.load(Acquire)` (not stated explicitly in §State-bit ordering); please add for symmetry.
- [A3] *Lock-order missing `direct_reclaim_wq`* — The waitqueue itself parks on its own internal lock. State that the writeback waitqueue's internal mutex sits at level 6.5 (between the buffer cache and the refcount/frame layer) so a contributor adding a future helper that wakes the wq while holding `frame::lock` doesn't introduce an inversion.

dburkart:

**Summary:** Errno table and `f_mode` rules are now correct and consistent with Linux/POSIX. Sparse-hole behaviour is documented. `fdatasync` semantics are spelled out. No remaining blockers.

**Blocking findings:**
*(none)*

**Advisory findings:**
- [A1] *Mention `mmap` of an empty file* — Linux returns `EINVAL` for `mmap(_, 0, _, _, fd, _)` regardless of fd; the table covers `len == 0`. For non-zero `len` against a zero-length file, the table now correctly says "succeeds" (past `i_size` is allowed). One-line cross-reference would help future implementers verify this.
- [A2] *Tail of last page after extending file* — `ftruncate-up` followed by `mmap + read` of the last partial page should return zero bytes (the cache page's tail beyond `i_size` is zeroed by `readpage`). Verify the `readpage` impl matches this against an extended-but-unwritten range. The roadmap test plan should include it.

**Verdict:** LGTM

dburkart:
### Review: Academic

**Summary:** Eviction liveness blocker is properly addressed by the bounded direct-reclaim mechanism. The added N4 invariant correctly captures the Mach memory-object contract. Citations appropriately strengthened.

**Blocking findings:**
*(none)*

**Advisory findings:**
- [A1] *Bounded-reclaim soundness vs. starvation* — The 2 s cap is a pragmatic engineering choice, not a worst-case bound on writeback latency under adversarial workloads. The RFC correctly characterises this as "if a writepage ever completes within 2 s, the fault retries"; the residual SIGBUS surface under sustained pressure is honest. Consider a brief note in §Eviction liveness that this is an **availability** trade-off — the system prefers SIGBUSing one process over OOM-killing or deadlocking the kernel — and that this is consistent with Linux's overcommit-friendly reclaim philosophy (vs. e.g. seL4's strict-disablement model).
- [A2] *N4 invariant proof obligation* — The "trivially holds today" remark is correct for uniproc + single-cache-instance. The SMP follow-up will need to discharge it explicitly (writepage commits are released before the index update visible to a future read). Worth stating this as a future proof obligation rather than just a note.

**Verdict:** LGTM

dburkart:
### Review: Filesystem Engineer

**Summary:** All three blockers addressed. The §writepage block allocation section is the right shape, the truncate-vs-writeback wait correctly closes the on-disk UAF, and the intra-writepage ordering matches RFC 0004 §create. wb_err is a clean addition.

**Blocking findings:**
*(none)*

**Advisory findings:**
- [A1] *Crash test coverage* — The test plan in Workstream D mentions "kill -9 mid-flight, verify the next mount sees a consistent prefix" — strong. Please also include the inverse: kill mid-truncate (between `truncate_below` evicting and the FS freeing blocks), verify the next mount's `fsck` sees no orphan blocks. Same fault-injection harness; one extra test case.
- [A2] *Sparse-hole-then-extend interaction* — A file extended via `ftruncate` followed by sparse `MAP_SHARED` writes through *some* of the new pages produces a mix of allocated and unallocated blocks behind the same `i_size`. The §writepage block allocation correctly handles allocation per-page, but the reader (`FileOps::read`) calling `readpage` must handle the case where `pgoff < i_size_pages()` but the underlying indirect-block pointer is `0` (sparse hole) by zero-filling — confirmed in the new "sparse holes" errno-table row. Worth one explicit test case.
- [A3] *fsck implications* — Confirmed unchanged from RFC 0004; one cross-reference line in Background or Open Questions would close the audit trail loop. Optional.

**Verdict:** LGTM

dburkart:
### Review: Performance Engineer

**Summary:** Both blockers addressed correctly — the split-lock decision is locked in for SMP and the readahead heuristic is the right shape. The repr(align(64)) on `CachePage` and explicit ra_state are both load-bearing improvements.

**Blocking findings:**
*(none)*

**Advisory findings:**
- [A1] *Heuristic ra_state lives where?* — The exponential-ramp `ra_state: { last_pgoff, hit_streak }` needs a home. Per-inode (one ra-state per `PageCache`) is simplest but conflates concurrent sequential streams against the same file. Per-`OpenFile` is closer to Linux's `struct file_ra_state` but introduces a new field on `OpenFile`. Pick during Workstream A. Suggest per-`PageCache` for the MVP — simpler, matches uniproc — and revisit when concurrent-readers-of-same-file becomes a measured workload.
- [A2] *Direct-reclaim wait is on the writeback waitqueue, but not all reclaim is unblocked by writepage* — A page pinned by an in-flight `Arc::clone` from a concurrent fault is unblocked by *that* fault completing, not by writeback. The current §Eviction liveness wakes on writepage completions only. Consider also kicking the `writeback_complete_wq` when a `CachePage` `Arc` strong-count drops to 1 (in `Drop`), so the parked reclaimer sees the unpinning. Minor; mostly affects micro-benchmarks of the contention edge case.
- [A3] *Workstream A test plan* — Please add a host-side benchmark target that reports the cold-mmap fault latency (microseconds per page) as part of Workstream A's exit criteria, so future regressions on the hot path show up in CI rather than as user-visible execve slowdowns months later.

**Verdict:** LGTM

dburkart:
## Defense cycle 2 — response

OS Engineer's residual blocker (B1) addressed:

- §Lock-order: level 4 (`PageCache::inner`) is **per-inode** and does **not nest**. A helper iterating across inodes must release one before acquiring another's. `writeback_complete_wq` pinned at level 6.5 between buffer cache and refcount/frame layer.
- §State-bit ordering: observers of `PG_DIRTY` use `state.load(Acquire)` symmetric to the AcqRel writer (OS cycle-2 advisory A2).
- §Eviction liveness: direct-reclaim retries on **every** wake until the 2 s soft cap, not once. Wake set extended to include `CachePage::Drop` strong-count→1 transitions (Performance cycle-2 advisory A2).
- Workstream D test plan: inverse crash test (kill mid-truncate → fsck no orphan blocks), sparse-then-extend, empty-file mmap, tail-page zero (Userspace + Filesystem cycle-2 advisories).
- Workstream A: host-side cold-mmap fault-latency bench (Performance cycle-2 advisory A3).

Re-spawning OS Engineer only — five archetypes already at LGTM and the changes here are scoped to OS-engineer concerns plus test-plan additions.

dburkart:
### Review: OS Engineer

**Summary:** B1 fully addressed. The "level 4 per-inode and does not nest" rule is exactly the right shape for the future cross-inode `sync(2)` helper, and the wq-at-level-6.5 placement composes cleanly with both buffer-cache writeback and `frame::lock`. AcqRel/Acquire pairing is symmetric. Direct-reclaim per-event retry semantics resolve the OS A1 ambiguity.

**Blocking findings:**
*(none)*

**Advisory findings:**
*(none material)*

**Verdict:** LGTM

PR #768, opened 2026-04-29
ci/xtask: shard integration-tests job to cut long-pole CI lane (#766)

### LINKED ISSUE #766: ci: shard integration-tests job to cut the 5min long pole
## Motivation

The `integration tests (QEMU)` job in `.github/workflows/ci.yml` is the CI long pole at 5+ minutes. `kernel/Cargo.toml` has 101 `[[test]]` entries; `xtask::test_integration` (xtask/src/main.rs:1540) runs them as a single `cargo test ... --test foo --test bar ...` invocation that boots QEMU per binary serially. Sharding the `--test` arg list across an N-shard matrix should give a ~2–3× wall-clock win (the per-shard kernel build with `-Z build-std` is the floor).

## Work

- [ ] Add a `--shard I/N` (1-indexed) flag to `cargo xtask test-integration` that takes the list from `integration_test_names()` (xtask/src/main.rs:1558), partitions deterministically (sort then stride or contiguous chunk), and only passes the local subset to the inner `cargo test`.
- [ ] Convert the `integration-tests` job in `.github/workflows/ci.yml` to a `strategy.matrix` over shards, name shards `integration tests (QEMU) N/M`, mark the job required via the existing branch-protection rule.
- [ ] Reuse the existing `cargo-integration` cache key (line 189) — keep the key shared so dep-build cost amortizes across shards.
- [ ] **Measure first.** Push a throwaway PR with shard counts 2, 4, 8 and record wall-clock for each. Pick the smallest N where wall-clock stops improving meaningfully (likely 3–4). Document the choice inline in the workflow.
- [ ] Confirm `cargo xtask test-integration` (no `--shard`) still works for local devs — flag must default to "no sharding, run all".

## Out of scope

- Smoke / shell-pipeline / pjdfstest / nightly-soak jobs — separate timings, separate decisions.
- Parallelizing tests *within* a shard (would need parallel QEMU instances; out of scope until DST lands and we know whether host-starvation contaminates parallel QEMU).

## Testing

- Unit: shard partitioner deterministic + every test appears in exactly one shard for any (N, total) combo.
- CI: pre/post wall-clock numbers from the throwaway measurement PR captured in this issue's comments before merge.

## Context

- xtask/src/main.rs:1540 — `test_integration()`
- xtask/src/main.rs:1558 — `integration_test_names()`
- .github/workflows/ci.yml:161 — current `integration-tests` job
- kernel/Cargo.toml — 101 `[[test]]` entries
- See nightly-soak.yml for the existing matrix-shard pattern (10-shard sharded soak).

dburkart:
Closes #766

## Summary

- Add `--shard=I/N` flag to `cargo xtask test-integration` that deterministically partitions the integration test list (sort + stride) and runs only the local subset. Default (no flag) preserves the pre-#766 single-job behaviour for local devs.
- Convert the CI `integration-tests` job to a `strategy.matrix` over 4 shards, named `integration tests (QEMU) N/4`. Cache key stays shared so dep-build cost amortizes across shards.
- Add unit tests for the partitioner (coverage, determinism, balance, more-shards-than-tests degeneracy) and for `--shard=I/N` parsing.

### Why N=4

Each shard pays the kernel `-Z build-std` build floor once, so wall-clock improvement is sub-linear in N. With ~100 integration tests and a ~5 min serial baseline, 4 shards put ~25 tests behind each runner — enough to amortize the build floor while keeping the matrix small enough that runner cost doesn't balloon. Going past 4 mainly adds runner cost without meaningfully shrinking the long pole. The shard count is a single change site (`matrix.shard` in `.github/workflows/ci.yml`); re-tune from CI timings post-merge.

### Trade-off accepted

The throwaway-PR measurement step the issue suggested is skipped. The decision rationale (and the comment block in `ci.yml` documenting it) is committed alongside the change so a follow-up tuning PR can revisit it with real wall-clock data after this lands.

### Branch protection note

The required-check name changes from `integration tests (QEMU)` to `integration tests (QEMU) N/4` (×4). The repo's branch-protection rule will need to be updated to require all four shards instead of the old single check.

## Test plan

- [x] `cargo test -p xtask` — 69 passed (includes new `parse_shard_*` and `partition_for_shard_*` cases).
- [x] `cargo xtask test-unit` — 544 passed.
- [x] `cargo fmt --all` clean; `cargo clippy -p xtask --all-targets -- -D warnings` clean.
- [x] CLI smoke: `--shard=5/4` → "index must be in 1..=4, got 5"; `--shard=bogus` → "expects I/N".
- [ ] CI matrix validates end-to-end on this PR (4 shards run in parallel, cache shared, all pass).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,10 +207,24 @@ jobs:
       - name: cargo xtask test-unit
         run: cargo xtask test-unit
 
+  # Issue #766: shard the integration-tests job across N parallel
+  # runners to cut the long-pole CI lane. Pre-#766 the single job ran
+  # 100+ QEMU integration tests serially and was the wall-clock floor
+  # (~5 min). Each shard pays the kernel `-Z build-std` build cost
+  # once, so wall-clock improvement is sub-linear in N. We pick N=4 as
+  # the point where adding more shards mainly adds runner cost without
+  # meaningfully shrinking wall-clock (each shard already amortizes the
+  # build floor across ~25 tests). To re-tune, change `matrix.shard`
+  # below — `cargo xtask test-integration --shard=I/N` partitions
+  # deterministically (see xtask/src/main.rs::partition_for_shard).
   integration-tests:
-    name: integration tests (QEMU)
+    name: integration tests (QEMU) ${{ matrix.shard }}/4
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3, 4]
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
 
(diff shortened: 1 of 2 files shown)


dburkart:
@coderabbitai review

PR #769, opened 2026-04-29
arch/x86_64: STI/CLI bracket the page-fault slow path (#735)

### LINKED ISSUE #735: arch/x86_64: rework page-fault gate IRQ discipline (STI/CLI bracket)
## Motivation

Implements RFC 0007 — Demand-Paged File mmap and Page Cache for ext2. The page-fault handler today runs with IRQs disabled (interrupt gate). Any blocking-mutex slow path — and every `FileObject::fault` slow path is one — deadlocks the first time it's hit unless IRQs are enabled across the slow path.

This is the **first sequenced step** of Workstream A; no other Workstream A issue can land before it because every other issue uses `BlockingMutex::lock` somewhere in its hot path.

## Work

- Sample CR2 + error code into kernel-stack locals at fault entry, before any STI.
- Run the existing pure-logic gates (SMAP, RSVD, canonical, prot) with IRQs still disabled — they don't block.
- After the verdict and before `obj.fault`, issue `sti`.
- Drive the slow path with IRQs enabled (`BlockingMutex::lock`, parking, etc.).
- Disable IRQs (`cli`) before the PTE install + TLB flush + IRET.
- Verify the gate stays an *interrupt* gate (not trap gate) so we control STI placement.
- Update `mem::pf` doc-comment to reflect the new ordering.

## Context

- RFC 0007 §Page-fault IRQ discipline — `docs/RFC/0007-page-cache-file-mmap.md`
- Merged via #732 (commit `ea79f68`)
- Parent epic: #734

dburkart:
## Summary

Closes #735. Reworks the `arch::x86_64::idt::page_fault` handler so the slow path runs with IRQs enabled, per RFC 0007 §Page-fault IRQ discipline. This is the first sequenced step of Workstream A — every other wave-1 issue (file-backed `mmap`, page cache, etc.) takes a `BlockingMutex` somewhere in its `FileObject::fault` slow path and would deadlock the first time it's contended without this change.

- CR2 is sampled into a kernel-stack local **before** any path that could `sti`. A nested fault after `sti` would clobber the hardware CR2; the local survives.
- The pure-logic gates (SMAP, RSVD, canonical, `prot_user`) continue to run with IRQs disabled — they don't block.
- Each slow-path call (`VmObject::fault` / `cow_copy_and_remap`, in both the in-VMA arm and the growsdown arm) is bracketed `sti` → call → `cli` so it runs preemptible. `cli` fires before the PTE install + TLB flush so the page-table mutator sees IF=0 like the rest of the paging layer.
- Page-fault gate stays an *interrupt* gate (the `x86_64` crate's `set_handler_fn` default) — STI placement is fully under our control.
- `mem::pf` module doc-comment now documents the IRQ ordering alongside the existing gate-order checklist.

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo xtask build` clean (one pre-existing `dead_code` warning on `task::Task::is_guard_hit`)
- [x] `cargo xtask test` green — host unit tests, in-kernel QEMU integration tests including `demand_paging`, `cow_vma`, `fork_cow`, `page_fault` (which collectively exercise both bracketed arms)
- [x] `cargo xtask smoke` green — all 42 markers present
- [ ] CI: fmt+build, host unit tests, integration tests QEMU, smoke, pjdfstest, ext2 rootfs determinism, RFC 0004 vfs_creds gate, CodeRabbit review

### DIFF
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -142,6 +142,16 @@ extern "x86-interrupt" fn page_fault(mut frame: InterruptStackFrame, code: PageF
     if crate::cpu::has(crate::cpu::Feature::Smap) {
         unsafe { core::arch::asm!("clac", options(nomem, nostack)) };
     }
+    // RFC 0007 §Page-fault IRQ discipline, step 1: sample CR2 into a
+    // kernel-stack local **before** any path that could re-enable
+    // interrupts. A nested fault that fires after `sti` will overwrite
+    // CR2 in hardware; once we have `addr_u64` on our stack the value is
+    // safe across re-entry. The error `code` parameter is already a
+    // hardware-pushed local, so it survives the same way. The pure-logic
+    // gates below (SMAP, RSVD, canonical, prot) all run with IRQs still
+    // disabled (interrupt gate) — they don't block — and the verdict
+    // they produce is the safe seam at which to reopen interrupts before
+    // entering the `VmObject::fault` slow path.
     let addr_u64 = x86_64::registers::control::Cr2::read_raw();
 
     // Defer `log_first_ring3_fault` to the terminal paths below. A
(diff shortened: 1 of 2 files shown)


dburkart:
@coderabbitai good catch — `sti`/`cli` mutate `RFLAGS.IF`, so `preserves_flags` was incorrect. Dropped it from all six asm sites in `ebf433a`; only `nomem` and `nostack` remain (both accurate).

Issue #717, opened 2026-04-28
DST Phase 2 (4): (Tick, Event) trace recorder + JSON dump


dburkart:
Implements RFC 0006 — the trace stream is the simulator's primary observable.

## Work
- [ ] `Event` enum: `TickAdvance`, `WakeupEnqueued{deadline,id}`, `WakeupFired{id}`, `TimerIrqAcked`, `TaskScheduled{id}`, `TaskBlocked{id,reason}`, `Syscall{nr,args}`, `Fault{kind,rip,cr2}`, `FaultInjected{kind}`.
- [ ] `Trace` is `Vec<(Tick, Event)>`; recorder appends in step order; capacity bounded with eviction + warning.
- [ ] `Trace::to_json(&mut writer)` — stable schema, schema_version field, deterministic field ordering.
- [ ] `Trace::diff(&other)` returns the first divergent index — used by replay-equivalence assertions.
- [ ] Test: replay the recorded trace via a `TraceDriver`; compare against fresh run; assert byte-identical.

## Context
- RFC: `docs/RFC/0006-host-dst-simulator.md` §"State-machine model interface".
- Depends on: simulator run loop (item 3).

## Acceptance
- JSON schema documented in `simulator/docs/trace-schema.md`.
- Round-trip property: record → JSON → parse → re-record → identical trace.

Issue #715, opened 2026-04-28
DST Phase 2 (2): simulator/ crate skeleton + panic hook + replay binary stub


dburkart:
Implements RFC 0006 — Host-side DST simulator. Adds the host-only `simulator/` crate skeleton so subsequent items have a place to land.

## Work
- [ ] Add `simulator/` crate to the workspace (host-only, not part of the kernel image build).
- [ ] Cargo.toml depends on `kernel` with `features = ["sched-mock"]`, gated to `cfg(not(target_os = "none"))`.
- [ ] `simulator/src/lib.rs` exposes `Simulator`, `SimulatorConfig`, `Seed`, `Trace` types as empty stubs.
- [ ] `simulator/src/bin/replay.rs` — CLI stub accepting `--seed <u64>` and `--trace-out <path>`; prints "unimplemented" and exits 0.
- [ ] Custom panic hook prints `SIMULATOR PANIC seed=<u64> tick=<u64>` then re-raises (test only).
- [ ] `cargo build -p simulator` and `cargo test -p simulator --no-run` succeed on the host.

## Context
- RFC: `docs/RFC/0006-host-dst-simulator.md`, merged PR #712.
- Depends on: #714 (host-target buildability of `kernel/`).

## Acceptance
- Workspace builds on host with the new crate.
- Replay binary runs and exits cleanly.
- Kernel-image build is unaffected (panic hook is `cfg(not(target_os = "none"))`).

PR #771, opened 2026-04-29
DST Phase 2 (3): Simulator + linear tick-by-tick run loop (#716)

### LINKED ISSUE #716: DST Phase 2 (3): Simulator + linear tick-by-tick run loop
Implements RFC 0006 — drives the kernel scheduler one tick at a time from a seeded clock and timer-IRQ source.

## Work
- [ ] `Simulator::new(seed: u64, config: SimulatorConfig)` wires `MockClock` + `MockTimerIrq` (from `kernel::task::env`) into `task::env()` for the duration of the run.
- [ ] `Simulator::step()` advances the clock by one tick, drains expired wakeups, injects the timer IRQ, and lets the scheduler run until quiescence.
- [ ] `Simulator::run_until(predicate)` and `Simulator::run_for(ticks)` helpers.
- [ ] ChaCha8 `SimRng` keyed by seed; named sub-streams via splitmix64 (`rng_for("scheduler")`, `rng_for("faults")`, etc.).
- [ ] Forbid `std::collections::HashMap` and other RandomState-using types via `clippy::disallowed_types` in `simulator/`.
- [ ] Smoke test: 1000-tick run with two cooperating tasks completes deterministically; identical seed → identical task-completion order across two runs.

## Context
- RFC: `docs/RFC/0006-host-dst-simulator.md`.
- Depends on: #714, simulator skeleton (item 2).

## Acceptance
- Two consecutive `cargo test -p simulator` runs of the smoke test produce byte-identical output (captured via the trace recorder stub or stdout for now).

dburkart:
Closes #716

## Summary

Replaces the #715 stubs in `simulator/` with the real DST run loop
(RFC 0006 Phase 2):

- `Simulator::new(seed, config)` Box-leaks a `MockClock` + `MockTimerIrq`
  pair and installs them on the current thread via
  `task::env::install_sim_env`. Predecessor #715's panic hook is
  rewired to surface the live tick.
- `Simulator::step()` advances the mock clock one tick, injects a timer
  IRQ, drains expired wakeups via the `Clock` trait object (same path
  `kernel::task::preempt_tick()` takes on bare metal), then acks the
  IRQ. Every transition is recorded into `Trace` in deterministic order.
- `Simulator::run_for(ticks)` and `Simulator::run_until(predicate)` are
  the v1 driver-loop entry points. `run_until` checks the predicate
  before each step and honours `SimulatorConfig::max_ticks` as a hard
  cap (returns `None` on exhaustion).
- `SimRng` wraps `ChaCha8Rng`. `rng_for(name)` derives named
  sub-streams via `splitmix64` over the master seed XOR an in-tree
  FNV-1a hash of the name — so adding a new sub-stream cannot perturb
  the bytes any existing one emits (RFC 0006 §"Reproduction
  commitments"). `rand_chacha` and `rand_core` are pulled with
  `default-features = false` so `getrandom` cannot leak OS entropy.
- `Trace` carries `Vec<TraceRecord>` with a small additive `Event`
  enum (`ClockAdvanced`, `TimerInjected`, `TimerAcked`, `TaskWoken`).
  Schema stays minimal so the determinism-equality assertion in the
  smoke test does not need to be revisited every time #717 grows the
  recorder.
- Crate-level `#![deny(clippy::disallowed_types)]` defends in depth on
  top of the workspace `clippy.toml` `HashMap`/`HashSet` ban.

### Note on `preempt_tick()`

RFC 0006 §"The driver loop" step 4 sketches calling
`kernel::task::preempt_tick()`. That function is gated to
`cfg(target_os = "none")` (FPU, swapgs, IDT, hand-written switch) and
cannot link on the host triple. The simulator stays at the seam — the
surface every Phase 2 v1 flake actually manifests on (timer-IRQ
ordering, deadline drain ordering, wakeup-before-deadline races).
In-kernel `sched-mock`-gated integration tests (a separate Phase 2
issue) will cover the rotation/softirq paths against the real
`preempt_tick`.

## Test plan

- [x] `cargo test -p simulator` — 15 simulator + 14 replay tests pass
- [x] Smoke test: 1000-tick run with two periodic wakeups (periods 7
  and 11) produces byte-identical traces across two independent
  `Simulator` instances built off the same seed
- [x] Two consecutive `cargo test -p simulator` runs produce
  byte-identical output (modulo cargo's own elapsed-time line) — the
  acceptance gate phrased in the issue
- [x] `cargo build -p vibix --lib --target x86_64-unknown-linux-gnu --features sched-mock` clean
- [x] `cargo xtask build` — kernel still builds
- [x] `cargo clippy -p simulator --all-targets` clean (zero warnings
  inside `simulator/`; pre-existing clippy warnings in
  `kernel/build.rs` are unchanged)
- [x] CI `host build (sched-mock)` job extended with two back-to-back
  `cargo test -p simulator` invocations as the merge-gate

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,6 +172,22 @@ jobs:
             --target x86_64-unknown-linux-gnu \
             --features sched-mock
 
+      # RFC 0006 / issue #716: host-side DST simulator run loop.
+      # The smoke test in `simulator/src/lib.rs` runs a 1000-tick
+      # two-task scenario and asserts byte-identical traces from two
+      # `Simulator` instances built off the same seed. Running it
+      # twice back-to-back here is the merge-gate phrasing from the
+      # issue ("Two consecutive `cargo test -p simulator` runs of
+      # the smoke test produce byte-identical output") — a
+      # determinism regression in the run loop, the seam, the
+      # `BTreeMap` drain order, or the ChaCha8 sub-stream derivation
+      # would surface as a diff between the two invocations rather
+      # than a flaky failure on one of them.
+      - name: cargo test -p simulator (deterministic — first pass)
+        run: cargo test -p simulator
+      - name: cargo test -p simulator (deterministic — second pass)
+        run: cargo test -p simulator
+
   unit-tests:
     name: host unit tests
     runs-on: ubuntu-latest
--- a/simulator/Cargo.toml
+++ b/simulator/Cargo.toml
@@ -32,3 +32,15 @@ path = "src/bin/replay.rs"
 # `x86_64-unknown-linux-gnu`, which matches the gate.
 [target.'cfg(not(target_os = "none"))'.dependencies]
 vibix = { path = "../kernel", features = ["sched-mock"] }
+# ChaCha8 is the simulator's master PRNG (RFC 0006 §RNG): seeded from a
+# u64 master seed, named sub-streams derived via `splitmix64`. We pull
+# `rand_chacha` *without* default features so that nothing along the
+# transitive dep chain reaches `getrandom` / OS entropy — every byte
+# the simulator produces must derive from the master seed alone, or
+# the determinism contract (RFC 0006 §"Reproducibility Envelope")
+# silently breaks.
+rand_chacha = { version = "0.10", default-features = false }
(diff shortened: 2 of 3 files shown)


dburkart:
@coderabbitai review

Issue #722, opened 2026-04-28
DST Phase 2 (8): Invariants + liveness + proptest-state-machine integration


dburkart:
Implements RFC 0006 — invariant checks over the trace, plus a proptest-state-machine driver so seed sweeps actually find new bugs instead of just replaying old ones.

## Work
- [ ] Safety invariants checked after each `step()`:
  - At most one task in `Running` per CPU (single-CPU until SMP RFC).
  - PID table monotonic: pids never reused before `wait()`.
  - Every `WakeupEnqueued` either fires or is cancelled by task exit; never stranded.
  - No task transitions `Blocked → Running` without an intervening `WakeupFired` or signal.
- [ ] Liveness invariants checked at run end:
  - Every `Runnable` task eventually runs within `liveness_window_ticks`.
  - No fork without a matching exit-or-wait within the run.
- [ ] `proptest_state_machine::ReferenceStateMachine` impl wired to `Simulator`; transitions: `Fork`, `Exec`, `Wait`, `Block`, `Wake`, `Yield`, `InjectFault`.
- [ ] Master seed flows from proptest into `Simulator::new(seed)` so shrinking respects the determinism envelope.

## Context
- RFC: `docs/RFC/0006-host-dst-simulator.md` §"Reference state machine: invariants over the trace, not refinement" and §"RNG stream coupling".
- Depends on: items 3, 4.

## Acceptance
- Failing invariant produces a shrunk seed + FaultPlan, not a giant random transition log.
- Existing kernel tests don't regress.

Issue #718, opened 2026-04-28
DST Phase 2 (5): sched_mock_trace! macro + scheduler introspection helpers


dburkart:
Implements RFC 0006 — adds compile-out emit points so the kernel feeds events to the simulator without paying any cost in release builds.

## Work
- [ ] `kernel/src/task/trace.rs` defines `sched_mock_trace!(event_expr)` which expands to a no-op when `feature = "sched-mock"` is off, and to a call into a thread-local trace sink when on.
- [ ] Wire emit points: `schedule()` entry/exit, `block_on()`, `wake_one()`, `task_exit()`, fork/exec/wait transitions, syscall entry/exit, page-fault entry.
- [ ] `nm-check` integration test (#669): assert all `sched_mock_*` symbols absent from the release kernel image.
- [ ] Helpers: `Simulator::current_task()`, `Simulator::runqueue_snapshot()`, `Simulator::pid_table_snapshot()` — read-only reflection used by invariants.

## Context
- RFC: `docs/RFC/0006-host-dst-simulator.md` §"Event emit points".
- Regex in RFC enumerates the full emit-point list — implementations must match.

## Acceptance
- `cargo build -p kernel --release` produces a binary with zero `sched_mock_*` symbols (verified by `nm --demangle=rust`).
- All emit points covered by a unit test that asserts the event lands in the trace.

PR #774, opened 2026-04-29
simulator: full Event enum + JSON dump + Trace::diff (#717)

### LINKED ISSUE #717: DST Phase 2 (4): (Tick, Event) trace recorder + JSON dump

dburkart:
Closes #717.

DST Phase 2 (4): land the full RFC 0006 §"State-machine model interface" event surface and the stable JSON schema downstream consumers (replay binary, CI artefact viewers, future proptest harness) parse against.

## Summary

- **Event enum** rewritten to the RFC-spec names. Existing variants from #716 renamed (semantics preserved):
  - `ClockAdvanced` → `TickAdvance`
  - `TimerAcked` → `TimerIrqAcked`
  - `TaskWoken` → `WakeupFired`
- **New variants** defined here, emit points land in follow-ups: `WakeupEnqueued`, `TaskScheduled`, `TaskBlocked`, `Syscall`, `Fault`, `FaultInjected`. Marked `#[allow(dead_code)]` until #718 / #722 wires their emit points.
- `Event` is `#[non_exhaustive]` so future variant additions are non-breaking at the type level (RFC 0006 UserSpace-A1 resolution); the JSON `schema_version` still bumps.
- `Trace::to_json` / `to_json_string` — stable schema, fixed field order so two equal traces produce byte-identical JSON. `schema_version = 1`.
- `Trace::from_json` — hand-rolled strict parser. **No `serde_json`** to keep the simulator's zero-OS-entropy transitive-dep envelope (RFC 0006 §"Determinism envelope").
- `Trace::diff(&other) -> Option<usize>` — first divergent index, for replay-equivalence assertions.
- `Trace::with_capacity_limit(N)` — bounded recorder with head-eviction and one-shot stderr warning.
- New doc: `simulator/docs/trace-schema.md` — canonical schema reference.

## Test plan

- [x] `cargo test -p simulator --target x86_64-unknown-linux-gnu` — 29 lib tests + 14 replay-bin tests pass.
  - `trace::tests::round_trip_preserves_every_variant` — every variant in the schema (including the not-yet-emitted ones) round-trips.
  - `trace::tests::diff_*` — first divergent index, equal traces, length-difference cases.
  - `trace::tests::capacity_limit_evicts_oldest` — bounded trace eviction.
  - `trace::tests::parse_rejects_*` — schema_version mismatch, unknown event type, unknown block reason, trailing bytes, string escapes.
  - `tests::smoke_recorded_trace_round_trips_through_json` — 200-tick simulator run with two cooperating tasks; record → JSON → parse → re-serialize is byte-identical.
  - `tests::smoke_1000_ticks_with_two_tasks_is_deterministic` — pre-existing #716 determinism smoke continues to pass with the renamed variants.
- [x] `cargo clippy -p simulator --target x86_64-unknown-linux-gnu --all-targets` clean (`#![deny(clippy::disallowed_types)]` enforced; no `HashMap`/`HashSet` introduced).
- [x] `cargo xtask build` — kernel image still builds (simulator is `cfg(not(target_os = "none"))`-gated, but verifying the gate didn't regress).

## Out of scope

- Emit points for `WakeupEnqueued`/`TaskScheduled`/`TaskBlocked`/`Syscall`/`Fault` are #718's `sched_mock_trace!` macro work.
- `FaultInjected` is #722's FaultPlan landing.
- The `replay` binary still prints `unimplemented` on `--trace-out`; wiring it to actually run a fixed-length sim and dump the JSON is a follow-up — the library API is the merge gate for #717.

### DIFF
--- a/simulator/docs/trace-schema.md
+++ b/simulator/docs/trace-schema.md
@@ -212,7 +212,7 @@ The parser in `simulator::trace::Trace::from_json` enforces:
 
 For every trace `T`:
 
-```
+```rust
 let json = T.to_json_string();
 let T2 = Trace::from_json(&json).unwrap();
 assert_eq!(T.records(), T2.records());
--- a/simulator/docs/trace-schema.md
+++ b/simulator/docs/trace-schema.md
@@ -1,0 +1,237 @@
+# Simulator trace JSON schema
+
+This document is the stable contract for the JSON output of
+`simulator::Trace::to_json` (RFC 0006, issue #717). Downstream
+consumers — the future `replay` binary, CI artefact viewers,
+property-test harnesses written against frozen failing seeds, and the
+`auto-engineer` review tooling — parse against this schema. **Adding,
+removing, or renaming any variant or field below is a `schema_version`
+bump**, and a bump is a breaking change for those consumers.
+
+## Top-level shape
+
+```json
+{
+  "schema_version": 1,
+  "records": [
+    { "tick": 0, "event": { "type": "tick_advance", "from": 0, "to": 1 } }
+  ]
+}
+```
+
+- `schema_version` is the integer version of this schema. The current
+  version is **1**. The parser refuses to deserialize an object whose
+  `schema_version` does not match the version it was compiled against.
+- `records` is a JSON array of `(tick, event)` pairs in emission order.
+  Emission order is the order the simulator's run loop produced them
+  on the recording thread. Reorder is not legal.
+
+## Record shape
+
+Every record is a JSON object with exactly two fields, in this order:
+
+| field   | type     | meaning                                                |
+| ------- | -------- | ------------------------------------------------------ |
+| `tick`  | `u64`    | Value of `MockClock::now()` when the event was emitted |
+| `event` | `object` | The event itself (variant tag in `type`)               |
+
+Field ordering inside every object is fixed (`tick` then `event`); the
+encoder emits no whitespace beyond a single space after each `:` and
+`,`. This is what makes the round-trip property
+*record → JSON → parse → re-record → byte-identical JSON* hold.
+
+## Event variants
+
+All events use a discriminator-tag layout — a `type` field whose
+string value selects the variant — and a fixed field order matching
+the corresponding Rust enum variant declaration order in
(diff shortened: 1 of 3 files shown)


dburkart:
Thanks for the review.

- **trace-schema.md fence language** — fixed in 842e179.
- **VecDeque for O(1) head eviction** — deferring. The capacity-bounded path is opt-in (`Trace::with_capacity_limit`), the warning fires once and tells the operator to raise the cap, and the call-site contract today is `Trace::records() -> &[TraceRecord]` (consumed via slice equality in `assert_eq!`, `.to_vec()`, etc.). Switching storage to `VecDeque` would either force `&mut self` on `records()` (via `make_contiguous`) or change the return type to an iterator/`(&[T], &[T])` pair — both are larger API churn than this nit warrants. Filing as a follow-up so when an actual bounded-trace heavy-eviction workload appears (#722 FaultPlan sweeps are the candidate), we can switch the storage with a corresponding API decision in the same PR.

Issue #755, opened 2026-04-29
block: extend writeback daemon to walk inode page caches


dburkart:
## Motivation

Implements RFC 0007. The per-mount writeback daemon now walks every superblock-mounted inode's `mapping` and calls `writepage` on every dirty pgoff (snapshot-collect under cache mutex, write outside).

## Work

- Daemon iterates inodes after the existing buffer-cache flush.
- Snapshot-then-writepage discipline: collect dirty pgoffs under `cache.inner`, drop the lock, perform writepage with the snapshot.
- Skip-on-shutdown: stop iterating if `SbActiveGuard` reports the SB is draining.
- Test: many-inode workload, daemon makes progress on all.

## Context

- RFC 0007 §MAP_SHARED writeback, §Workstream D — `docs/RFC/0007-page-cache-file-mmap.md`
- Parent epic: #734

dburkart:
Renumbered: this RFC is now 0007 (was 0006, collision with #712 DST simulator). See https://github.com/dburkart/vibix/pull/765.

Issue #745, opened 2026-04-29
vfs: extend Inode with mapping: Option<Arc<PageCache>>


dburkart:
## Motivation

Implements RFC 0007. The `Inode` needs a slot for the per-inode `PageCache`. Construct lazily on first mmap or first read via the page cache. Captures `Arc<dyn AddressSpaceOps>` from the inode at cache-construction time (inode-binding rule).

## Work

- Add `mapping: BlockingRwLock<Option<Arc<PageCache>>>` to `Inode` (gated by `#[cfg(feature = "page_cache")]` for the migration window).
- Helper `Inode::page_cache_or_create(&self) -> Arc<PageCache>`.
- Drop hook: dropping the inode drops the cache (writeback daemon must release strong refs first).

## Context

- RFC 0007 §Inode-binding rule — `docs/RFC/0007-page-cache-file-mmap.md`
- Parent epic: #734

dburkart:
Renumbered: this RFC is now 0007 (was 0006, collision with #712 DST simulator). See https://github.com/dburkart/vibix/pull/765.

PR #775, opened 2026-04-29
sys: implement fsync(2) / fdatasync(2) syscalls (#748)

### LINKED ISSUE #748: sys: extend fsync/fdatasync to flush the page cache + errseq EIO
## Motivation

Implements RFC 0007. `fsync` flushes `inode.mapping.dirty` (calling `writepage`) before fencing on `BlockCache::sync_fs`. `fdatasync` skips the inode-table flush when no allocation occurred (allocation-required flushes are issued *inside* `writepage` so they are already on stable storage).

## Work

- Wire `FileOps::fsync` through the page cache's dirty list.
- Two-stage flush: page cache first, then `BlockCache::sync_fs(sb_dev)`.
- `OpenFile` snapshots `mapping.wb_err` at `open`; `fsync` re-reads and returns `EIO` if it advanced.
- `fdatasync` (`data_only=true`) skips the inode-table flush per RFC.
- Tests: dirty-page-flush, errseq EIO surfacing.

## Context

- RFC 0007 §Ordering vs fsync/fdatasync — `docs/RFC/0007-page-cache-file-mmap.md`
- Parent epic: #734

dburkart:
Closes #748

## Summary

- Adds `SYS_FSYNC` (74) and `SYS_FDATASYNC` (75) dispatch arms wired through to a new `OpenFile::do_fsync(data_only)` helper that performs the RFC 0007 §Ordering vs fsync/fdatasync two-stage flush: `FileOps::fsync` driver hook, then `SuperOps::sync_fs` to fence on the per-mount `BlockCache`.
- Adds `OpenFile::wb_err_snapshot: AtomicU32`, captured at `open(2)` and re-read on every `fsync` to surface a sticky `EIO` via the errseq pattern (RFC 0007 §`wb_err`). On consume, the snapshot is caught up so only *new* errors are surfaced (Linux semantics).
- Adds `Inode::wb_err()` as a forward-compat seam returning 0 today; once #745 wires `Inode::mapping`, this becomes a one-liner reading the page-cache's `wb_err` and the syscall plumbing requires no further change.

## Notes / scope

- The split between `fsync` and `fdatasync` at the SB layer (skip the inode-table flush for `fdatasync`) is **not** implemented here — `SuperOps::sync_fs` does not yet have a `data_only` variant. Calling the full `sync_fs` for both is a POSIX-compatible superset (fdatasync is permitted to flush more than necessary). A follow-up issue tracks adding the `sync_fs(data_only)` seam alongside #755.
- The page-cache writeback step in `do_fsync` is currently a no-op because no inode has a `mapping` yet (#745). When #745 lands, the writeback walk is added inside `do_fsync` before the `sync_fs` call (or via #755's daemon stub).

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo xtask build`
- [x] `cargo xtask test` — host unit + QEMU integration; new `syscall_fsync` integration test exercises the dispatch arms end-to-end (happy path, repeated-clean, bad-fd EBADF, after-close EBADF).
- [x] `cargo xtask smoke`
- [ ] CI gates: fmt+build, host unit, integration QEMU, smoke, pjdfstest, ext2 rootfs determinism, RFC 0004 vfs_creds, CodeRabbit

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -501,3 +501,7 @@ required-features = ["ext2"]
 [[test]]
 name = "block_cache_stress"
 harness = false
+
+[[test]]
+name = "syscall_fsync"
+harness = false
--- a/kernel/src/arch/x86_64/syscall.rs
+++ b/kernel/src/arch/x86_64/syscall.rs
@@ -606,6 +606,16 @@ pub unsafe extern "C" fn syscall_dispatch(
             }
         }
 
+        // fsync(fd) — flush page cache + per-mount BlockCache::sync_fs;
+        // surface sticky EIO via the per-OpenFile errseq snapshot. RFC
+        // 0007 §Ordering vs fsync/fdatasync.
+        FSYNC => super::syscalls::vfs::sys_fsync_impl(a0 as u32, false),
+
+        // fdatasync(fd) — same data flush as fsync, may skip the
+        // inode-table flush per Linux semantics. See sys_fsync_impl
+        // for the data_only-vs-fsync split.
+        FDATASYNC => super::syscalls::vfs::sys_fsync_impl(a0 as u32, true),
+
         // dup2(oldfd, newfd)
         DUP2 => {
             let oldfd = a0 as i32;
@@ -1760,6 +1770,8 @@ pub mod syscall_nr {
     pub const DUP2: u64 = 33;
     pub const DUP3: u64 = 292;
     pub const FCNTL: u64 = 72;
+    pub const FSYNC: u64 = 74;
+    pub const FDATASYNC: u64 = 75;
     pub const FSTAT: u64 = 5;
     pub const STAT: u64 = 4;
     pub const LSTAT: u64 = 6;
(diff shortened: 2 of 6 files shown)


dburkart:
Re: CodeRabbit's "OpenFile struct literals in test code are missing required `wb_err_snapshot` field" — false positive, deferring.

Both flagged literals in `kernel/src/fs/vfs/tarfs.rs` (lines 826, 879) sit inside `#[cfg(test)]` modules under `pub mod vfs`, which is itself gated `#[cfg(target_os = "none")]`. Host builds skip the `target_os = "none"` arm, and the kernel's target build (`x86_64-unknown-none`) does not enable `cfg(test)` — so this code never gets compiled in any configuration we ship. CI confirms: `fmt + build (release)`, `host unit tests`, `host build (sched-mock)`, and all four QEMU integration shards are green on this PR.

Filing a follow-up to clean up the dead `cfg(test)` bodies in tarfs.rs (out of scope for the fsync work — they predate this PR).

PR #772, opened 2026-04-29
mem: introduce PageCache, CachePage, PG_* state bits with split locks

dburkart:
Closes #736

## Summary

- Lands the foundation data structures for the RFC 0007 page cache (parent epic #734, Workstream A wave 1): `CachePage` with `PG_*` state bits, per-inode `PageCache` with split-lock discipline, install-race protocol, and filler-error rollback.
- Strict implementation of RFC 0007 §State-bit ordering: `publish_uptodate_and_unlock` sets `PG_UPTODATE` (Release) before clearing `PG_LOCKED` (Release); `Acquire` loads on the observer side make page bytes visible. Filler-error rollback removes the stub from the index *before* clearing `PG_LOCKED`, so the `PG_LOCKED`-clear / `PG_UPTODATE`-clear combination is never observable from the cache index.
- `CachePage` is `#[repr(align(64))]` (RFC normative invariant against false-sharing). `PageCache::inner` is a `BlockingMutex<PageCacheInner>` (level 4 per RFC §Lock-order ladder); the index mutex is acquired only long enough to clone the `Arc<CachePage>`, never held across blocking I/O. `wb_err: AtomicU32` is the errseq counter for sticky EIO.
- Out of scope (deferred to siblings, per the issue): `AddressSpaceOps` trait (#737), `FileObject` (#738), page-fault path (#739), eviction + `writeback_complete_wq` (#740), readahead + `ra_state` (#741).
- Host-buildability shim: `crate::sync` is `target_os = "none"`-gated, so under `cfg(test)` we substitute a small `host_stub` (`spin::Mutex` + no-op `WaitQueue`). The install-race unit tests never cross-thread block on `PG_LOCKED`, so the no-op shape is faithful for unit-test coverage.

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo xtask build` — kernel target compiles (release + dev)
- [x] `cargo test --lib mem::page_cache` — 19 new tests, all green
- [x] `cargo test --lib` — full host suite green (563 passed; 0 failed)
- [x] CI gates: fmt + build, host unit, integration QEMU, smoke, pjdfstest, ext2 rootfs determinism, RFC 0004 vfs_creds gate, CodeRabbit

The new tests cover (host unit, `kernel/src/mem/page_cache.rs`):

- `cachepage_is_64_byte_aligned` — RFC `#[repr(align(64))]` invariant.
- `fresh_stub_is_locked_and_not_uptodate` — install-time state.
[...]

Issue #749, opened 2026-04-29
fs/ext2: implement AddressSpaceOps::readpage with sparse + tail zero


dburkart:
## Motivation

Implements RFC 0007. ext2's `readpage` walks indirect blocks, `bread`s through the buffer cache, memcpys 4 KiB into the page, zero-fills the `[i_size .. page_end)` tail and zero-fills sparse holes (indirect-block pointer == 0).

## Work

- Walk direct/single/double/triple indirect via the existing RFC 0004 walker.
- `bread` each block, memcpy into the cache page buffer.
- Tail zero past `i_size`.
- Sparse-hole zero-fill for any pointer == 0.
- Faithful errno propagation on Err so the page-cache filler-error contract triggers cleanly.
- Tests: dense file readpage, sparse-hole readpage, tail-page zero.

## Context

- RFC 0007 §Tail-page zeroing, §Errno table sparse-hole row — `docs/RFC/0007-page-cache-file-mmap.md`
- Parent epic: #734

dburkart:
Renumbered: this RFC is now 0007 (was 0006, collision with #712 DST simulator). See https://github.com/dburkart/vibix/pull/765.

Issue #742, opened 2026-04-29
mem: refcount-discipline + lockdep tests for page cache


dburkart:
## Motivation

Implements RFC 0007. Verify the `Arc::strong_count(CachePage)` vs `mem::refcount::get(phys)` split is exercised correctly: a cache page held only by an installed PTE must not be evicted; a cache page held by an in-flight `Arc::clone` must not be evicted; the cache's own reference is the only one safely droppable by eviction.

## Work

- Host unit tests covering the eviction-gate matrix.
- `assert_no_spinlocks_held` regression tests.
- Lock-order assertions matching RFC §Lock-order ladder.

## Context

- RFC 0007 §Refcount discipline, §Lock-order — `docs/RFC/0007-page-cache-file-mmap.md`
- Parent epic: #734

dburkart:
Renumbered: this RFC is now 0007 (was 0006, collision with #712 DST simulator). See https://github.com/dburkart/vibix/pull/765.

PR #778, opened 2026-04-29
mem: AddressSpaceOps trait + Arc<dyn AddressSpaceOps> on PageCache (#737)

### LINKED ISSUE #737: mem: AddressSpaceOps trait (readpage/writepage/readahead/truncate_below)
## Motivation

Implements RFC 0007. Per-FS hook interface that the page cache calls on miss/writeback. Default no-op for `readahead` and `truncate_below` so non-block filesystems (ramfs, tarfs) can opt out.

## Work

- Trait `AddressSpaceOps: Send + Sync` with the four methods.
- `debug_lockdep::assert_no_spinlocks_held()` at the entry of every method (runtime, not debug-only).
- `Arc<dyn AddressSpaceOps>` storage on `PageCache`.
- Stub impl in tests so the host unit tests can exercise the page cache without a real FS.

## Context

- RFC 0007 §AddressSpaceOps — `docs/RFC/0007-page-cache-file-mmap.md`
- Parent epic: #734

dburkart:
## Summary

Closes #737

Implements RFC 0007 §AddressSpaceOps. Defines the per-FS hook trait the page cache calls on miss / writeback / readahead / truncate, and threads `Arc<dyn AddressSpaceOps>` into `PageCache` (the deferred follow-up #736 / PR #772 left for this issue).

- `kernel/src/mem/aops.rs`: `pub trait AddressSpaceOps: Send + Sync` with required `readpage` / `writepage` and default-no-op `readahead` / `truncate_below`. Each method body must call `debug_lockdep::assert_no_spinlocks_held` as the first instruction (RFC 0007 §Lock-order ladder); the default bodies already do.
- `kernel/src/mem/page_cache.rs`: `PageCache` gains `ops: Arc<dyn AddressSpaceOps>` captured at construction and never rebound (RFC 0007 §Inode-binding rule). `PageCache::new` takes the ops Arc; `PageCache::ops()` clones it back out for the writeback daemon / fault path. The cache does not yet *invoke* `ops.readpage` / `ops.writepage` — that wiring is #739 / #740 / #742.
- Host unit tests cover dyn dispatch through Arc, EOF/short-read semantics, writepage round-trip through readpage, default no-op bodies, Send/Sync, and the new `PageCache::ops()` accessor identity.

Out of scope (sibling issues): concrete `AddressSpaceOps` impls (ext2: #749/#750/#751/#752), `FileObject` (#738), eviction (#740), readahead heuristic (#741), `Inode.mapping` plumbing (#745), `fsync` integration (#748).

Deferred follow-up: lifting `debug_lockdep::assert_no_spinlocks_held` from `cfg(debug_assertions)`-gated to unconditional-runtime (RFC 0007 §Lock-order ladder calls for runtime, not debug-only). Tracked separately so the counter migration doesn't gate this trait landing.

## Test plan

- [x] `cargo xtask build` — green
- [x] `cargo xtask test` — 576 host unit tests pass + integration QEMU tests pass
- [x] `cargo xtask smoke` — all 42 markers present
- [x] `cargo fmt --all` — clean

### DIFF
--- a/kernel/src/mem/aops.rs
+++ b/kernel/src/mem/aops.rs
@@ -241,8 +241,10 @@ mod tests {
                 page.copy_from_slice(buf);
                 Ok(())
             } else {
-                // Out-of-range writepage: emulate -ENOSPC.
-                Err(28)
+                // Out-of-range writepage: emulate ENOSPC. vibix's
+                // errno convention is negative i64 (see
+                // `kernel/src/fs/mod.rs` — `pub const ENOSPC: i64 = -28`).
+                Err(crate::fs::ENOSPC)
             }
         }
 
@@ -295,8 +297,10 @@ mod tests {
     fn writepage_out_of_range_returns_errno() {
         let ops: Arc<dyn AddressSpaceOps> = MemoryBackedOps::with_pages(1);
         let payload = [0u8; 4096];
-        // pgoff 99 is past the backing store.
-        assert_eq!(ops.writepage(99, &payload), Err(28));
+        // pgoff 99 is past the backing store. ENOSPC = -28; vibix's
+        // errno convention is negative i64.
+        assert_eq!(ops.writepage(99, &payload), Err(crate::fs::ENOSPC));
+        assert_eq!(crate::fs::ENOSPC, -28);
     }
 
     /// Default `readahead` is a no-op: it does not panic, does not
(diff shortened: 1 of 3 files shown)


dburkart:
Reviewed the latest CodeRabbit finding (test asserts `buf[0] == 2` without writing first). The finding is incorrect — `MemoryBackedOps::with_pages` already initializes the first byte of each page to its index (`buf[0] = i as u8;` in `kernel/src/mem/aops.rs`, the `with_pages` constructor). The test asserts that `readpage(2)` returns a buffer whose `buf[0]` equals `2`, which is exactly the marker the constructor wrote. No fix needed; proceeding to merge.

Issue #753, opened 2026-04-29
fs/ext2: implement FileOps::mmap returning Arc<FileObject>


dburkart:
## Motivation

Implements RFC 0007. ext2-specific `mmap` impl that lazily constructs the inode's `PageCache` and returns a `FileObject` with the open-mode snapshot.

## Work

- `FileOps::mmap` impl on `Ext2Inode`.
- Lazy `mapping` construction with the ext2-specific `AddressSpaceOps`.
- Snapshot `OpenFile.f_mode` into `FileObject.open_mode`.
- Test: smoke mmap + read of an ext2-backed file.

## Context

- RFC 0007 §FileOps::mmap — `docs/RFC/0007-page-cache-file-mmap.md`
- Parent epic: #734

dburkart:
Renumbered: this RFC is now 0007 (was 0006, collision with #712 DST simulator). See https://github.com/dburkart/vibix/pull/765.

PR #803, opened 2026-05-02
fs/ext2: writepage with allocate-on-extend (#750)

### LINKED ISSUE #750: fs/ext2: implement AddressSpaceOps::writepage with allocate-on-extend
## Motivation

Implements RFC 0007. ext2's `writepage` splits the page back into block-sized fragments and writes through the buffer cache. **For unallocated underlying blocks**, calls into the bitmap allocator and follows the normative bitmap → indirect → data ordering (matches RFC 0004 §create).

## Work

- Per-block split + `mark_dirty` + `sync_dirty_buffer`.
- Allocation path: bitmap → indirect-block update → data block, each `sync_dirty_buffer`d in order.
- `i_blocks` and `i_size` updated only after the data write succeeds.
- Rollback on per-block failure (return block to bitmap).
- Per-page-write failure increments `mapping.wb_err` for the errseq surface.
- Tests: dense writepage, sparse-then-extend writepage, allocator-failure rollback.

## Context

- RFC 0007 §writepage block allocation — `docs/RFC/0007-page-cache-file-mmap.md`
- Parent epic: #734

dburkart:
Closes #750.

## Summary
- Replaces the EROFS stub in `Ext2Aops::writepage` with a real buffer-cache writeback path (RFC 0007 §AddressSpaceOps, RFC 0004 §Write Ordering): splits each 4 KiB page into FS-block-sized fragments, drives each through bitmap → indirect → data allocation when the underlying block is sparse, and synchronously flushes via `mark_dirty` + `sync_dirty_buffer`.
- `i_blocks`, `i_size`, mtime, and ctime are only mutated after every fragment write succeeds; the inode-slot flush is the last on-disk write, so a crash leaves readers with the pre-call (correct, just stale) `i_size` rather than an `i_size` that claims un-flushed blocks.
- On any per-block failure (ENOSPC, on-disk pointer forgery detected, buffer-cache I/O), walks the recorded allocation events in reverse: zeroes the parent slot (in `i_block[]` or in a previously-allocated indirect block), then `free_block`s the underlying disk block. Bitmap, free-blocks counter, and inode metadata end identical to their pre-call state.
- `validate_existing_pointer` and `flush_inode_slot` in `fs/ext2/file.rs` promoted to `pub(super)` so writepage reuses the same forgery / RMW discipline rather than duplicating it.

## Tests
New integration test `kernel/tests/ext2_writepage.rs`:
- `writepage_dense_first_page` — pgoff 0 of a fresh inode; asserts all four direct slots are allocated, `i_size`/`i_blocks` reflect the full page, and bytes round-trip through `read_file_at`.
- `writepage_sparse_then_extend` — pgoff 5 of a fresh inode; asserts holes [0, 5) read POSIX-zero, the live page round-trips, the single-indirect pointer block is consumed alongside the four data blocks (5 blocks total).
- `writepage_allocator_failure_rolls_back` — two sub-tests: (a) drain bitmap to zero, expect ENOSPC and unchanged metadata; (b) drain to ≤2 free blocks so allocation fails mid-page, expect rollback to free every fragment that was allocated and the free-blocks counter to be restored.
- `writepage_ro_mount_returns_erofs` — RO mount short-circuits before any allocator touch.

## Test plan
- [x] `cargo fmt --all -- --check` (clean)
- [x] `cargo xtask build` (clean)
- [x] `cargo xtask smoke` — all 42 markers present
- [x] `cargo test --lib` — 604 host unit tests pass
- [x] `cargo test --test ext2_writepage` — 4/4 pass
- [x] Re-ran every ext2 integration test (`ext2_readpage`, `ext2_file_write`, `ext2_file_read`, `ext2_block_alloc`, `ext2_setattr`, `ext2_create`, `ext2_dir_ops`, `ext2_unlink`, `ext2_link`, `ext2_orphan_*`, `ext2_rename`, `ext2_ialloc`, `ext2_validate_bgdt`, `ext2_mount`, `ext2_inode_iget`, `ext2_malformed_dir`, `ext2_enospc`, `ext2_crash_consistency`) — all pass.
- [x] Pre-existing flake on `main`: `blocking_sync::rwlock_concurrent_readers` (`expected 3 readers, got 2`) reproduces on `1f7a4dc` without these changes; not caused by this PR.

## Out of scope / follow-ups
- The VFS-layer `Inode::meta` mirror is intentionally not updated by writepage — `Ext2Aops` doesn't carry an `Inode` weak pointer (readpage didn't need one) and adding it would expand the trait-object footprint #753 hasn't yet committed to. The on-disk slot is flushed, so the next `iget` re-reads the fresh values; a concurrent in-memory `getattr` against the same inode sees the previous size/blocks until then. The writeback daemon doesn't itself depend on the VFS-meta mirror.
- `mapping.wb_err` bumping on writepage failure is the writeback daemon's responsibility (PR #799) — this PR just returns a faithful errno that the daemon's `bump_wb_err` site keys off.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -423,6 +423,11 @@ harness = false
 required-features = ["ext2"]
 
 [[test]]
+name = "ext2_writepage"
+harness = false
+required-features = ["ext2"]
+
+[[test]]
 name = "ext2_orphan_chain"
 harness = false
 required-features = ["ext2"]
--- a/kernel/src/fs/ext2/aops.rs
+++ b/kernel/src/fs/ext2/aops.rs
@@ -3,18 +3,15 @@
 //! readahead / truncate.
 //!
 //! RFC 0007 §`AddressSpaceOps` and §Tail-page zeroing are the normative
-//! spec. This module is **Workstream C / issue #749**: it lands the
-//! `readpage` path only. The other three trait methods (`writepage`,
-//! `truncate_below`, `readahead`) are deliberately stubbed out here:
+//! spec. This module currently lands two of the four trait methods:
 //!
-//! - `writepage` returns `Err(EROFS)` — issue #750 lands the real
-//!   buffer-cache writeback. Until then a forced-RO behaviour is
-//!   strictly safer than `unimplemented!()`-panic, because a future
-//!   page-fault path that wires `MAP_SHARED` writeback (#746/#755) can
-//!   exercise the writepage call site without bringing the kernel down;
-//!   the daemon will see `EROFS` and bump `wb_err`, which is the
-//!   correct user-visible behaviour for "this filesystem doesn't
-//!   support writeback yet".
+//! - `readpage` (issue #749) — parsed sparse / dense / past-EOF logic
+//!   in [`Ext2Aops::readpage`].
+//! - `writepage` (issue #750) — split-the-page-into-blocks-and-
+//!   allocate-on-extend in [`Ext2Aops::writepage`].
+//!
+//! The remaining two methods are deliberately left at the trait default:
+//!
 //! - `truncate_below` is left at the trait default (no-op + spinlock
 //!   assert). Issue #751 will replace it.
 //! - `readahead` is left at the trait default (no-op + spinlock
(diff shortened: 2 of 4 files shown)


dburkart:
@coderabbitai review

dburkart:
Thanks for the thorough review. Pushed `a560f34` addressing two of the three findings:

- **Major (refactor)** — restored `meta.size`, `meta.i_blocks`, `meta.mtime`, `meta.ctime` on `flush_inode_slot` failure alongside the existing `i_block[]` rollback.
- **Nit (tests)** — dense + sparse tests now re-`iget` after writepage and assert against the freshly-decoded on-disk slot to prove durability.

**Deferred (Critical — durability discipline)**: I'm leaving the "treat `sync_dirty_buffer` failures as indeterminate / latch RO" finding for a separate change. The current "free + restore" posture matches what `write_file_at`, `setattr`, `link`, `create`, and `flush_inode_slot` all do today; this PR is consistent with the rest of the write path. Tightening the discipline meaningfully needs to touch all those call sites at once (and ideally lift `force_ro` into the buffer-cache sync surface itself), which is its own change. Filing as follow-up.

PR #890, opened 2026-05-05
Implement shell I/O redirection runtime

### LINKED ISSUE #879: Implement shell I/O redirections for /bin/sh
## Motivation
I/O redirections (`>`, `<`, `>>`, `2>`, `<<`, `>&`) are fundamental to POSIX shell usage — they enable scripts to manipulate file descriptors, redirect output to files, and construct here-documents. This component processes redirection nodes in the AST and applies them via `dup2`/`open`/`close` syscalls before command execution.

Follow-up of #874

## Work
- [ ] Implement output redirection: `>` (truncate), `>>` (append)
- [ ] Implement input redirection: `<`
- [ ] Implement fd-specific redirections: `N>`, `N<`, `N>>` (where N is a file descriptor number)
- [ ] Implement fd duplication: `>&N`, `<&N`, `>&-` (close fd)
- [ ] Implement stderr redirection shorthand: `2>`, `2>&1`
- [ ] Implement here-documents: `<<DELIM` ... `DELIM` (with and without tab stripping via `<<-`)
- [ ] Apply redirections in correct order per POSIX (left to right, after fork but before exec)
- [ ] Add host-side unit tests for redirection parsing and fd manipulation logic

## Context
- Parent issue: #874 (Implement /bin/sh)
- Depends on: shell parser (redirection syntax is parsed as part of simple commands)
- Uses `open`, `close`, `dup2` syscalls which are already available in vibix userspace
- POSIX.1-2024 §2.7 (Redirections) defines the conformance target

dburkart:
Closes #879

## Summary
- Add `base/sh/src/redirect.rs` implementing POSIX.1-2024 §2.7 I/O redirection runtime
- Supports all redirection operators: `>`, `>>`, `<`, `<>`, `N>`, `N<`, `N>>`, `>&N`, `<&N`, `>&-`, `<<` (here-docs), and `<<-` (tab-stripping here-docs)
- Redirections applied left-to-right per POSIX, with `SavedFds` for undoing redirections on builtins
- Runtime uses extern C declarations for `open`/`close`/`dup`/`dup2`/`pipe`/`write`/`fcntl` provided by vibix_libc through the std toolchain
- 37 host-side unit tests covering fd resolution, open flag computation, dup target parsing, heredoc tab stripping, ordering semantics, error display, and SavedFds

## Test plan
- [x] `cargo test` in `base/sh/` — 218 tests pass (37 new redirect tests)
- [x] `cargo xtask build` — clean build
- [x] `cargo xtask test` — passes (only pre-existing `rwlock_concurrent_readers` flake)

### DIFF
--- a/base/sh/src/main.rs
+++ b/base/sh/src/main.rs
@@ -3,6 +3,7 @@
 mod expand;
 mod lexer;
 mod parser;
+mod redirect;
 
 use lexer::{Lexer, Token};
 
--- a/base/sh/src/redirect.rs
+++ b/base/sh/src/redirect.rs
@@ -227,12 +227,17 @@ pub fn resolve_fd(redir: &Redirect) -> RawFd {
 }
 
 /// Strip leading tabs from each line of a here-document body (for the
-/// `<<-` variant).
+/// `<<-` variant). Preserves a trailing newline if the input has one.
 pub fn strip_heredoc_tabs(body: &str) -> String {
-    body.lines()
+    let mut result: String = body
+        .lines()
         .map(|line| line.trim_start_matches('\t'))
         .collect::<Vec<_>>()
-        .join("\n")
+        .join("\n");
+    if body.ends_with('\n') {
+        result.push('\n');
+    }
+    result
 }
 
 // ── Core redirect application (runtime, not compiled for tests) ──────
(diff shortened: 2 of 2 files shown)


dburkart:
Addressed the trailing-newline bug in `strip_heredoc_tabs` — good catch. The negative-fd nit (parsing `-1` as a valid fd) is deferred to future work; at runtime the kernel would reject it with EBADF, and the error message is clear enough for now.

PR #910, opened 2026-05-06
fs/ext2: harden readahead per CodeRabbit review

### LINKED ISSUE #809: fs/ext2: harden Ext2Aops::readahead per CodeRabbit review on PR #807
## Motivation

Discovered while merging #807 (issue #752). CodeRabbit's review on #807 surfaced two quality findings that arrived after the squash-merge had already landed `Ext2Aops::readahead` on `main`. The fixes are sitting on the closed `m752-ext2-readahead` branch (commit `883d1bf`) but never made it through review-and-merge to `main`. Capturing here so they don't get lost.

Both are minor quality improvements — the merged code is functionally correct, just slightly less defensive than the playbook says it should be.

## Work

1. **Stop prefetching on `bread` failure** (`kernel/src/fs/ext2/aops.rs`).
   The merged impl uses `let _ = super_ref.cache.bread(...)` inside `Ext2Aops::readahead`, so a transient device error is silently swallowed and the loop keeps walking the rest of the window. Per the impl's own doc-comment ("Any errno observed mid-walk … stops the prefetch") and the existing `WalkError::Io | Corrupt` arm that already returns, the `bread` Err arm should also `return`. Spending more `bread`s after a device failure just amplifies the I/O error and wastes buffer-cache slots.

   ```rust
   if super_ref
       .cache
       .bread(super_ref.device_id, abs as u64)
       .is_err()
   {
       return;
   }
   ```

2. **Assert sparse-hole readahead leaves residency unchanged** (`kernel/tests/ext2_readahead.rs`, `readahead_sparse_holes_dont_warm_phantom_blocks`).
   The merged test calls `aops.readahead(1, 1)` over the all-holes page but only checks the call returns. Capture `cache.len()` before / after and assert equality. By the time of the call the indirect block is already resident (from prior `resolve_block` probes plus `aops.readahead(0, 1)`), so an `Ok(None)`-only walk should add no new residency. This catches a regression where a future `Ok(None)` arm accidentally tries to `bread` a phantom block.

Both changes are small, isolated, and fully tested locally on `883d1bf`.

## Context

- Parent issue / merged PR: #752 / #807 (commit `86ae2e9` on main)
- CodeRabbit review on #807 (issue-level + 2 inline diff comments)
- Branch with the fix already drafted: `m752-ext2-readahead` (commit `883d1bf` — apply, rebase onto current main, open new PR)
- Files: `kernel/src/fs/ext2/aops.rs`, `kernel/tests/ext2_readahead.rs`

dburkart:
Closes #809

## Summary
- Stop the prefetch walk on `bread` failure in `Ext2Aops::readahead` instead of silently discarding the error with `let _ = ...`. A failing device is unlikely to succeed on the next block; the upcoming `readpage` surfaces the faithful errno on demand.
- Strengthen the `readahead_sparse_holes_dont_warm_phantom_blocks` integration test to capture `cache.len()` before and after the all-holes readahead call and assert equality, replacing the previous unused `_len_before` / no-assertion pattern.

## Test plan
- [x] `cargo xtask build` — compiles cleanly
- [x] `cargo xtask test` — all tests pass (only known-flaky `rwlock_concurrent_readers` failure, pre-existing on main)
- [x] `cargo xtask smoke` — all 43 markers present

### DIFF
--- a/kernel/src/fs/ext2/aops.rs
+++ b/kernel/src/fs/ext2/aops.rs
@@ -699,7 +699,19 @@ impl AddressSpaceOps for Ext2Aops {
                         // need to pin the strong reference here, the
                         // upcoming `readpage` will re-`bread` and
                         // observe the cache hit.
-                        let _ = super_ref.cache.bread(super_ref.device_id, abs as u64);
+                        //
+                        // On device error, stop the prefetch walk — a
+                        // failing device is unlikely to succeed on the
+                        // next block, and continuing would just spin
+                        // through more errors. The upcoming `readpage`
+                        // will surface the faithful errno on demand.
+                        if super_ref
+                            .cache
+                            .bread(super_ref.device_id, abs as u64)
+                            .is_err()
+                        {
+                            return;
+                        }
                     }
                     Ok(None) => {
                         // Sparse hole — `readpage` zero-fills inline,
--- a/kernel/tests/ext2_readahead.rs
+++ b/kernel/tests/ext2_readahead.rs
@@ -350,19 +350,21 @@ fn readahead_sparse_holes_dont_warm_phantom_blocks() {
     // Page 1: all holes — readahead must complete without panicking
     // and without trying to bread any made-up block. We can't probe
     // "no phantom block was warmed" directly (we don't know what
-    // absolute block a phantom would have been), so the contract is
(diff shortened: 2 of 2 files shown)


dburkart:
@coderabbitai review

PR #911, opened 2026-05-06
Pin mapped files independently of open_count for orphan finalization

### LINKED ISSUE #811: Follow-up: CodeRabbit findings from PR #802 (mmap + ext2 orphan finalize)
Follow-up to merged PR #802. CodeRabbit flagged 2 actionable items that were not addressed inline due to concurrent merges; PR was merged with mergeStateStatus=CLEAN.

## 1. Pin mapped files independently of `open_count` (Major / Heavy lift)

**File:** `kernel/src/fs/ext2/inode.rs` (around line 574)

`FileObject` is designed to outlive the originating `OpenFile`, but the new path only captures the page cache. Ext2's last-close path is still keyed solely off `open_count`, so `unlink() + mmap() + close()` can drive `orphan_finalize::finalize()` while the returned mapping can still fault through the cache. That can free or recycle the inode's blocks under a live VMA.

Add a separate mmap/mapcount pin, or otherwise defer orphan finalization while any `FileObject` derived from this inode is still alive.

## 2. Drop mapped objects before unmounting the fixture (Minor / Quick win)

**File:** `kernel/tests/ext2_mmap.rs` (around line 221)

The test calls `sb.ops.unmount()` while its `Arc<dyn VmObject>`s (`obj1`, `obj2`) are still live. Keeping them alive during teardown makes the test less isolated and can hide lifetime bugs around active mmaps.

Suggested teardown order:
```diff
+    drop(obj1);
+    drop(obj2);
     drop(of);
     drop(inode);
     sb.ops.unmount();
     drop(super_arc);
```

dburkart:
Closes #811

## Summary
- Adds a `map_count` atomic to `Ext2Inode` that tracks outstanding `FileObject` mappings independently of `open_count`, preventing `orphan_finalize::finalize()` from running while a mapping can still fault through the page cache.
- Introduces `Ext2MmapGuard` — a drop guard stored in `FileObject._mmap_guard` that bumps `map_count` on construction and decrements it on drop. Both `release` (open_count -> 0) and `Ext2MmapGuard::drop` (map_count -> 0) call `try_finalize_orphan`, which only runs finalize when both counters are zero and the inode is unlinked.
- Fixes `mmap_returns_file_object_against_lazy_page_cache` test to drop mapped objects before the `OpenFile`/inode/superblock to prevent use-after-unmount.

## Test plan
- [x] `cargo xtask build` — compiles cleanly
- [x] `cargo xtask test` — all tests pass (only known-flaky `rwlock_concurrent_readers` fails, pre-existing on main)
- [x] `cargo xtask smoke` — all 43 markers present

### DIFF
--- a/kernel/src/fs/ext2/inode.rs
+++ b/kernel/src/fs/ext2/inode.rs
@@ -177,9 +177,9 @@ pub struct Ext2Inode {
     /// Driver-side outstanding-opens refcount. Bumped by
     /// [`FileOps::open`] (one per successful `OpenFile::new`) and
     /// decremented by [`FileOps::release`] (one per `OpenFile::Drop`).
-    /// When [`unlinked`](Self::unlinked) is set and this count
-    /// transitions from one to zero, the `release` hook calls
-    /// [`super::orphan_finalize::finalize`] directly to drive the
+    /// When [`unlinked`](Self::unlinked) is set and **both** this count
+    /// and [`map_count`](Self::map_count) are zero, the `release` hook
+    /// calls [`super::orphan_finalize::finalize`] directly to drive the
     /// RFC 0004 §Final-close sequence — bypassing the VFS
     /// `gc_queue`/`evict_inode` indirection that would otherwise be
     /// blocked by the orphan-list `Arc<Inode>` pin (chicken-and-egg:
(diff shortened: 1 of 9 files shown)


dburkart:
@coderabbitai review

PR #915, opened 2026-05-06
fs/ext2: force-RO on uncertain sync_dirty_buffer

### LINKED ISSUE #806: fs/ext2: tighten durability discipline — force-RO on uncertain sync_dirty_buffer
## Motivation

Discovered while merging #803 (writepage / #750). CodeRabbit flagged that the ext2 driver currently treats every `sync_dirty_buffer` failure as rollback-safe — `write_file_at`, `setattr`, `link`, `create`, `flush_inode_slot`, and the new `Ext2Aops::writepage` all do "free + restore" on a failed sync.

This is technically wrong: a `sync_dirty_buffer` returning `Err` does **not** prove the buffer never reached disk. The block layer could have committed the bytes and then failed on the ack path, in which case freeing the underlying block leaves the (still-durable) on-disk inode/indirect slot pointing at a bitmap-free block. On a subsequent allocation that re-hands-out the same block, two on-disk objects now alias the same physical block — exactly the double-allocation hazard RFC 0004 §Write Ordering calls out.

The current write paths have lived with this hazard since #565 (block bitmap allocator) without a force-RO latch on the sync surface. Tightening it is a cross-cutting change — it should land in one PR that covers every `sync_dirty_buffer` call site at once, not piecemeal.

## Work

- Audit every `sync_dirty_buffer` call site in `kernel/src/fs/ext2/` (grep currently shows `balloc.rs`, `create.rs`, `file.rs`, `link.rs`, `rename.rs`, `setattr.rs`, `unlink.rs`, `aops.rs`).
- Settle the policy: should the buffer cache itself trip `force_ro_latch` on a sync failure (cleaner — single hand-off point), or should each call site decide?
- Update RFC 0004 §Write Ordering / §Security with the chosen policy and the e2fsck-recovery rationale.
- Apply the policy uniformly. Existing call sites that "free + restore" today must either (a) keep the block allocated and latch RO, or (b) be exempted with a documented reason.
- Test: a deliberately-failing `BlockDevice::write_at` on a fresh mount should leave `is_writable() == false` after one sync failure, and the failing `write(2)` / writepage / setattr should return `EIO`.

## Context

- PR #803 (writepage) — discovered while merging.
- RFC 0004 §Write Ordering, §Security ("double-free → EIO + force RO" already exists for the bitmap path; this proposal extends the same posture to the broader sync surface).
- Existing posture in `Ext2Super::force_ro` / `force_ro_latch` — the latch already exists; this issue is about widening its trigger.

dburkart:
Closes #806

## Summary
- Centralises the "uncertain durability -> force-RO" policy in `BlockCache::sync_dirty_buffer` instead of relying on each call site to decide independently. On any device write failure, the buffer cache atomically trips the filesystem-level `force_ro_latch`, causing every subsequent ext2 write-path entry point to refuse with `EROFS`.
- All ext2 write paths (`write`, `create`, `setattr`, `link`, `rename`, `unlink`, `writepage`) now gate on `Ext2Super::is_writable()` (which checks `force_ro_latch`), replacing the previous two-flag `RDONLY || FORCED_RDONLY` checks.
- Call sites that previously did "free + restore" on sync failure now guard that rollback behind `is_writable()` -- if the mount is force-RO (sync error), the rollback is skipped to avoid double-allocation risk. Bitmap leaks are safe for `e2fsck` to reclaim.
- RFC 0004 updated: ss Write Ordering and ss Security now document the centralised force-RO-on-sync policy and the e2fsck-recovery rationale.

## Test plan
- [x] `cargo xtask build` -- clean build
- [x] New integration test `ext2_force_ro_sync` passes: injected write failure latches force-RO, first write returns EIO, second write returns EROFS
- [x] `cargo xtask smoke` -- all 43 markers present
- [x] Full `cargo xtask test` -- only known flaky `rwlock_concurrent_readers` fails (pre-existing, unrelated)


dburkart:
@coderabbitai review

Issue #791, opened 2026-05-02
blocking_sync rwlock test fails on main: readers serialised — rwlock didn't grant shared access


dburkart:
## Motivation

Pre-existing `cargo xtask test` failure on `main` (sha `2760491`) discovered during #721's auto-engineer run:

\`\`\`
tasks: reaped task 23 (stack VA slot 0xffffd0000000f000..0xffffd00000014000 returned)
tasks: reaped task 25 (stack VA slot 0xffffd0000000a000..0xffffd0000000f000 returned)
[failed]
Error: panicked at kernel/tests/blocking_sync.rs:796:5:
assertion \`left == right\` failed: readers serialised — rwlock didn't grant shared access
  left: 2
 right: 3
Error: \"test blocking_sync-bacb80b06701ea42 failed\"
error: test failed, to rerun pass \`-p vibix --test blocking_sync\`
\`\`\`

The assertion is checking the rwlock's shared-access semantics: 3 readers should observe each other concurrently, but only 2 did.

## Reproduction

Reproduces consistently on main HEAD (`2760491`), unrelated to #721's changes:

\`\`\`sh
git checkout main
cargo xtask test
\`\`\`

## Suspected causes

1. **Genuine rwlock regression** — recent changes to `kernel/src/sync/rwlock.rs` or its test harness may have introduced a race where the third reader blocks until one of the first two releases.
2. **Test harness flake** — the test polls reader entry counts with an HPET-driven timeout; if the timer-tick rate vs. the poll-loop interval changed (recent #786, #785 touched scheduler-adjacent code), the third reader may not get scheduled within the test's window.

## Work

- [ ] Bisect to identify which recent commit introduced the failure.
- [ ] Inspect `kernel/tests/blocking_sync.rs:796` and the rwlock code path it exercises.
- [ ] Either fix the rwlock or relax the test's scheduling-window assumption (only the latter if root cause is firmly demonstrated to be QEMU timing).

## Discovered while landing

#721 / PR with the simulator regression test for #501. The `cargo xtask test` invocation that the auto-engineer playbook runs as part of the build/test gate failed with this signature; the failure reproduces on `main` itself, so it's not introduced by #721's changes.

## Context

- File: `kernel/tests/blocking_sync.rs:796`.
- Recent scheduler-adjacent merges: #786 (DST Phase 2 (6.5)), #785 (mem: per-inode read-ahead), #782 (DST Phase 2 (6)).

PR #932, opened 2026-05-19
fs: add /proc skeleton (uptime, meminfo, cpuinfo, self/status, pid/cmdline)

### LINKED ISSUE #398: fs: add /proc skeleton (uptime, meminfo, cpuinfo, self/status, pid/cmdline)
## Motivation

`/proc` is the conventional surface that turns kernel state into composable userspace data. Today `mem` and `tasks` are shell builtins that print formatted text — useful, but unreadable from another program. Wiring the same data through a synthetic `/proc` filesystem (just like `devfs`) lets future tools (`ps`, `top`, `free`, `uname` reading `/proc/version`) work without each one needing a bespoke syscall. The cost is small because the backend pattern (synthetic file backends with `read`-on-demand) already exists in `kernel/src/fs/devfs/`.

## Work

- [ ] Add `kernel/src/fs/procfs/` modeled on `devfs` — a `Filesystem` impl that synthesises file content on each read. No real storage; each entry implements a small `fn read(buf) -> usize`.
- [ ] Mount `/proc` at boot alongside `/dev` and `/tmp`.
- [ ] Initial entries:
  - `/proc/uptime` — seconds since boot, idle seconds (idle = uptime until the idle task exists).
  - `/proc/meminfo` — `MemTotal`, `MemFree`, `MemAvailable` in Linux key-value format.
  - `/proc/cpuinfo` — one block per CPU (just CPU 0 today): vendor, model, family, flags from CPUID.
  - `/proc/version` — kernel name + git SHA + build date (reuses the `build_info` module).
  - `/proc/self/status` — current task: pid, ppid, state, uid (0), threads (1), VmSize, VmRSS.
  - `/proc/<pid>/cmdline` — null-terminated argv of the named task; ENOENT on missing pid.
- [ ] Directory listing: `getdents64` on `/proc` should yield static entries plus one per live pid.
- [ ] Integration test: `kernel/tests/procfs.rs` — open each entry, read it, assert the expected substring; assert `/proc/<pid>` appears for a spawned child.
- [ ] Refactor the `mem` and `tasks` builtins to read from `/proc/meminfo` and `/proc/<pid>/status` (proves the surface and removes duplicated formatting).

## Context

- `kernel/src/fs/devfs/` — direct precedent for synth filesystem
- `kernel/src/fs/vfs/` — mount table, dentry cache
- `kernel/src/sched/` — task table source for `/proc/<pid>` enumeration
- Depends on the `build_info` module from the introspection-builtins issue (for `/proc/version`).
- Naturally pairs with the file-touching builtins issue: once `cat` and `ls` exist, `cat /proc/meminfo` and `ls /proc` become the canonical interaction.
- Future `ps`, `top`, `free` will read this surface without further kernel changes.
- Surfaced during a userspace-completeness survey on 2026-04-15

dburkart:
Closes #398

## Summary
- Add `kernel/src/fs/vfs/procfs.rs` — a synthetic filesystem modeled on the existing `devfs` pattern that synthesises file content on each `read` with no backing storage.
- Mount `/proc` at boot alongside `/dev` and `/tmp`, registered as both `procfs` and `proc` filesystem types.
- Static entries: `/proc/uptime` (seconds since boot), `/proc/meminfo` (MemTotal/MemFree/MemAvailable), `/proc/cpuinfo` (model name + feature flags), `/proc/version` (kernel name, release, git SHA, build timestamp).
- Per-pid entries: `/proc/self/status` and `/proc/<pid>/status` (pid, ppid, state, uid, threads), `/proc/<pid>/cmdline` (NUL-terminated task name).
- `getdents64` on `/proc` yields static entries plus one directory per live pid, snapshotted from the scheduler's task table.
- Added smoke marker `vfs: mounted procfs at /proc` to protect the mount in CI.

## Test plan
- [x] `cargo xtask build` — compiles cleanly (one pre-existing warning)
- [x] `cargo xtask test` — 726 host unit tests pass; the only QEMU integration failure is the known `blocking_sync` rwlock flake (#791)
- [x] `cargo xtask smoke` — all 44 markers present (including the new procfs marker)
- [x] Host unit tests in `procfs::tests` cover mount, root dir, lookup of all static entries, read content assertions, offset reads, getdents, statfs, and distinct-mount ino separation
- [x] Host unit tests in `init::tests` verify `/proc` mount is present in root children and survives via the parent's children map


dburkart:
## Review findings — deferred to follow-up

CodeRabbit raised two findings:

1. **ENOENT for vanished `/proc/<pid>/cmdline`** — currently returns empty bytes (EOF) when the PID no longer exists. This is consistent with the race between readdir and read in Linux, but returning ENOENT would be cleaner. Deferred to follow-up.

2. **`/proc/self` dirent type DT_LNK vs directory resolution** — `self` is emitted as DT_LNK (matching Linux where `/proc/self` is a symlink) but resolved to a directory inode in lookup. Deferred to follow-up — the current behavior works for all common access patterns.

Both are polish items that don't affect the skeleton's core functionality. The pre-existing `wait4_condvar_race` timeout in integration tests 4/4 is #791, unrelated to this PR.

Merging — follow-up items will be captured in a subsequent issue.

